PR #96, opened 2014-10-18
New package easyoptions

renatosilva:
[EasyOptions](https://github.com/renatosilva/easyoptions) makes it easy parsing options in Bash and Ruby scripts. For example:

* https://github.com/renatosilva/scripts/blob/master/bzrgrep.rb
* https://github.com/renatosilva/pidgin-plus/blob/master/build/winbuild.sh

### DIFF
--- a/easyoptions-git/PKGBUILD
+++ b/easyoptions-git/PKGBUILD
@@ -1,0 +1,31 @@
+# Maintainer: Renato Silva <br.renatosilva@gmail.com>
+
+_realname='easyoptions'
+pkgname="${_realname}-git"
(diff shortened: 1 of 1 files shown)


fracting:
I spent some time looking on the CI failure build, and found it is a false positive, I can confirm the package builds correctly in my local machine.

Cross Reference: https://github.com/Alexpux/MSYS2-packages/commit/0cc5c7b85251878c163ee97503522735557acadb

fracting:
Now the build should success on latest CI server. @renatosilva I guess you can rebase and push again to trigger a new build?

PR #97, opened 2014-10-18
Fix Bazaar cloning in makepkg

renatosilva:
This is because makepkg was requiring the cloned repository to be removed across builds, because it was erroneously thinking that the repository is not a real clone of `$source`. More info in the commit message.


renatosilva:
This has been reported upstream: https://bugs.archlinux.org/task/43448.

PR #157, opened 2015-02-05
Set correct ACLOCAL_PATH in /etc/profile for mingw{32,64}_shell

elieux:


elieux:
Fixed according to comments on IRC.

PR #167, opened 2015-02-15
Fix MAN_DIRS expansion in makepkg.conf

rossy:
MAN_DIRS was missing a brace, which prevented man pages for mingw packages from being compressed. I added the missing brace and also moved around some slashes to prevent doubled and missing slashes in the paths.


rossy:
Hmm, I forgot to increase pkgrel. Should it be increased so the package is rebuilt?

mingwandroid:
Yes please

PR #176, opened 2015-02-23
Add ccache-wrapper PKGBUILD

flatwhatson:
On linux systems (eg. ubuntu), ccache installs symlinks to /usr/lib/ccache which provide transparent caching when added to PATH.  On windows we don't really have symlinks, but a shell script can do the job.

This package:
  - installs shell script wrappers to /usr/lib/ccache-wrapper/{mingw32,mingw64}
  - provides wrappers for current gcc and clang executables
  - adds a hook to profile.d adding the appropriate wrappers to PATH

Concerns:

No hard dependencies in the package means that it's possible to get silly errors using wrappers when ccache or the selected compiler isn't installed.  Should this package be split into separate mingw32/mingw64 versions, with appropriate dependencies?  We don't want to force people to install toolchains they dont' need.

The hard-coded list of compilers will be annoying to maintain.  It's probably already missing some supported toolchains, and would need updating regulary for version changes.  Maybe we could add wrappers for installed compilers as a post-install hook?  Any compilers installed *after* ccache-wrappers would be missed though, so maybe a setup-ccache-wrappers script to be run manually?  Ugh.

Automatically modifying PATH via the profile.d hook might be controversial.  Setting CCACHE_DISABLE=1 prevents caching when needed, but maybe the PATH modification should be opt-in?

Any suggestions for improvement are greatly appreciated.

### DIFF
--- a/ccache-wrapper/PKGBUILD
+++ b/ccache-wrapper/PKGBUILD
@@ -1,0 +1,69 @@
+# Maintainer: Andrew Whatson <whatson@gmail.com>
+
+pkgname=ccache-wrapper
+pkgver=1
+pkgrel=1
+pkgdesc='Wrapper scripts for ccache'
+arch=('any')
+license=('GPL')
+depends=('bash')
+optdepends=(
+  'mingw-w64-i686-ccache-git'
+  'mingw-w64-i686-gcc'
+  'mingw-w64-i686-clang'
+  'mingw-w64-x86_64-ccache-git'
+  'mingw-w64-x86_64-gcc'
+  'mingw-w64-x86_64-clang'
+)
+
+package() {
+  reallibdir=/usr/lib/${pkgname}
+  libdir=${pkgdir}/${reallibdir}
+  etcdir=${pkgdir}/etc/profile.d
+
+  compilers_any='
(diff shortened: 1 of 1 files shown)


Alexpux:
I don't like this hack. So in this form it will not be accepted.

flatwhatson:
What in particular don't you like about it?  I'm happy to make improvements.

Have a look at the archlinux ccache PKGBUILD, where they put a bunch of symlinks into /usr/lib/ccache - that's what I'm trying to replicate.  Unfortunately 'ln' is 'cp' by default in this environment, so that can't possibly work.  Hence shell script wrappers.

https://projects.archlinux.org/svntogit/community.git/tree/ccache/trunk/PKGBUILD?id=ccd635118624e537bb0804fa1823ff71bcf5fb05

Would it be better to include the wrappers in the mingw-w64-ccache-git package?

Alexpux:
Having discussed it on IRC, you need create package in MINGW-packages instead.
Also maybe good to create separate package like "gcc-ccache" as part of ming-gcc PKGBUILD.
Or extend GCC package by adding some folders for ccache.

PR #178, opened 2015-02-25
Updated PKGBUILD and added 0001_msysize.patch for 0.8.17

useretail:


Alexpux:
Don't need this. Need run "autoreconf" after patch.

useretail:
Actually 0000_msysize.patch doesn't work with 0.8.17, it's for 0.8.15 (I guess).

Alexpux:
iris package is under development still. Feel free to update patch, but patch need to modify next autotools files: configure.ac, Makefile.am, *.m4(except libtool macroses)

PR #193, opened 2015-03-13
bash: Add missing build dependencies

dscho:

### DIFF
--- a/bash/PKGBUILD
+++ b/bash/PKGBUILD
@@ -11,7 +11,8 @@ license=('GPL')
 url="http://www.gnu.org/software/bash/bash.html"
 groups=('base')
 validpgpkeys=('7C0135FB088AAF6C66C650B9BB5869F064EA74AB') # Chet Ramey
-makedepends=('gettext-devel' 'libreadline-devel>=6.3' 'ncurses-devel')
+makedepends=('gettext-devel' 'libreadline-devel>=6.3' 'ncurses-devel'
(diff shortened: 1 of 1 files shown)


Alexpux:
In this case you need add such dependencies to EVERY package. We decide to not do it as for develop packages you need install ````base-devel```` group of packages in any case.

dscho:
Ah, okay. I just thought that maybe for the most commonly to-be-rebuilt packages (such as msys2-runtime and bash) it would make sense to make it easier for contributors who do not know about that `base-devel` thing (I, for one, did not)...

elieux:
Just adding some info for the curious.

The implicit dependency on `base-devel` is documented on our [Buliding packages](https://sourceforge.net/p/msys2/wiki/Contributing%20to%20MSYS2/) wikipage and it's a heritage from Arch Linux (documented on [Creating packages](https://wiki.archlinux.org/index.php/Creating_packages) on the ArchWiki). Upstream [FS#11766](https://bugs.archlinux.org/task/11766) describes some reasons for the current state of things -- why there's no check in makepkg a why PKGBUILDs omit these dependencies.

dscho:
Here's an idea: why not just teaching `makepkg -s` to add that dependency implicitly? (And the equivalent dependency on the toolchain with `makepkg-mingw -s`?) Then nobody would need to be puzzled, and the documentation would also be unnecessary...

PR #200, opened 2015-03-26
Added markdown.

joca-bt:
Added package markdown, that converts text to markdown.


elieux:
Isn't the package missing `depends`?

joca-bt:
No, it only depends on perl, which is part of base.

mingwandroid:
You mean base-devel? We removed perl from base (the Python developers complained :-))

joca-bt:
Bollocks! I'll correct this tonight. Thanks and sorry.

mingwandroid:
No problem.

PR #202, opened 2015-03-29
New package sharutils

renatosilva:
This package contains the `uuencode` and `uudecode` utilities.


sharutils/install
```diff
@@ -0,0 +1,18 @@
+infodir=/usr/share/info
```
Alexpux: remove slash before usr

PR #209, opened 2015-04-07
Lua: new package

huettner94:
added new package "lua"
patch required to build liblua.so


mingwandroid:
Why would we want lua as a MSYS2 package instead of a MinGW-w64 package?

huettner94:
Sorry, I did not see that.
I will close this pull request and i will create a new pull request on the mingw repo to update the lua version to 5.3

PR #220, opened 2015-04-16
New package vimpager

renatosilva:
Syntax highlighting for man pages and the cat command.

vimpager

### DIFF
--- a/vimpager-git/PKGBUILD
+++ b/vimpager-git/PKGBUILD
@@ -1,0 +1,33 @@
+# Maintainer: Renato Silva <br.renatosilva@gmail.com>
+
+_realname='vimpager'
+pkgname="${_realname}-git"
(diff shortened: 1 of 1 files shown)


vimpager-git/PKGBUILD
```diff
@@ -0,0 +1,33 @@
+# Maintainer: Renato Silva <br.renatosilva@gmail.com>
+
+_realname='vimpager'
+pkgname="${_realname}-git"
+pkgdesc='Use ViM as PAGER'
+url='http://www.vim.org/scripts/script.php?script_id=1723'
+license=(BSD)
+arch=(any)
```
Alexpux: is pager executable? If yes you need set arches

renatosilva: Both vimpager and vimcat are scripts.

PR #241, opened 2015-05-07
elinks-git: Rebuild with gnutls.

mashir43:


Alexpux:
Why build it with gnutls instead openssl?

mashir43:
I notice that elinks's GNU license dosen't include OpenSSL exception.
So, I rebuild with GnuTLS.

PR #248, opened 2015-05-15
python3: add patch from upstream preventing crashes in the ctypes module

samhocevar:


elieux:
To provide some context: We had this patch for v3.4.2, but it got removed for 3.4.3, but I don't see why.

PR #254, opened 2015-05-22
Add ragel as a tool

ebraminio:
It is based on:
* https://github.com/KaOS-Community-Packages/ragel/blob/master/PKGBUILD
* https://github.com/johnbartholomew/pkgbuilds/blob/master/ragel/PKGBUILD

And it works well locally in order to compile a library I wanted to use this for.

### DIFF
--- a/ragel/PKGBUILD
+++ b/ragel/PKGBUILD
@@ -1,0 +1,24 @@
+pkgname=ragel
+pkgver=6.9
+pkgrel=1
+pkgdesc="Compiles finite state machines from regular languages into executable C, C++, Objective-C, or D code."
+arch=('i686' 'x86_64')
+url="http://www.colm.net/open-source/ragel/"
+license=('GPL')
+depends=('gcc-libs')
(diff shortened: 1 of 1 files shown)


elieux:
Wouldn't this be better suited as a mingw package? I don't see much reason for including it as a msys package.

ebraminio:
Hmm. Isn't MSYS2 packages meant for tools and Mingw packages for things needed as a library? Just a bit background, ragel is a build tool and isn't meant to be linked on any library. So which repo is more suitable for it?

elieux:
The distinction is a bit different. The MSYS2 project (as a distro) is focused on providing native Windows (i.e. mingw-compiled) libraries **and** software and the POSIX part (msys-compiled) is mostly for when porting is impossible or infeasible.

Of course, build tools are often problematic to port correctly, so they end up as msys packages, but so far, I don't see any reason to go this route for the proposed tool. Note that I don't know this software, so I'm open to arguments.

ebraminio:
Ok :)

PR #255, opened 2015-05-23
atool: New package.

goreliu:
Name           : atool
Description    : A script for managing file archives of various types
Architecture   : any
URL            : http://www.nongnu.org/atool/
Licenses       : GPL3


goreliu:
Moved to MINGW-packages.

PR #261, opened 2015-06-04
gnu-cobol: New package.

mashir43:


Alexpux:
Use mingw version of cobol instead.
https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-gnu-cobol-svn

Issue #214, opened 2015-04-12
Once PKGBUILD is added to this repository, when would a new package be available for installation?


legends2k:
Around a month ago, I'd added a new package `bc` in this repository. It's yet to be available for installation using `pacman`.

Is there something more that I need to do (other than pushing the `PKGBUILD` file) in order to get the package's binaries built and use `pacman` to install and use them? I'd like to know the process that happens between someone submitting a `PKBUILD` and it's binaries appearing in the [sourceforge.net repository](http://sourceforge.net/projects/msys2/files/REPOS/MSYS2/).

elieux:
The process is that @Alexpux notes what packages to build and then builds them after some time. `bc` was probably missed by accident.

Alexpux:
I'm try to build msys version of "bc" package but it fail and I don't yet have time to fix it.
@legends2k, try building both 32/64 versions of this package itself first.

legends2k:
@Alexpux Sure, I'll do it. I noticed that it's available on mingw64's repos which is also not working correctly. I'll look at both, will do it in a week, stuck in something else now.

@elieux Thanks for the info, I thought it was some formal procedure.

legends2k:
@Alexpux I built `bc` on a fresh, clean installation of 32-bit MSYS2. Here's the output

```
$ makepkg
==> Making package: bc 1.06-1 (Fri, May 01, 2015 10:15:09 AM)
==> Checking runtime dependencies...
==> Checking buildtime dependencies...
==> Retrieving sources...
  -> Found bc-1.06.tar.gz
==> Validating source files with md5sums...
    bc-1.06.tar.gz ... Passed
==> Extracting sources...
  -> Extracting bc-1.06.tar.gz with /usr/bin/bsdtar
==> Starting build()...
loading cache ./config.cache
checking for a BSD compatible install... (cached) /usr/bin/install -c
checking whether build environment is sane... yes
checking whether make sets ${MAKE}... (cached) no
checking for working aclocal... found
checking for working autoconf... found
checking for working automake... found
checking for working autoheader... found
checking for working makeinfo... found
checking for gcc... (cached) gcc
checking whether the C compiler (gcc -march=i686 -mtune=generic -O2 -pipe -pipe) works... yes
checking whether the C compiler (gcc -march=i686 -mtune=generic -O2 -pipe -pipe) is a cross-compiler... no
checking whether we are using GNU C... (cached) yes
checking whether gcc accepts -g... (cached) yes
[...]

PR #267, opened 2015-06-11
Build bc without readline support

legends2k:
- Fix build failure on mingw64 when enabling readline feature


legends2k:
Fix for #214 .

PR #275, opened 2015-06-26
ncurses: fix crash when TERM refers to an unknown terminal

dscho:
A but that was recently introduced into the 6.x development branch of
ncurses causes free()d data to be reused by mistake, leading to a crash.

You can verify this by setting TERM to, say, `msys` and then running
`bash.exe --login -i`.

This fixes https://github.com/git-for-windows/git/issues/222

### DIFF
--- a/ncurses/0001-CRITICAL-fix-return-value-of-drv_CanHandle-on-unknow.patch
+++ b/ncurses/0001-CRITICAL-fix-return-value-of-drv_CanHandle-on-unknow.patch
@@ -1,0 +1,58 @@
+From 41f152b6e8bea7e92be960158ef4bd9df34dd8a0 Mon Sep 17 00:00:00 2001
+From: Johannes Schindelin <johannes.schindelin@gmx.de>
+Date: Fri, 26 Jun 2015 13:19:49 +0000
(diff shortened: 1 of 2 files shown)


dscho:
@Alexpux I am about to send a mail to the curses maintainers, too.

dscho:
Please note that after merging this, it is necessary to release not only a new ncurses and ncurses-devel package, but then also rebuild bash based on the new ncurses-devel package, to fix the bug described in git-for-windows/git#222 completely.

dscho:
@Alexpux is there any documentation about the process of releasing new packages in MSys2, so that I do not have to pester you all the time (I am thankful for your work, and want to avoid annoying you...)?

Alexpux:
No. I'm build/upload new packages when have free time for it

Alexpux:
@dscho, can you test latest uploaded packages? I'm disable your patch in ncurses as it seems fixed upstream

dscho:
It is indeed fixed upstream (and a bit differently: they now re-`#define` a `SETUP_FAIL` return value). I am testing as we speak.

dscho:
Tested, fixed! Thank you!

Alexpux:
ok. cool!

Issue #271, opened 2015-06-19
tmux-git: can't create socket: Permission denied


goreliu:
tmux-git 2.0.171.g7acc4ad-1:

```
$ tmux
can't create socket: Permission denied
```

tmux-git 1.9.4851.f8481f9-1 is ok.

shosca:
I suggest for now add tmux-git in IgnorePkg IgnoreGroup and NoUpgrade in pacman.conf and install the 1.9 manually

vruyr:
I've got "tmux-git 2.0.171.g7acc4ad-1" and managed to lunch it with manually specifying the socket file that doesn't exist and /dev/null for configuration:

    $ mkdir /tmp/tmux-hack/
    $ tmux -S /tmp/tmux-hack/default -f /dev/null

Not sure why, but if either of options are removed, tmux fails to start.

moguriso:
Probably tmp directory's create mask permission is 0755. But "tmux.c" expect to other's permission is not set. The patch is below. It seems that could modify our issue.

```diff
diff --git a/tmux.c b/tmux.c
index 62f8a80..4633f76 100644
--- a/tmux.c
+++ b/tmux.c
@@ -149,7 +149,7 @@ makesocketpath(const char *label)
 		errno = ENOTDIR;
 		return (NULL);
 	}
-	if (sb.st_uid != uid || (sb.st_mode & S_IRWXO) != 0) {
+	if (sb.st_uid != uid || (sb.st_mode & S_IWOTH) != 0) {
 		errno = EACCES;
 		return (NULL);
 	}
```

However I don't know the modification which is correct. And I don't confirm other environment... (ex: linux, mac...)

amosbird:
same problem here

elieux:
Probably solved by #286.

PR #285, opened 2015-07-07
Modify "can't create socket" occurred

moguriso:
This patch belongs to issue #271.
It affects to only MSYS2 environment.


Alexpux:
Don't use TAB's in PKGBUILD. Use 2 spaces instead.

PR #286, opened 2015-07-07
Modify "can't create socket" occurred

moguriso:
This patch belongs to issue #271.
(https://github.com/Alexpux/MSYS2-packages/issues/271)
It affects to only MSYS2 environment.


moguriso:
Pardon me. I closed previous ticket by mistake.

PR #317, opened 2015-08-06
addpkg: python-ply 3.6-1

felixonmars:
Add as dependency in the python-cffi chain, preparing for pyopenssl and other cffi-based packages.


Alexpux:
Why not add this packages to mingw instead?

felixonmars:
Ah... Actually I don't know when to put into MSYS2 and when for MinGW.

Alexpux:
MSYS packages in most cases needed for build tools that helps to build mingw packages. Our primary target is MINGW. MSYS is designed to simplify build process under Windows.

felixonmars:
I see, thanks for the info. I'll send another PR to add the packages to MinGW, sorry!

PR #343, opened 2015-09-11
bash-completion: avoid unnecessary tilde expansion

dscho:
When connected to an Active Directory (a variant of LDAP), tab completion
becomes very slow because the tilde is expanded all the time, even if
completing paths that do not contain a tilde at all. This developer
observed delays of several seconds at a time.

This is completely unnecessary, so let's do not do that.

The patch was provided by Ove Risberg with a bug report to the original
bash-completion project:

	https://bugs.launchpad.net/ubuntu/+source/bash-completion/+bug/1390061

Since it fixes code introduced by another add-on patch, that add-on patch
was modified instead of adding yet another patch.

This fixes https://github.com/git-for-windows/git/issues/377

### DIFF
--- a/bash-completion/PKGBUILD
+++ b/bash-completion/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=bash-completion
 pkgver=2.1
-pkgrel=5
+pkgrel=6
 pkgdesc="Programmable completion for the bash shell"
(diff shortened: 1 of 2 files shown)


dscho:
Any issues with this patch?

PR #346, opened 2015-09-13
Include the 'binutils' package into base-devel

dscho:
The recommended way to start building MSys2 packages is to call

	pacman -S base-devel

once and then run `makepkg -s` in the respective subdirectories of the
MSYS2-packages project.

Many of these packages require the `strip` utility which is provided by
the `binutils` package, and as a matter of fact, `makepkg` refuses to
start if it does not find the `strip` executable.

So let's just include the `binutils` package in the `base-devel` groups so
that it gets installed in the recommended workflow.


Alexpux:
Group ````base-devel```` designed for packages that commonly using during configure/build process and is not target specific. So binutils can't be placed in this group

elieux:
To elaborate, since building for the msys2 platform is not the primary (or even secondary) goal of a large portion of MSYS2 users, a decision has been made to keep the msys2-platform-specific packages separate. This is documented on our [wikipage about Building packages](http://sourceforge.net/p/msys2/wiki/Contributing%20to%20MSYS2/), along with the name of the additional package group recommended for building msys2 binaries -- `msys2-devel`.

dscho:
Thanks for the clarification!

PR #347, opened 2015-09-14
Add the git-flow package

dscho:
The "git flow" way to organize branches in a large team has gained some
popularity in the recent years. Let's package it for use with MSys2.

See http://nvie.com/posts/a-successful-git-branching-model/ for more
details about this branching model.

### DIFF
--- a/git-flow/PKGBUILD
+++ b/git-flow/PKGBUILD
@@ -1,0 +1,45 @@
+# Maintainer: Johannes Schindelin <johannes.schindelin@gmx.de>
+
+_realname=git-flow
+pkgname=("${_realname}")
+pkgver=0.4.1.108.g15aab26
+pkgrel=1
+pkgdesc="Git extensions to provide high-level repository operations for Vincent Driessen's branching model"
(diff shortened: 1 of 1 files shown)


git-flow/PKGBUILD
```diff
@@ -0,0 +1,45 @@
+# Maintainer: Johannes Schindelin <johannes.schindelin@gmx.de>
+
+_realname=git-flow
+pkgname=("${_realname}")
+pkgver=0.4.1.108.g15aab26
+pkgrel=1
+pkgdesc="Git extensions to provide high-level repository operations for Vincent Driessen's branching model (mingw-w64)"
```
Alexpux: Remove ````(mingw-w64)```` as we build package for MSYS

dscho: Whoopsie, sorry! Fixed.

For background: I had originally played with the idea of making `git-flow` a MinGW package depending on MinGW Git, but decided against it due to its use of the `getopt.exe` file of `util-linux`.

Issue #344, opened 2015-09-11
msystem.bat


userzimmermann:
Hi. I want to contribute **msystem.bat** from my Chocolatey MSYS2 package:

https://github.com/userzimmermann/choco-packages/blob/master/msys2/msystem.bat

It activates and switches **MSYS**, **MINGW32**, and **MINGW64** environments directly in CMD shell, by prepending the according **bin\** paths to `%PATH%` and adding tags to `%PROMPT%`, like **/etc/profile** does in Bash, with the difference that **MINGW32/64** environments don't include the basic **MSYS bin\** paths and that `%MSYSTEM%` is not set. It determines MSYS2 root by looking up `%MSYS2_ROOT%` or trying to find **msys2_shell.bat** via `%PATH%`.

    C:\> msystem msys
    <MSYS> C:\> msystem mingw32
    <MINGW32> C:\> msystem mingw64
    <MINGW64> C:\> msystem /d
    C:\>

@Alexpux's first suggestion in IRC was to put it under **filesystem/**. Is it maybe worth a separate package? Also because I have some further plans like:

* Extra script dirs dynamically prepended to `%PATH%`, containing wrapper **.bat** scripts for most important tools like **bash**, **pacman**, **makekg**, etc., which directly run in CMD shell, but internally run the actual tools properly through `bash --login` with `%MSYSTEM%` set before.
* Special **pacman.bat** scripts for **MINGW32/64** environments, which implicitly add **mingw-w64-i686/x86_64-** prefixes to requested package names.
* ...
* And last but not least a PowerShell module offering the same features via cmdlets.

But first: what do others think about that stuff?

mingwandroid:
This looks good to me, perhaps put it in a filesystem-cmd package, but it doesn't really matter either way.

I don't like trying to hide the reality of the package names (and thus the difference between them) by implicitly adding anything. This will just lead to increased confusion for little gain.

mingwandroid:
Also, why does it look for msys2_shell.bat in the path; isn't %~dp0 always what you want here?

userzimmermann:
@mingwandroid 
Because the choco msys2 package neither installs msys2 under the choco package path nor installs msystem.bat into msys2. They are separated. And I wanted the choco package extra tools to be able to work with different msys2 installations depending on env settings...

The msys2-integrated msystem.bat variant will of course work with %~dp0

mingwandroid:
[...]

PR #350, opened 2015-09-16
CMD shell integration package

userzimmermann:
Hi. And here it is: the pull request for #344 

I created a **filesystem-cmd** package, based on the suggestion from @mingwandroid 

It contains an improved **msystem.bat** (MSYS2 environment manager for CMD shell) and initial wrapper scripts for **bash** and **pacman** in **MSYS** mode.

I also added Clink auto-completion scripts for **msystem** and **pacman**. `msystem /i clink` creates a small loader script for them and installs it into Clink's settings directory.

**PKGBUILD** creates an extra **/cmd/** root dir for installing that stuff into.

The wrapper scripts still need improvements, like proper working directory delegation to **bash.exe**. Also the auto-completion for **pacman** doesn't have proper support for combined short flags yet.

I don't see that stuff as a perfect solution yet and i am open to any complaints ;) But it's a beginning for better integrating MSYS2 in the CMD shell, for those who prefer native Windows shells over Bash under Windows :)

Is it better to develop this stuff in a separate repo and just downloading release zips from that repo in **PKGBULD**?


lazka:
Is any of this still relevant?

PR #353, opened 2015-09-23
Fix winpty's locale

dscho:
This is necessary to deal with non-ASCII command-line parameters correctly.

I also submitted this as a [Pull Request to winpty](https://github.com/rprichard/winpty/pull/37) but I figured it would not hurt to have it in MSys2 as soon as possible.

This fixes https://github.com/git-for-windows/git/issues/412


dscho:
Thanks! Please note that upstream merged the corresponding PR already (I am positively surprised!), so the patch might need to be removed before building a new `winpty` package.

PR #359, opened 2015-10-06
irssi-git: Add glib2 a dependency

noahm:
After trying to run irssi on my system without glib2 installed gives the error:

    C:/msys64/usr/bin/irssi.exe: error while loading shared libraries: msys-gmodule-2.0-0.dll: cannot open shared object file: No such file or directory

Irssi runs just fine after glib2 is installed.


mingwandroid:
Thanks!

Issue #322, opened 2015-08-08
Add PHP


atrauzzi:
As per advice [here](http://sourceforge.net/p/msys2/tickets/183/).

Would love to see a build of PHP made available with most `.so` modules available as separate packages, similar to how Linux distros do it.

Worth noting: Don't compile-in cURL (even though it's perfectly reasonable to want to), otherwise Google App Engine users could encounter issues depending on how their `app.yaml` file is configured.

bardware:
PHP 7 including the php-ast extension would be nice.

mingwandroid:
Ok, but why as an MSYS2-package? I know nothing of PHP. Someone from that community would have to do the work here of maintaining it I think.

elieux:
+1 but I don't think keeping this ticket open will help anything.

the-liquid-metal:
Hello.

It has been 2 years now since this thread was created. I still not found any php package in msys2. is there any progress about it? why it took so long?

the-liquid-metal:
If i am not wrong, there are already php packages in cygwin. Is it hard to replicate it on msys?

php is able to run in console mode natively. i feel something is missing when i call php on command prompt.  Somehow, the coloring output is disappears. did i forget something?

atrauzzi:
TBH, I do suggest anyone reading consider Windows Subsystem for Linux. If you don't have Windows 10 and can't use Linux, it's worth that alone.

mati865:
Cygwin emulates POSIX and MinGW is about creating native software (without emulation).

muv1:
https://github.com/msys2/MSYS2-packages/pull/371

PR #371, opened 2015-10-24
add package: php

notona:
a popular General purpose scripting language.

I can't use stable release of php. stable version of php is 5.6.
but php 5.6 needs bison 2.6 to build. so i  used RC version to build..

### DIFF
--- a/php/PKGBUILD
+++ b/php/PKGBUILD
@@ -12,11 +12,12 @@ md5sums=('7ff1fd4d57f40f5286aff102c67973bb')
 build() {
   cd "${srcdir}/php-src-php-${pkgver}"
    ./buildconf --force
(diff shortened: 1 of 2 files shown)


mingwandroid:
Would this not be better in MINGW-packages

notona:
oh sorry. that is better i think.

i will move to this pull request to MINGW-packages repository

thanks.

notona:
@mingwandroid 

i have built test as MINGW-packages, but an unknown errors has occured.
so i wound like to this pull request again.

thank you for making the time to this matter.

Alexpux:
Sorry but we don't want to have PHP as MSYS package. MSYS2 is designed for developing MINGW applications in first place. If you want PHP as Unix application you seems want to install Cygwin but not MSYS2

notona:
okay. i understand.

thanks!

muv1:
https://github.com/msys2/MSYS2-packages/issues/322

PR #377, opened 2015-11-02
Run update-core as a shell function instead of its own shell

DavidEGrayson:
This is the second part of https://github.com/Alexpux/MSYS2-pacman/pull/20

It fixes a crash that happens if $PS1 runs a subcommand.

### DIFF
--- a/filesystem/profile
+++ b/filesystem/profile
@@ -135,6 +135,12 @@ export PATH MANPATH INFOPATH PKG_CONFIG_PATH USER TMP TEMP PRINTER HOSTNAME PS1 
 export TERM=xterm-256color
 unset PATH_SEPARATOR
 
+# This allows update-core to change PS1 and PROMPT_COMMAND
+# for the user's shell, which avoids a crash.
+update-core() {
+  source /usr/bin/update-core
+}
+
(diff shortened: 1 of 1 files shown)


DavidEGrayson:
I haven't thought carefully about whether all the stuff defined in update-core will be visible in the user's shell and cause problems for them in the case where update-core returns without updating anything.

mingwandroid:
You can run something in a sub shell by just enclosing in parentheses:

TEST=world; (TEST=hello; echo $TEST); echo $TEST

.. so I'd just do `(/usr/bin/update-core)` instead

DavidEGrayson:
Unfortunately, `TEST=world; (export TEST=hello; echo $TEST); echo $TEST` prints "hello" and "world" so I think the parentheses would defeat the point of this pull request.

mingwandroid:
Ah right, I didn't read it carefully enough. Can you fix the typo then "modif" -> "modify"? Yeah, I'm being pedantic at this stage.

DavidEGrayson:
The namespace pollution really bothers me; update-core defines a lot of shell functions with generic names.  I'll do some tests and see if it is sufficient to just have a while-loop at the end of update-core that asks the user to close the window if any packages were updated.

PR #391, opened 2015-11-21
emacs for MSYS2

RickLeaf:
emacs POSIX version for MSYS2.

Since no X in MSYS2, we use win32 native GUI like cygwin.


mingwandroid:
We need to think about this one. Is the mingw-w64 Emacs package not enough v for you?

Generally we don't like to have both versions of packages if we can avoid it. If the used with the mingw-w64 version can be fixed then that is better.

I need the opinions of outer people here too. I'm not opposed to merging the PKGBUILDs under the understanding that we will not package it, but perhaps that needs to be made explicit via a variable in the PKGBUILDs from now on, e.g:

not_packaged_rationale="Packaged in MINGW-packages"

Or (not applicable to this case!):

not_packaged_rationale="Packaged in non-git PKGBUILD"

RickLeaf:
The mingw-w64 emacs version is for windows native only, we call it NTEmacs some years ago.
If you use it in mintty, you will get a lot of issue about POSIX and pip.
I use msys2 as my GNU environment like linux, so I prefer use posix version like vim for msys.

RickLeaf:
A simple example is you can not use "emacs -nw" in mintty, and wintty is not best choice too.

elieux:
My general opinion -- this is a job for Cygwin.

As for pulling the PKGBUILDs to our repository -- that sounds okay if they satisfy the usual requirements.

It'd be nice to have some way to comment on PKGBUILDs, about its quality and about package status. I was hoping the way would be something less involved than putting them into git. (For example the [googlesheet](https://docs.google.com/spreadsheets/d/1hDHQFVzKfC1u5h8lNmaZvdCsxLdGo9LHYCNjyoOi8Zw/edit) I created, or some AUR-like web interface.)

RickLeaf:
Thanks for your reply

mingwandroid:
@elieux maybe it's just me, but I don't regard git, or pulling them in as that heavy-weight a thing .. so long as it's annotated as not-to-be-built, but I'm repeating myself I fear.

Nice spreadsheet by the way.

orontee:
My understanding is that msys packages provides build tools and mingw-64 ones are `redistribute' package that one can bundle with its binaries to distribute them. With this in mind I'd like to see emacs moving to msys. Note that Vi is already there!

Alexpux:
As I understand emacs use GTK for GUI. To build it as MSYS package need lots of dependencies and someone who will maintain this. Also it need to build XOrg for MSYS2. MSYS2 - is MINIMAL system.
So I will never add emacs as msys package. If you want to see it as msys package - create your own repo and build it yourself.

RickLeaf:
emacs can use win32 native gui, it don't need any xorg relative package.

PR #406, opened 2015-12-18
readline: convert clipboard data to UTF-8 encoding

dscho:
The call to GetClipboardData(CF_TEXT) returns the text in the system
encoding, which was then incorrectly treated as UTF-8 encoded. Instead we
need to use GetClipboardData(CF_UNICODETEXT) to get the text in UTF-16,
followed by a conversion to UTF-8.

Also fixed a problem where CloseClipboard was only called if
GetClipboardData returns non-null. According to the documentation, every
successful call to OpenClipboard should be matched with a call to
CloseClipboard.

This fixes https://github.com/git-for-windows/git/issues/567

Signed-off-by: Kim Gybels <kgybels@infogroep.be>


mingwandroid:
Thanks!

Issue #376, opened 2015-11-02
Current tmux broken


573:
I tried using `tmux`:

    $ tmux -V
> tmux 2.2

    $ pacman -Ss tmux-git
> msys/tmux-git 2.1.61.gc975de2-1 [Installiert]
    A terminal multiplexer

When I'm running `tmux` this is what I get (`tmux.exe.stackdump`):

> Exception: STATUS_ACCESS_VIOLATION at rip=0018019C1FD
rax=0000000000000008 rbx=0000000600075280 rcx=0000000000000008
rdx=000000000022CE00 rsi=0000000000229C50 rdi=0000000000000008
r8 =0000000000000158 r9 =0000000000000158 r10=0000000000230000
r11=00000001004436EA r12=0000000000000000 r13=0000000000000000
r14=0000000000000000 r15=0000000600069F80
rbp=0000000000000000 rsp=0000000000229AB8
program=C:\ProgramData\msys64\usr\bin\tmux.exe, pid 9964, thread main
cs=0033 ds=002B es=002B fs=0053 gs=002B ss=002B
Stack trace:
Frame        Function    Args
00000000000  0018019C1FD (001801297AB, 00000229B18, 00000229B20, 00600075280)
00000000000  001004437AF (3F300000018, 0060006B100, 00000000000, 00600074FC8)
00000000000  0010042F52F (00000000001, 00000000030, 00000000000, 00000000008)
00600051BC0  0010040BAB4 (00600051BC0, 00000229CF0, 000000003F3, 737365732D77656E)
00000229CF0  0010040D02D (00600051BC0, 00000000000, 00600057500, 00000000001)
00600057AB8  0010040D188 (0010042B71B, 00000000000, 00600057500, 0000022A198)
00600057AB8  0010042B4B7 (7FEFF5A00E0, 0000022A220, 00000395B90, 00000000000)
00600057AB8  0010042BBE7 (00600057AB0, 000000000C8, 001802FC488, 0000022A680)
00600057AB8  00100426510 (00600016138, 00000000005, 001801297AB, 00600067AF8)
0060001D140  004E2CA338C (001800CDDE3, 00000000005, 00100469380, 00000000000)
00000000001  004E2CA3AA1 (0060001D330, 00000000000, 00000000002, 00600015FD0)
00000000005  0010042676F (00600057500, 000FFFFFFFF, 00000000000, 00700000006)
[...]

PR #407, opened 2015-12-19
Fixed Tmux some issue on msys2

moguriso:
This fixes issue #376


moguriso:
sorry, this patch is merged in upstream source.

PR #417, opened 2016-01-09
* Disable debug messages. They are annoying for normal use.

waterlan:
* Disable debug messages. They are annoying for normal use.
* Removed patch file, adding make options in PKGBUILD.


waterlan:
Hi,
I don't understand why you added STRIP=true. As long as you don't call "make strip" there is no stripping done, it is not part of the normal "all" target.
I think that STRIP=true can be removed. I doesn't hurt, it can be removed in the next release.

mingwandroid:
Thanks!

DavidEGrayson:
Thanks, dos2unix did seem unusually chatty for a Unix utility when I used it!

PR #423, opened 2016-01-20
cyrus-sasl: Remove annoying debug message

nalla:
Every time an svn call is made, the message "DIGEST MD5 common mech free"
appears in the application event log viewer. Other communities (See
https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1372451.html)
are supressing the message with a logcheck snippet. Since we do not have
such a luxury in the msys2 environment, lets just remove the debug message
for good.

This fixes https://github.com/git-for-windows/git/issues/616

Signed-off-by: 마누엘 <nalla@hamal.uberspace.de>

### DIFF
--- a/cyrus-sasl/09-digest-md5-Remove-annoying-debug-message.patch
+++ b/cyrus-sasl/09-digest-md5-Remove-annoying-debug-message.patch
@@ -1,0 +1,37 @@
+From 37494bedb64dda4fb6621aa8a75e012d78ef4488 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?=EB=A7=88=EB=88=84=EC=97=98?= <nalla@hamal.uberspace.de>
+Date: Wed, 20 Jan 2016 07:22:10 +0100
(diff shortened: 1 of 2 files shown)


nalla:
@Alexpux any thoughts about this PR?

mingwandroid:
Removing debug messages is generally wrong, IMHO, SASL_LOG_DEBUG shouldn't get seen in the application event log view unless SASL has been build in debug mode, and this is the problem that should be fixed here.

Can you please investigate that instead?

nalla:
@mingwandroid unless it is in https://github.com/Alexpux/MSYS2-packages/blob/master/cyrus-sasl/PKGBUILD#L85 `$CFLAGS` I do not see how the `-DDEBUG` could have been set. At least in my setup `$CFLAGS` only contains `-march=x86-64 -mtune=generic -O2 -pipe`. So I believe that the warning is not dependent of `-DDEBUG` being set.

mingwandroid:
It isn't but I think it should be. However I'll merge it anyway.

PR #431, opened 2016-02-02
Add drone build script.

fracting:
Hi, I've setup a Wine based Drone CI server at wine-ci.org, and created build script for MSYS2-packages, could you accept this script and register your account at wine-ci.org, so I can have real data to test the Wine CI server?

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,14 @@
+build:
+  image: fracting/msys32
+  shell: msys32
+  commands:
+    - drone/build.sh
+
+notify:
+  irc:
+    prefix: build
+    nick: MSYS2-packages
+    channel: msys2
+    server:
+      host: irc.oftc.net
+      port: 6667
--- a/drone/build.sh
+++ b/drone/build.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# fetch first changed file, assume at most one package touched per commit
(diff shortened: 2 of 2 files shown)


fracting:
BTW, with our own CI server, there is 50min of time limitation like Appveyor does, instead, we can build as long as 900 mins, configurable at Web UI in repo "settings".

fracting:
Here is an example how the build script will work:

http://wine-ci.org/fracting/MSYS2-packages/2

Everytime when a package is touched and the commit is pushed, the CI server will clone source code from github, cd to that package directory, and call `makepkg` to build the package.

We can integrate continuous deployment in the future.
We can invent some way to trigger build by upstream change using the drone-downstream plugin in the future.

fracting:
BTW, please ignore any 'permission error' when activate an repo in wine-ci.org, i'll try to fix that error in the drone ci server later, but that error should not prevent trigger CI builds.

The 'permission error' is a unknown (but safe) side effect when I tried to patch my own Drone CI server to downgrade github permission, see https://github.com/drone/drone/issues/1451 for more details.

fracting:
Hi @Alexpux 

Are you able to login to http://wine-ci.org/Alexpux/MSYS2-packages and http://wine-ci.org/Alexpux/MINGW-packages, then click the 'ACTIVATE NOW' button?

Thanks!

Alexpux:
done

PR #434, opened 2016-02-03
Introduce matrix build in Drone CI. Fix makedepends of pacman.

fracting:
In this pull request, I add matrix build to .drone.yml:

```
matrix:
  runtest:
    - false
    - true # allow failures
```

Drone CI will run two build for every commit, one without `make check`, the other with `make check` but ignore the test errors.

Example: http://wine-ci.org/fracting/MSYS2-packages/16

In a long term, I'll try to maintain a white list / black list, incrementally force `make check` for more and more packages, push more packages to pass their test suite (on both Windows and Wine).

(BTW, wine-ci.org is slow, with only one vCPU, I'll upgrade it in a few days)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -2,6 +2,7 @@ build:
   image: fracting/msys32
   shell: msys32
   commands:
+    - export RUNTEST=$$runtest
     - drone/build.sh
 
 notify:
@@ -12,3 +13,8 @@ notify:
     server:
       host: irc.oftc.net
       port: 6667
+
+matrix:
+  runtest:
+    - false
+    - true # allow failures
--- a/drone/build.sh
+++ b/drone/build.sh
@@ -6,7 +6,7 @@ git config --global user.email "ci@msys2.org"
 git config --global user.name "MSYS2 Build Bot"
 
 # fetch first changed file, assume at most one package touched per commit
-TOUCHED=`git show --pretty="format:" --name-only | grep . | head -1`
+TOUCHED=`git show -m --pretty="format:" --name-only | grep . | head -1`
 PKGDIR=`dirname $TOUCHED`
 if [ "$PKGDIR" = "." ]
 then
--- a/drone/build.sh
+++ b/drone/build.sh
@@ -13,6 +13,11 @@ then
     echo Nothing to test
 else
     pushd $PKGDIR > /dev/null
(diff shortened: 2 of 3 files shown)


fracting:
The build fail could be safely ignored, what we need is to address https://github.com/Alexpux/MSYS2-packages/issues/433, with some feedback I can provide a solution soon :)

mingwandroid:
Does this mean we get good CI-on-Windows as well as good CI-on-Windows-on-Wine?

fracting:
No, Wine is not good enough, there are a couple of things to improve. Here is a summary:

1. If you don't care about `make check`, then Wine is almost as good as Windows for building msys2 packages. In my local testing, I've successfully build 300+ msys2 packages without `make check`. However, if you run the testsuites, then there are a lot of false negative. Lots to do in the Wine side.

2. While Wine is good enough to build 300+ msys2 packages, it is not good enough to build mingw packages yet, I'm still working on that.

3. The performance of Wine is still not good enough.

4. Wine still doesn't support 64bit MSYS2 yet, it's on the todo list, but no progress yet.

5. However, our Wine CI has no 50 mins limitation (compare to Appveyor CI), we can build at most 900 mins for a task.

Welcome to read / edit our todo list :)
https://github.com/fracting/wine-fracting/wiki/MSYS2-on-Wine#to-do

fracting:
Hi @Alexpux or @mingwandroid , could you merge this pull request to see if my improvement work as expect ( for future coming pull request) ?

Thanks!

mingwandroid:
Sure thing!

fracting:
Thanks! After this patch set, I rerun a build for https://github.com/Alexpux/MSYS2-packages/pull/435, now it seems the github pull request system starts to work with our CI system!

http://wine-ci.org/Alexpux/MSYS2-packages/6

mingwandroid:
What was involved in hooking up the Wine CI system to github? I am very interested in this? Is there some documentation you can point me at?

fracting:
@mingwandroid sure, that's on my todo list. It's a bit complicate, I think i'll have to split it to multiple parts. I'll notify you when I have update :)

fracting:
@mingwandroid btw, before I document all the Wine specific hack, you can also have an overview at http://readme.drone.io/usage/overview/

I'll document the Wine specific hack maybe next week. :)

PR #455, opened 2016-02-20
znc-git: New package

mattn:
made package for [znc](http://znc.in)

This will be useful to use this with irssi-git package.

### DIFF
--- a/znc-git/PKGBUILD
+++ b/znc-git/PKGBUILD
@@ -1,0 +1,46 @@
+# Maintainer: Yasuhiro MATSUMOTO <mattn.jp@gmail.com>
+
+_pkgname=znc
+pkgname=${_pkgname}-git
+pkgver=8f2c17c3099623b7165c7624c3020fc6242196c1
+pkgrel=1
+pkgdesc='An IRC bouncer with modules & scripts support'
+url='http://znc.in/'
+license=('Apache 2.0')
+arch=('i686' 'x86_64')
+provides=('znc')
+conflicts=('znc')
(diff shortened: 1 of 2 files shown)


mattn:
Hmm, strange. In my environment, it works.

elieux:
@mattn, maybe `autoreconf -fi` is needed?

elieux:
Okay, so that didn't help, but replacing autoreconf with `./autogen.sh` seems to get configure past the install.sh check. Can you try this change in your environment (delete `src` before running makepkg) and amend the pull request if it succeeds?

fracting:
@mattn Thanks for your contribution!

In order to confirm there is no false positive on the Wine CI server, I manually test the package, and found two makedepends() are missing: 'gettext-devel' and 'libsasl-devel'

(-lintl is needed when checking for Py_Initialize, maybe we need to add -lintl check in configure script as well?)

After fixing that, znc starts to build:
http://wine-ci.org/fracting/MSYS2-packages/71

mattn:
@fracting thank you

fracting:
@mattn could you squash three commits and `push --force`?

fracting:
BTW, the Appveyor CI build error could be ignore in this case if I remember correctly.

@renatosilva Do you know if that's a known issue? What's the right way to solve that?
```
./configure: line 557: 0: Bad file descriptor
```
```
 557 test -n "$DJDIR" || exec 7<&0 </dev/null
```

mattn:
squashed.

PR #459, opened 2016-02-23
Fix for spaced directory names

usta:
I'm using this script with cd $OLDPWD , on the other hand whenever i try
to open a new bash console with spaces it couldn't initialize a new
console whenever if the folder/directory has a space in its path.
Because current code gets the_new_dir with $1 and system gets only first
part. So we need to change the_new_dir as $\* to get whole arguments .


elieux:
I have not tried this code, but the change from `$1` to `$*` doesn't look very okay. Does that mean that this `cd_func` is invoked with the directory name not inside quotes, i.e. as multiple arguments? Even if that would work correctly (which IMO it doesn't -- try directories containing multiple spaces in sequence), why not just correctly use quotes?

usta:
I'm using "Open Console Here" from windows registry :

[HKEY_CLASSES_ROOT\Directory\Background\shell\bash]
@="Open Console Here"
"NoWorkingDirectory"=""

[HKEY_CLASSES_ROOT\Directory\Background\shell\bash\command]
@="D:\msys64\mingw64_shell.bat %"

And in my .bashrc at the last line i added cd $OLDPWD

so whenever i didn't change $1 -> $* it doesn't work with cd_func

maybe we need to add an option in mingw64_shell.bat like --cd "%"

fracting:
Hi @usta 

Next time when you create a pull request changing a package, please also consider increasing the number of pkgrel in PKGBUILD, if your patch changes non trivial behavior, we hope a CI rebuild is  triggered. Our ci-build.sh script detects PKGBUILD changing in order to decide which package to rebuild.

Thanks!

usta:
@fracting sorry about it. I didn't know about PKGBUILD thing before. I have fixed PR. Could you check now? I have seperated .bashrc and .bash_functions

fracting:
@usta No sorry at all, thanks for the contribution! Leave the pr to @elieux or someone else to review, I'm not the right person. Cheers :)

elieux:
What if you put `cd "$OLDPWD"` in your `.bashrc` instead?

forgottenswitch:
```
$ bash -c 'f() { x=$1; echo "$x"; }; f before:_a b              c; f "instead:a b c"'
before:_a
instead:a b c
```

forgottenswitch:
```
$ bash -c 'f() { x=$1; echo "$x"; }; g() { x=$*; echo "$x"; }; f before:a b c; g after: a  b   c; f "instead: a b c";'
before:a
after: a b c
instead: a b c
```

forgottenswitch:
```
$ bash -c 'f() { x=$1; echo "$x"; }; g() { x=$*; echo "$x"; };
>  f before:a b c; g after: a  b   c; f "instead: a b c";'
before:a
after: a b c
instead: a b c
````

lazka:
No response for years, so closing.

PR #460, opened 2016-02-23
Improvements for continuous integration

renatosilva:
* New implementation for detection of changed recipes.
* Merge dependency installation and actual build into one single makepkg call.
* Use `--cleanbuild` and `--rmdeps` in makepkg call.
* Source code refactoring.


elieux:
Won't the `--rmdeps` cause unnecessary double installation of the package's runtime dependencies?

PR #468, opened 2016-02-27
move most remaining packages to sha256, url updates

vszakats:
* plus some URL updates
* referenced `cygrunsrv/cygrunsrv-1.61.tar.bz2` is missing from the repository, I'm not sure if this is intentional.

With this commit, there are no more `md5sums` in this repository.


vszakats:
In `binutils` this patch only corrected two broken, but _commented_ urls, so the breakage in AppVeyor seems unrelated to this commit, while the same build went fine in Drone.

With `cygrunsrv`, again two _commented_ URLs changed, plus the hashes. The hashed `.bz2` file can be found _inside_ the `.xz` found under the updated `mirrorservice.org` URL [*], so I could rechecksum, but the `mksource` logic is and was missing, so no wonder the file is not found by CI.

Should I revert these regardless?

[*]
```
drwxr-xr-x  0 corinna vinschen    0 Aug 21  2014 cygrunsrv-1.61-1.src/
-rw-r--r--  0 corinna vinschen  297 Aug 21  2014 cygrunsrv-1.61-1.src/cygrunsrv.cygport
-rw-r--r--  0 corinna vinschen 90875 Aug 21  2014 cygrunsrv-1.61-1.src/cygrunsrv-1.61.tar.bz2
Archive Format: GNU tar format,  Compression: none
```

vszakats:
Let me resubmit these in a cleaner way. Closing.

PR #477, opened 2016-03-01
added patch to enable Windows SSPI authentication for msys

nightman68:
This patch enables SSPI - the Windows integrated authentication to the serf library. As result serf is using the users Kerberos TGT from the MS LSA for authentication against a kerberised Subversion server.

@dscho created the patch  originally, have a look to the [conversation](https://github.com/git-for-windows/git/issues/550).


nightman68:
I forgot to merge your changes.

Issue #482, opened 2016-03-03
python2 disable ssl


mattn:
![](http://go-gyazo.appspot.com/c9c59c316ccf1b35.png)

Alexpux:
Fixed packages uploaded.

PR #487, opened 2016-03-04
python2: Remove 0030-ssl-threads.patch

mattn:
Maybe fix #390 #482

I could import ssl module.


Alexpux:
You need just rebuild python without removing patches

mattn:
Why?, I guess that the cause of issue is using WITH_THREAD.

mattn:
Just tried build without this patch. And confirmed fixes.

PR #491, opened 2016-03-08
Add ag to MSYS2-packages because isatty() doesn't work on mintty seamlessly

mattn:
ag built on mingw hang on mintty because isatty(fileno(stdin)) always return 1 on mintty. (I know it should work with winpty) So I suggest to add ag into MSYS2-package too.

### DIFF
--- a/ag/PKGBUILD
+++ b/ag/PKGBUILD
@@ -1,0 +1,52 @@
+# Maintainer: Yasuhiro Matsumoto <mattn.jp@gmail.com>
+
+pkgname=ag
+pkgver=0.31.0.r1696.edae22e
+pkgrel=1
+pkgdesc="The Silver Searcher: An attempt to make something better than ack, which itself is better than grep (mingw-w64)"
+arch=('any')
+url="http://geoff.greer.fm/ag"
+license=("Apache")
+makedepends=("pcre-devel"
+             "liblzma-devel"
+             "zlib-devel")
+depends=("pcre"
+         "xz"
+         "zlib")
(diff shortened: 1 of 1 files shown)


mattn:
will fix

mingwandroid:
Pass . as final parameter to mingw64 ag to work around this. We definitely should not add packages to msys2 repo for something like this. The point is to make mingw-w64 software, not msys2 software!

mattn:
Why not? For example, MINGW-packages/mingw-w64-ruby and MSYS2-packages/ruby both are exists in each repos. And I believe that ag is software for *NIX.

mingwandroid:
Because MSYS2 is about native software for Windows! There are specific reasons why ruby needs to exist for both (because it is needed to help build native software).

We are not trying to be or to replace cygwin.

ag is all about very fast file searching. The cygwin overhead removed that benefit immediately.

I already gave you a satisfactory workaround (add a .). If it is not satisfactory explain in what way this is so and we can try to address that.

Please understand that the M in MSYS2 stands for minimal. We do not want to add msys2 packages without very good reason.

mattn:
Okay, but many people doesn't know that hack you gave me (add a .). :)

mingwandroid:
I know but every time we add 2 versions of any package it leads to confusion.

We must resist adding msys2 packages.

Microsoft and cygwin have discussed this exact problem on the cygwin mailing list FWIW.

mattn:
Then, how about to patch mingw-w64-ag to be possible to handle stdin correctly even if it run on mintty?

PR #497, opened 2016-03-11
remove .orig file from msys2 patch

staalmannen:
Sorry for all the mess. Stuff built locally but I noticed issues in the automated build testing when committing. I *hope* this is the last pull request I have to do for these packages :)


Alexpux:
I'm already have this fixed in local tree.

staalmannen:
OK thanks!

PR #506, opened 2016-03-22
autoconf: Add perl to depends.

fracting:


Alexpux:
Increase pkgrel also

fracting:
Thanks for review, updated.

PR #531, opened 2016-04-04
.bat-files: exit only from .bat file instead of terminating parent command processor

Karlson2k:
bypass result of "start" to calling process.

Currently if Msys is started from another .bat file by `call path/mingw64_shell.bat`, those .bat file is terminated at this command. To exit only from .bat file instead of terminating cmd.exe, `exit /b` must be used.


Karlson2k:
Required to be started by `cmd.exe` with flag `/C` (default flag for starting .bat files)
As installer used flag `/K` cmd terminal wasn't closed after MSys shell lanched: #537

PR #534, opened 2016-04-04
Some improvement of .bat-files plus detection of ConEmu

Karlson2k:
Should make ConEmu users happy

### DIFF
--- a/filesystem/mingw32_shell.bat
+++ b/filesystem/mingw32_shell.bat
@@ -1,9 +1,9 @@
 :
 @echo off
 
-if NOT "x%WD%" == "x" set WD=
+set "WD=%__CD__%
 
-if NOT EXIST %WD%msys-2.0.dll set WD=%~dp0usr\bin\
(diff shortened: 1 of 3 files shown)


Alexpux:
Seems by default you try to use conemu. Don't do it

Karlson2k:
ConEmu will be used only if it is install **and** configured as default terminal (or automatic attach is used).
If it's not acceptable, I'll implement another scheme.

Хотя мне кажется, что те, кто установил ConEmu и используют его по-умолчанию - ожидают именно такого поведения.

Alexpux:
By default always mintty is used, no matter if ConEmu is installed.

Karlson2k:
OK

Karlson2k:
@Alexpux Should I update sha256sums in PKGBUILD?
Seems that I missed that in previous commits.

Alexpux:
Yes, run "updpkgsums" and change pkgver to "2016.04" and pkgrel to 1

Karlson2k:
@Alexpux Thanks, will do.

Karlson2k:
Better solution: PR #538

PR #538, opened 2016-04-05
.bat-files: added support for ConEmu terminal emulation

Karlson2k:
ConEmu used only if requested by `-conemu` parameter.


Alexpux:
See also https://github.com/Alexpux/MSYS2-packages/issues/537

PR #540, opened 2016-04-07
More consistent Windows path setting

renatosilva:
The SET_FULL_PATH variable has been replaced with the more specific MSYS2_PATH_TYPE, with the following possible values:

* *inherit*, for instructing /etc/profile to reuse the existing path, normally the Windows path.
* *strict* for disabling inheritance and allowing custom definition of the Windows path if needed.
* Other values for enabling default behavior, currently a minimal Windows path.

Test case in batch script:

```batch
bash -c "echo $PATH"
bash --login -c "echo $PATH"

set MSYS2_PATH_TYPE=inherit & bash --login -c "echo $PATH"
set MSYS2_PATH_TYPE=strict  & bash --login -c "echo $PATH"
```


filesystem/profile
```diff
@@ -24,41 +24,42 @@
...
-if [ -n "$SET_FULL_PATH" ]
-then
-  WIN_PATH="${PATH}"
-else
-  WIN_ROOT="$(PATH=${MSYS2_PATH} cygpath -Wu)"
-  WIN_PATH="${WIN_ROOT}/System32:${WIN_ROOT}:${WIN_ROOT}/System32/Wbem:${WIN_ROOT}/System32/WindowsPowerShell/v1.0/"
-fi
+
+case "${MSYS}" in
+  *winpath:none*) unset WIN_PATH ;;
```
Alexpux: Really don't do it. Empty windows path will lead to not working msys2 as you will not have system libraries in the PATH

Alexpux:
I think I will not accept it.
First, we already have MSYS variable set in bat files for other purposes (See commented lines) so need handle situation when some lines are uncommented with different MSYS values.
Second full path is not controlled via msys2-runtime and we don't need to be consistent with cygwin way

renatosilva:
The MSYS option has been replaced with just a more specific variable MSYS2_PATH_TYPE. Cygwin actually defines a similar variable CYGWIN_NOWINPATH instead of having implemented it as a CYGWIN option. The *strict* value allows for more control over the path construction in special cases.

PR #558, opened 2016-04-17
ccache: add MSYS package

mati865:
Ccache for MSYS2.
Thanks to it recompilation with makepkg is much faster also allows to speed up manual recompilation after adding /usr/lib/ccache/bin/ to PATH (`export PATH="/usr/lib/ccache/bin/:$PATH"`).

Normally there would be symlinks to ccache inside /usr/lib/ccache/bin but they don't work out of box in MSYS2. This workaround creates bash scripts executing ccache with proper command.
If someone knows better solution to do it I'll change it. Also if this workaround is acceptable I'll do similar to MinGW ccache package.

EDIT: By the way could some restart tea-ci?
It freeze with `err:seh:setup_exception_record stack overflow 768 bytes in thread 017b eip 611244f9 esp 0060a9e0 stack 0x410000-0x60a9b0-0x610000` in 2 latest commits.


Alexpux:
Lets do the same hack for mingw cache

PR #560, opened 2016-04-18
added perl-Date-Calc perl-Encode-compat perl-ExtUtils-MakeMaker perl-…

smartmarmot:
…List-MoreUtils perl-Math-Int64 perl-Test-Harness perl-Carp-Clan

### DIFF
--- a/perl-CPAN-Meta/PKGBUILD
+++ b/perl-CPAN-Meta/PKGBUILD
@@ -1,0 +1,36 @@
+# Maintainer: Andrea Dalle Vacche
+
+_realname=CPAN-Meta
+pkgname=perl-${_realname}
+pkgver=2.150005
+pkgrel=1
+pkgdesc="the distribution metadata for a CPAN dist"
+arch=('any')
+license=('GPL' 'PerlArtistic')
+url="http://search.cpan.org/~dagolden/CPAN-Meta-2.150005/"
+depends=('perl')
+groups=('perl-modules')
+options=('!emptydirs')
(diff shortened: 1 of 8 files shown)


smartmarmot:
I think there is a wrong configuration serverside as the perlapi./h missing is available within:
$ pacman -Qo /usr/lib/perl5/core_perl/CORE/perlapi.h
/usr/lib/perl5/core_perl/CORE/perlapi.h is owned by perl 5.22.1-1

mati865:
Don't worry about the bots, they are buggy.
If it does compile in fresh msys2 everything is ok.

Alexpux:
I see issues with building List-MoreUtils and Math-Int4. Also CPAN-Meta not needed as it part of perl package.

smartmarmot:
Thank you for the feedback, as far I can see the Math-Int64 require for testing the ExtUtils::MakeMaker  . I think the error is due to that, I'll fix the PKGBUILD adding the depends=('perl-ExtUtils-MakeMaker').
I will check the issue with the List-MoreUtils module.

Alexpux:
Errors is due to linking to perl library

smartmarmot:
Ok then I need to double-check can you remove just the broken packages?

smartmarmot:
otherwise please revert the merge I don't want to block you with this issue

Alexpux:
I'm upload packages that build successfully, others need to fix

smartmarmot:
pefect thanks!

PR #564, opened 2016-04-18
autoconf-archive: new package.

KrullKorg:


Alexpux:
Why it needed?

KrullKorg:
for example it is needed by template-glib https://git.gnome.org/browse/template-glib/ that use AX_CHECK_COMPILE_FLAG

Alexpux:
Well seems this package conflicts with gnome-common

PR #573, opened 2016-04-25
Some improvements for ccache

### LINKED ISSUE #562: makepkg-mingw ccache build option doesn't work
makepkg-mingw executes makepkg with makepkg_mingwXY.config:
```
  if [ -f "/${_mingw}/bin/gcc.exe" ]; then
    MSYSTEM=${_msystem} \
    PATH=/${_mingw}/bin:$(echo $PATH | tr ':' '\n' | awk '$0 != "/opt/bin"' | paste -sd:) \
    /usr/bin/makepkg --config /etc/makepkg_${_mingw}.conf $@ || exit 1
```

default config:
`BUILDENV=(!distcc color !ccache check !sign)`
If we uncomment `!ccache` to `ccache` makepkg will run following:
```
	# use ccache if it is requested (check buildenv and PKGBUILD opts)
	if check_buildoption "ccache" "y" && [[ -d /usr/lib/ccache/bin ]]; then
		export PATH="/usr/lib/ccache/bin:$PATH"
		ccache=1
	fi
```
This not only makes ccache useless for makepkg-mingw but also modifies path set by makepkg-mingw (gcc in /usr/bin will be first choice instead of /mingwXY/bin).

My suggestion is to change `BUILDENV=(!distcc color !ccache check !sign)` like this `BUILDENV=(!distcc color !ccache_mingw check !sign)` and add proper check in makepkg:
```
	# use ccache_mingw if it is requested (check buildenv and PKGBUILD opts)
	if check_buildoption "ccache_mingw" "y" && [[ -d $MINGW_PREFIX/lib/ccache/bin ]]; then
		export PATH="/$MINGW_PREFIX/lib/ccache/bin:$PATH"
		ccache=1
	fi
```

renatosilva:
The compilers available from system path are now used, and the wrapper scripts have been replaced with symlinks. Fixes #562 and removes the need for mingw-w64-ccache.

* Support both MINGW and MSYS toolchains.
* Fix ccache option for makepkg-mingw.
* Update to 3.2.5.

### DIFF
--- a/ccache/ccache.install
+++ b/ccache/ccache.install
@@ -1,0 +1,10 @@
+post_install() {
+  mkdir -p /usr/lib/ccache/bin
+  for compiler in c++ cc cpp gcc g++ {i686,x86_64}-{pc-msys,w64-mingw32}-{g++,c++,gcc}; do
+    MSYS='winsymlinks:lnk' ln -sf /usr/bin/ccache /usr/lib/ccache/bin/${compiler}
+  done
+}
+
+pre_remove()   { rm -rf /usr/lib/ccache; }
+pre_upgrade()  { pre_remove; }
+post_upgrade() { post_install; }
--- a/ccache/PKGBUILD
+++ b/ccache/PKGBUILD
@@ -1,7 +1,8 @@
 # Maintainer: Mateusz Mikuła <mati865@gmail.com>
+# Contributor: Renato Silva <br.renatosilva@gmail.com>
 
 pkgname=ccache
-pkgver=3.2.4
+pkgver=3.2.5
 pkgrel=1
 pkgdesc="A compiler cache (mingw-w64)"
 arch=('i686' 'x86_64')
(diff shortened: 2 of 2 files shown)


ccache/ccache.install
```diff
@@ -0,0 +1,10 @@
+post_install() {
+  mkdir -p /usr/lib/ccache/bin
+  for compiler in c++ cc cpp gcc g++ {i686,x86_64}-{pc-msys,w64-mingw32}-{g++,c++,gcc}; do
```
Alexpux: don't see mingw-cpp and mingw-cc symlink

elieux:
Just to be clear, this won't work from cmd (or using mingw32-make), right?

mati865:
Looks promising I'll test it when I'm back home.

ccache/ccache.install
mati865: @Alexpux there is no need for it.

mati865:
This is very good solution and I like it however MinGW cmake still isn't using it.
Why? I think MinGW cmake looks only for `gcc`, `gcc.exe` and `g++`, `g++.exe`. But in `/usr/lib/ccache/bin` there are `gcc.lnk` and `g++.lnk` so it looks next directory in $PATH which is `/mingw{32,64}/bin` containing `gcc.exe` and `g++.exe`. If we want it to work with ccache we should do as @mingwandroid suggested in https://github.com/Alexpux/MSYS2-packages/issues/562 and patch MinGW cmake.

You can also update ccache to 3.2.5 and I would close https://github.com/Alexpux/MSYS2-packages/pull/571 then.

renatosilva:
There may be issues not just with CMake but any tool that have problems recognizing non-executable replacements. The wrapper scripts can be restored later if they demonstrate to generally behave better, or the symlinks can be replaced with actual executables. For now, this patch addresses the reported issue and adds MINGW support.

mati865:
I initially created wrapper scripts to get any compatibility (didn't know about `winsymlinks`). Your fix is way better: reduced overhead, cmake works with non-cached GCC instead being totally broken and finally ccache option is fixed very nicely.
And for MinGW cmake we either need patch for cmake or additional `.exe` wrappers (native, [SHC](http://opensourcepack.blogspot.com/2013/04/shc-389-msys-port.html) doesn't seem like proper solution) but it don't have to be resolved just now.

I hope it gets merged soon. In future if we prove there is no conflict between MSYS2 and MinGW ccache (by default all cached files are stored in /home/<user>/.ccache, which could potentially cause compile errors because of differences in compilers) we could add MinGW ccache to `replaces=()` in MSYS2 ccache.

PR #575, opened 2016-04-27
New MSYS package: task

Green0Photon:
This is a command line productivity tool also called [Taskwarrior](https://taskwarrior.org/). It is a highly flexible open source todo list.

It has no support for Windows, so it's on Cygwin.

I'm not an official maintainer or anything, but a user who wanted to use taskwarrior on MSYS2 rather than Cygwin. Dependencies were tested on my MSYS2-x86_64. I borrowed pieces from the official Arch package and the git AUR package, in addition to adding testing.

There are two oddities: the official tar has broken symlinks which causes makepkg to fail, so I download the git repo and switch to the tag instead. In addition, task originally crashes when run, because libgnutls did not install libhogweed, so that is included as a dependency.

The tests has a few failures (out of roughly a thousand), but I suspect those are normal. I can install a linux vm to compare, if you want, but it works. After skimming through the tests, the failure seemed inconsequential, or even false failures. I have the check function set to succeed regardless.

I also plan on making packages for the other taskwarrior tools (taskd, vit, and tasksh). They will be (a) separate pull request(s). I will probably put them all together in one pull request instead, but I want to get this into repo. If you don't pull this by the time I make the other packages, I'll merge them into this pull request.

Thanks!

### DIFF
--- a/task/PKGBUILD
+++ b/task/PKGBUILD
@@ -1,0 +1,90 @@
+# Maintainer: Green0Photon <green0photon AT gmail DOT com>
+
+pkgname=task
+pkgver=2.5.1
+pkgrel=1
+pkgdesc="A command-line todo list manager"
+arch=('i686' 'x86_64')
+url="https://taskwarrior.org/"
+license=("MIT")
+
+depends=("gcc-libs"
+         "libgnutls"
+         "libutil-linux"
+         "libhogweed" # for some reason, didn't install with libgnutls
+)
+
+makedepends=("git" # 'msys2-w32api-headers>=5.0.0.4624' 'msys2-w32api-runtime'
+             "gcc"
+             "make"
+             "cmake"
+             "libgnutls-devel"
+             "libutil-linux-devel"
+)
+
+# AUR packages had an odd colon notation
+# The comments were within the quotes so pacman knew about it
+optdepends=("bash-completion" # for bash completion
+            "python" # for python export addon
+            "ruby" # for ruby export addon
+            "perl" # for perl export addon
+            "perl-json" # for perl export addon
+            "rsync" # synchronisation via rsync
+            "curl" # synchronisation via http(s)/ftp
+            "openssh" # synchronisation via ssh
+)
+
+checkdepends=("python" # for tests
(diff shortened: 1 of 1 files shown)


Green0Photon:
I am working on a fix right now.

Green0Photon:
BTW AppVeyor has been failing but tea has been succeeding. The files appveyor says are missing are `/usr/include/sys/direct.h` and maybe `/usr/include/sys/unistd.h` or `/usr/include/cygwin/version.h`. I just brand new downloaded the base image and all three are there. I'm trying to see if appveyor has it.

Green0Photon:
Yeah, all are in `msys2-runtime-devel` and exist in appveyor. And yet, appveyor isn't working.

I know this is a pull request, but some help would be great. I don't know if this is a problem with appveyor. I think it is because tea and my own system work, but I don't know.

Green0Photon:
Alright.

Since tea works and appveyor doesn't, here are some comparisons.

- `gcc` has the same version (4.9.2-6).
- `msys2-w32api-runtime` has a higher version (4624) on tea and a higher version (4574) on veyor. It has an even higher version on my computer (4633). This is likely the problem and what I'm trying to constrain now.

Sorry for clogging the build logs on both of them. Is it more or less annoying that I've been doing amends the entire time?

Green0Photon:
It's either a version problem in a package, or something else outside of a package, which I don't think you do.

Sorry again for clogging your ci systems. :frowning:

mati865:
AppVeyor bug. It has similar problem with mpv not detecting libass.

Green0Photon:
I'm just gonna stop it with trying to fix appveyor, which I've clogged up (accidental hanging pacman + massive amounts of amends to get things to work on both ci). The problems might be because it doesn't get updated each time, unlike tea.

It might be good to find a way to stop clogging it. (And does @Alexpux have a plunger aka delete buttons?) There were two reasons my tea builds failed: I didn't delete **everything** for my last build on my computer, and breaking tea in an attempt to fix appveyor.

Sorry, for the last time, for clogging everything.

PR #577, opened 2016-04-28
Add new MSYS package: taskd and Update task

Green0Photon:
This primarily adds taskd, the task server. It also adds some quick improvements to task.

Unfortunately, I don't know how MSYS daemons are set up, so I can add those into it, if someone can point me to the information.

It does mean that users can easily get a compiled version of it, though, even if they need to write the scripts themselves.

That should otherwise be all from me regarding taskwarrior. :smile:

### DIFF
--- a/task/PKGBUILD
+++ b/task/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=task
 pkgver=2.5.1
-pkgrel=1
+pkgrel=2
 pkgdesc="A command-line todo list manager"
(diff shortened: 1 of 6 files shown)


Green0Photon:
Again appveyor failed. It thus didn't do taskd, so we don't know if it'd work in that.

Alexpux:
MSYS2 not Cygwin. We not provide environment for daemons. So I don't want to see daemon packages in MSYS2.

Issue #130, opened 2014-11-23
Start mintty bash in current working directory


Konfekt:
How about starting mintty in the current working directory instead of the home dir?

This is convenient for spawning msys2 for example in a GIT repo from the cmd line or external programs such as (g)Vim.

This can be achieved by modifying the following line in `c:\msys64\msys2_shell.bat` 

```
start %WD%mintty -i /msys2.ico /usr/bin/env CHERE_INVOKING=1 /usr/bin/bash --login %*
```

to

```
start %WD%mintty -i /msys2.ico /usr/bin/bash --login %*
```

If this is to intrusive, there are two other possibilities:
- How about pulling %\* forward, that is

```
start %WD%mintty -i /msys2.ico %* /usr/bin/bash --login
```

instead of

```
start %WD%mintty -i /msys2.ico /usr/bin/bash --login %*
```

so that the user can pass on /usr/bin/env CHERE_INVOKING=1 to msys2?
- How about an argument, say `-here`, that can be passed to `msys2_shell.bat` and then only in this case spawns `msys2_shell.bat` in the current working directory?

elieux:
Both `msys2-launcher` launchers and `msys2_shell.cmd` support starting in current directory.

PR #579, opened 2016-04-28
Add -where parameter to start_shell.cmd

ph4nt0m:
When passing the optional _-here_ argument to start_shell.cmd, mintty will be started in the current working directory instead of the home directory.

This should be a first step towards being able to conveniently spawn a new MSYS2 shell from within a folder of the Windows Explorer. The MSYS2 installer could offer the user to add a respective entry to the context menu, for example.

### DIFF
--- a/filesystem/start_shell.cmd
+++ b/filesystem/start_shell.cmd
@@ -1,4 +1,5 @@
 @echo off
+setlocal
 
 set "WD=%__CD__%"
 if NOT EXIST "%WD%msys-2.0.dll" set "WD=%~dp0usr\bin\"
@@ -29,6 +30,15 @@ if "x%~1" == "x-conemu" shift& set MSYSCON=conemu& goto :checkparams
 if "x%~1" == "x-defterm" shift& set MSYSCON=defterm& goto :checkparams
 rem Other parameters
 if "x%~1" == "x-use-full-path" shift& set MSYS2_PATH_TYPE=inherit& goto :checkparams
+if "x%~1" == "x-where" shift& (
+  set CHERE_INVOKING=1
+  rem Check if next argument is an existing directory
+  if not "%~2" == "" if exist "%~2\" shift& cd /d "%~2\"& goto :checkparams
+
+  rem If not, print error and exit
+  echo Invalid directory specified for -where. Exiting. 1>&2
+  exit /b 1
+)
 
 
 rem Autodetect shell type if not specified


DavidEGrayson:
I see you are using `/usr/bin/env` to set the `CHERE_INVOKING` environment variable, and this required you to duplicate a long line of code.  I think you should set `CHERE_INVOKING` the same way that `MSYSTEM` is set, so that you don't have to duplicate code and so that the feature will work for terminals other than MinTTY.

ph4nt0m3:
@DavidEGrayson Thanks for your suggestion. I wasn't aware that setting the variable within the batch script would pass it to mintty/bash as well. This makes the modification even more simple, so I pushed a new commit. It seems to work equally well so far, maybe @Karlson2k could have a look as well, as he did most of the recent restructuring of the starting scripts.

Karlson2k:
@ph4nt0m, I think, that merging in current state results in useless addition. Much better to provide a complete solution, so it will be obvious what is improved by PR.

ph4nt0m:
@Karlson2k I agree that the modification is very small, but I would not call it "useless". Surely there must be some users who would benefit from this change (see #130), even if the context menu entry would still have to be created by hand. The thing is that those registry changes remain untouched by MSYS2 updates, in contrast to any changes made to the start_shell.cmd. So not having to fiddle with those startup scripts anymore would definitely be a step forward.

If I knew enough about the install scripts, I would try to implement the registry settings there, but at the moment this seems rather unlikely to be honest.

mingwandroid:
The install scripts are Javascript in msys2-installer. There's a shell script to build a new installer. It's fairly straightforward I reckon.

The only way to find out is to dive in!

Karlson2k:
@ph4nt0m, no need to change this .bat-files if you only need to pass some environment variable.
Just create you .cmd file with:
```` .bat
Set MyVar=MyValue
Call start_shell.cmd
````
When I said "useless" I meant that you just set variable that is checked nowhere in this PR.
So, within this PR, this variable is useless.
If it will be merged in current state and next PRs will not be published, it will stay as rudiment. 
Create a functional solution, it can be merged much faster.

**Comment by DavidEGrayson** [comment]
The `CHERE_INVOKING` environment variable is already checked by MSYS2 (and cygwin) when it starts up and it has the desired effect.  I think this pull request is fine because it adds a useful feature to start_shell.cmd which could be used by people who are creating shortcuts for starting MSYS2 and want it to start in a particular directory.

**Comment by renatosilva** [comment]
I confirm this feature works. But can you please allow for the directory to be optionally specified? So the registry command line can be simplified from this:

```
cmd /c cd /d %V & ROOT\\start_shell.cmd -here
```

To something like this:

```
ROOT\\start_shell.cmd -here "%V"
```

**Comment by Karlson2k** [comment]
If `CHERE_INVOKING` is already used by MSys2/Cygwin then this PR is useful. Didn't know that.
In this case PR is valuable, only commits need to be squashed.
And @renatosilva suggestion is good too, but better to be implemented as `-here-dir` to allow simplified usage with `-here` when current directory is required.
Anyway it's up to @Alexpux.

**Comment by ph4nt0m** [comment]
@renatosilva I pushed a new commit to implement the behaviour you requested.

@Karlson2k Just saw your comment. If everybody agrees that introducing a separate parameter `-here-dir` would be better, I will change this.

**Comment by DavidEGrayson** [comment]
A separate parameter makes more sense to me.  Otherwise you have to use heuristics in the batch file to figure out what type of argument comes after `-here`.  I would call it something like `-cwd <dir>`.

**Comment by ph4nt0m** [comment]
@DavidEGrayson Good point, I changed it, so now we have `-cwd <dir>` and `-here`, where the last one is basically an abbreviation for `-cwd %cd%`.

**Comment by renatosilva** [comment]
No need to create two parameters for the same thing. Please revert to 618dc01, default value does work:

```
start_shell.cmd -here
start_shell.cmd -here C:\Somewhere
start_shell.cmd -here -another-param
```

**Comment by DavidEGrayson** [comment]
If a directory named `-another-param` actually exists, you'll get behavior that you were probably not expecting.

**Comment by renatosilva** [comment]
Then check if parameter starts with hyphen, or use equal sign as separator. Specific situation anyway, if you want to be meticulous you can go read the source code and improve default MSYSTEM detection instead. No need to complicate parameters.

**Comment by Karlson2k** [comment]
Directories can't start with hyphen?

**Comment by Karlson2k** [comment]
Parameters always must start with hyphen?
Can't you pass directory as parameter?

**Comment by renatosilva** [comment]
If you cannot do proper option parsing then maybe make the value mandatory:

```
start_shell.cmd -where .
start_shell.cmd -where C:\directory
start_shell.cmd -where -directory
```

Karlson2k:
This is a better suggestion.

elieux:
I see some bikeshedding has taken place. I would personally prefer something different from what the current state is, but I can live with that.

Anyway, I tested the new parameter and it works. Thanks for the pull request. I give my +1.

@mingwandroid, squash and/or merge?

ph4nt0m:
Ok, I changed everything to what @renatosilva suggested and hopefully got it right after two more commits, sigh (in fact, Git is totally new for me and I'm just about getting familiar with the basic commands).

Please let me know what you think and if there are further things to improve.

mingwandroid:
Ah but the bike shed is nice.

@elieux, I'd say go ahead and squash, it seems you can do this from within github's web interface now.

@ph4nt0m if you beat @elieux to it, the incantations are:

    git rebase -i HEAD~2
    git push origin mintty-chere:mintty-chere

There may be a shorter version if the last line though, but no matter, that'll work. Thanks all.

ph4nt0m:
@mingwandroid Thanks, squashing my last two commits into a single one seemed to work, although I had to use

    git push origin mintty-chere:mintty-chere --force

to force rewriting the history.

mingwandroid:
Ah yes. I forgot, use the --force, ph4nt0m.

mingwandroid:
While we're at it, can you `rebase -i` the whole series? More `rebase` practice :-)

mati865:
Also you can use `export EDITOR=nano` before git commands if you want simpler editor than vim.

ph4nt0m:
Thanks for your support (although I need to learn `vim` as well), should be all good now.

mingwandroid:
One final request from me, I promise (you're getting good with git though :-)), can you make the first line read:

`filesystem: Add -where pameter to start_shell.cmd`

`git commit --amend` is a good approach here.

Karlson2k:
Or just use [TortoiseGit](https://tortoisegit.org) - it will simplify things.

elieux:
The `setlocal` addition has also a nice side effect of not polluting the environment of the calling shell with stuff like `WD=D:\msys64\usr\bin\`.

elieux:
Shoot. We forgot about `PKGBUILD`. Fixed with 8f4825e69ea7d5c040e80398dcb17a2800ac3ece.

PR #582, opened 2016-05-01
Some colorizing tools

renatosilva:
* Colordiff colorizes output of diff and related tools.
* ColorMake colorizes output of make, gcc and similar.

### DIFF
--- a/colordiff/PKGBUILD
+++ b/colordiff/PKGBUILD
@@ -1,0 +1,31 @@
+# Maintainer: Renato Silva <br.renatosilva@gmail.com>
+
+pkgname='colordiff'
+pkgver='1.0.16'
+pkgrel=1
+pkgdesc='Diff wrapper with pretty syntax highlighting'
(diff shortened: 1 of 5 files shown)


colordiff/PKGBUILD
```diff
@@ -0,0 +1,31 @@
+# Maintainer: Renato Silva <br.renatosilva@gmail.com>
+
+pkgname='colordiff'
+pkgver='1.0.16'
+pkgrel=1
+pkgdesc='Diff wrapper with pretty syntax highlighting'
+url='http://www.colordiff.org'
+license=('GPL2')
+arch=(any)
```
Alexpux: This is not mingw package so you can set here appropriate arches

renatosilva: This package does not provide architecture-specific binaries.

Alexpux: ok then

PR #584, opened 2016-05-01
[WIP] cmake: update to 3.5.2

mati865:
It's still WIP but seems to be working. I've modified patches instead of recreating to make comparing easier. When it's ready to merge I'll recreate `cmake-3.2.3-msys.patch` to remove any fuzz.
It was tested by compiling doxygen, jsoncpp, nano. nano also have been manually tested. I know these 3 packages aren't enough to prove cmake is fine. How can I test if cmake is working as expected?

### DIFF
--- a/cmake/3.1.0-cpuinfo.patch
+++ b/cmake/3.1.0-cpuinfo.patch
@@ -44,7 +44,7 @@ index 3e1a1ab..690a40f 100644
 -#else // __CYGWIN__
 -  // does not have "physical id" entries, neither "cpu cores"
 -  // this has to be fixed for hyper-threading.
--  kwsys_stl::string cpucount =
+-  std::string cpucount =
 -    this->ExtractValueFromCpuInfoFile(buffer,"cpu count");
 -  this->NumberOfPhysicalCPU=
 -    this->NumberOfLogicalCPU = atoi(cpucount.c_str());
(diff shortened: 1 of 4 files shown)


mingwandroid:
Build https://github.com/Alexpux/MSYS2-packages/tree/master/clang-svn .. or if that fails and you can't fix it, find other msys2 packages that use CMake. I'd check but I'm not at a computer.

mati865:
Ok, I'll try it tomorrow.

mingwandroid:
It may need some updates unrelated to the CMake version change. I'll try to help out if that's the case.

mati865:
Right now it's compiling. I've had to make some workarounds which will probably break clang.
They can be found here: https://github.com/mati865/MSYS2-packages/commit/0a27fbf074daeb98d7be50202b27ab8652aee676
Note about `0001-cmake-hack-for-msys2.patch`

> For some reason CYGWIN is false under a MSYS Environment when using a msys compiler
> A MSYS2 compiler is essentially cygwin compiler so this should be set to true.
> 
> Before anyone point out cmake's MSYS variable to me.
> The MSYS variable refers to the host system and not the compiler.
> So CYGWIN should still be TRUE.
> 
> This should be fixed in msys-cmake so that we can remove this patch :)
I guess it's `+  #set(CYGWIN 1)` in `cmake/cmake-3.2.3-msys.patch` lines 80 and 100.

mati865:
Build failed at 69% because of my mistake at modifying patches. Fixed patch pushed, I'll recompile it somewhen in following days.

mati865:
Ok I managed to find some time to run compilation, clang, lld and llvm compiled. They are broken and I guess it's because of quick workarounds or some updates to source code. But it's up to you to decide what now @mingwandroid.

PR #587, opened 2016-05-05
Fixes for winpty

renatosilva:
Fix path conversion from unix to Windows format, and a build problem caused by mingw-w64-cross-crt-clang-git. Conversion example:

```
$ winpty cmd //c echo //path is /path
/path is C:/MSYS2/path
```


elieux:
I don't think we should put `mingw-w64-cross-crt-git` inside every package depending on the cross toolchain. This should be solved at the cross CRT package level.

PR #593, opened 2016-05-09
python-pyy: 1.4.31 - update to latest version

JPeterMugaas:
Update to latest version.  Specify path for python since there are so
many versions.


JPeterMugaas:
OOPS!!! Wrong branch.

PR #598, opened 2016-05-14
gdb: fix path for gcc-5.3.0 libstdcxx

mati865:
Fixes importError:
 ```
Traceback (most recent call last):
  File "<string>", line 3, in <module>
  ImportError: No module named libstdcxx.v6.printers
  /etc/gdbinit:6: Error in sourced command file:
  Error while executing Python code.
```


mingwandroid:
Thanks.

PR #610, opened 2016-05-20
Cmake

JPeterMugaas:
Upgrade to latest version and rework patches.

I assume that this is like CYGWIN version.


mati865:
https://github.com/Alexpux/MSYS2-packages/pull/584
In your PR some things are missing and some are done better.

JPeterMugaas:
Could you elaborate on what is missing in what I have so that I can look at it.  I was not familiar with #584 so I will have to look at that.  Note that I did not define MSYS as equaling CYGWIN because I feel that those are separate things and you might something Cygwin does not support and vice-versa.

I admit that there is a lot to this which is not surprising considering what it is and that the version I'm trying to upgrade from was in the 3.2.3.  And yes, I probably did omit some things.

mati865:
> Note that I did not define MSYS as equaling CYGWIN because I feel that those are separate things and you might something Cygwin does not support and vice-versa.

I also didn't define CYGWIN. It was just comment about patch.


> I admit that there is a lot to this which is not surprising considering what it is and that the version I'm trying to upgrade from was in the 3.2.3. And yes, I probably did omit some things. 
I'm astonished how many things I've missed.

> Could you elaborate on what is missing in what I have so that I can look at it.
Use diff tool like `Meld` to compare both patches.

Alexpux:
Split pull requests for cmake and environment

JPeterMugaas:
Okay.  I'll probably have to close this request because something went wrong when doing a rebase.  I only wanted to do cmake for this PR.    It might better if I restarted the PR elsewhere.  I had also made some other changes from #584 and I hope we both are now on the same exact page.  I know it's annoying to keep making PR requests for the same thing but I'm trying to do the best I can.

mati865:
There is `cherry-pick` function in git. When something goes totally wrong you can reset your local branch to Alexpux/master, cherry-pick commit from your remote master and force push.

PR #612, opened 2016-05-21
Environment improvements

renatosilva:
#### MSYSTEM information

The new script */etc/msystem* now provides information associated with the current MSYSTEM. This is sourced by /etc/profile but can be used by non-login scripts for getting information such as the current compiler architecture or mount point:

```bash
echo ${MSYSTEM_CARCH} ${MSYSTEM_CHOST} ${MSYSTEM_PREFIX}
echo ${MINGW_CHOST} ${MINGW_PREFIX} ${MINGW_PACKAGE_PREFIX}
```

#### Shell manager

The new script */usr/bin/shell* has been introduced for shell management tasks. The following functionalities are currently provided but more can be added:

* Switch between MINGW32, MINGW64 and MSYS shells without restarting terminal.
* Execute console programs from MINGW and system32 directories under winpty.
* Open multiple directories and files with Windows Explorer or default application.

```bash
# Switch shell on the fly
source shell mingw32

# Open documents and folders
shell document.odt /directory .

# Encoding conversion
shell ipconfig

# Better interactive mode
shell /mingw64/bin/python
```

#### Unified shell starter

The new unified script *msys2_shell.cmd* removes the unneeded shell starting wrappers that can be replaced with script arguments, and provides a more specific name. A warning about the required shortcut changes is displayed upon system upgrade. The script has also been improved as follows:

* Fix default MSYSTEM value which is now MSYS when not specified.
* Add simplified parameter -full-path and deprecate -set-full-path.
* Add simplified parameter -here and deprecate -where.
* Add -help parameter describing the available options.
* Remove the ambiguous parameters -mingw and -msys2.

```
:: Processing package changes...
reinstalling filesystem...

      WARNING: the shell starting scripts have been unified. Please update your
      shortcuts to the following targets, otherwise they will STOP WORKING:

          * MSYS2_ROOT\msys2_shell.cmd -mingw32
          * MSYS2_ROOT\msys2_shell.cmd -mingw64
          * MSYS2_ROOT\msys2_shell.cmd -msys
```

#### Warning mechanism

Applications can now add warnings to interactive bash sessions for notifying the user about possible problems. This is done by adding a file containing the warning text to one of the following locations:

* */etc/profile.d/NAME.warning*
* */mingw32/etc/profile.d/NAME.warning*
* */mingw64/etc/profile.d/NAME.warning*

#### Other improvements

* Better MSYSTEM behavior removing the difference between empty value and MSYS and the need for an explicit variable setting since a default value is now implied.

* New variable *MSYS2_PS1* allowing for configuration shared between MSYS2 and other environments to define a specific PS1 without manual detection of the system type.


DavidEGrayson:
I've wanted something like your `/etc/profile/msystem` script for while.  I intend to use those variables in build instructions and scripts to make those instructions work on both 32-bit and 64-bit.   Thanks!

PR #614, opened 2016-05-22
Python sphinx

JPeterMugaas:
Python sphinx and packages it requires.


Alexpux:
We already have mingw python sphinx package. I think this is extra for us.

PR #628, opened 2016-06-01
bashisms: remove arrays [+fix pinentry build]

achadwick:
A couple more have crept in since c3d40c1afe76810bdbb49e35f369279157fd4653.

### DIFF
--- a/gnupg/PKGBUILD
+++ b/gnupg/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=gnupg
 pkgver=1.4.20
-pkgrel=1
+pkgrel=2
 pkgdesc='Complete and free implementation of the OpenPGP standard'
 url='https://gnupg.org/'
 license=('GPL')
--- a/gnupg/gnupg.install
+++ b/gnupg/gnupg.install
@@ -1,5 +1,5 @@
 info_dir=usr/share/info
-info_files=(gnupg1.info)
+info_files="gnupg1.info"
 
 post_install() {
   [ -x usr/bin/install-info ] || return 0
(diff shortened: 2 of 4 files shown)


achadwick:
Failed, but Not My Fault:

```
configure:
***
*** You need libgpg-error to build this program.
**  This library is for example available at
***   ftp://ftp.gnupg.org/gcrypt/libgpg-error
*** (at least version 1.16 is required.)
***
configure:
***
*** You need libassuan to build this program.
*** This library is for example available at
***   ftp://ftp.gnupg.org/gcrypt/libassuan/
*** (at least version 2.1.0 (API 2) is required).
***
configure: error: 
***
*** Required libraries not found. Please consult the above messages
*** and install them before running configure again.
***
==> ERROR: A failure occurred in build().
    Aborting...
==> Removing installed dependencies...
checking dependencies...

Packages (1) ncurses-devel-6.0.20160220-1

Total Removed Size:  3.65 MiB

:: Do you want to remove these packages? [Y/n] 
:: Processing package changes...
removing ncurses-devel...

[MSYS2 CI] FAILURE: pinentry: Building binary failed.

[info] build failed (exit code 1)
```

achadwick:
Okay, build fixed and subject updated. Now it genuinely improves things :grinning:

mati865:
Well, there is the reason why makepkg script starts with `#!/usr/bin/env bash`.

achadwick:
@mati865
I don't understand you. pacman and makepkg are different tools.

Why should pacman/alpm's install scripts be forced to use bash? c3d40c1afe76810bdbb49e35f369279157fd4653 fixes their inability to run with a compliant Bourne shell, and did not introduce the build error you see above. That was there before.

(74f0d029b0dabcfed5d33498478b5c3a7714202a is on top of it, and fixes the build error - I got keen.)

mati865:
@achadwick 
Sorry my bad, I though you modified PKGBUILDS also but they are just release bumps

mingwandroid:
Well my commit from a year ago was because we were considering replacing bash with dash since it was statically linked and bash wasn't. I'm the end Alexey got bash statically linking.

But I don't mind having the install scripts shell agnostic, although IMHO there's only One True Shell and it's name is bash 😉

PR #631, opened 2016-06-06
tar J1.29 Update to latest version

JPeterMugaas:
Rekey patches validpgpkeys, make check workaround.


elieux:
I have to object against the `make check` workaround. The goal should be to make tests work, not to silence them. If you need to skip tests, use `makepkg --nocheck`.

DavidEGrayson:
Yeah, I would prefer to run tar 1.28 with all the tests passing instead of running  tar 1.29 with failing tests.  It would have been better to not merge this.

JPeterMugaas:
Some of the tests that failed have to do with symlinks and numeric ID owners but in Windows, those don't work the same as in Unix-like operating systems.  Ideally, I agree with you about not checking something you know does not work but in this case, I don't think it has anything to do with the code itself.

PR #635, opened 2016-06-15
Added support for 'config.site' files

Karlson2k:
Now users don't need to override installation prefix for MinGW programs and
'configure' scripts use more compatible triplet **'x86_64-w64-mingw32'**
automatically instead of **'x86_64-pc-mingw64'**.
Additionally, all MinGWx32 soft installed **automatically** to `/mingw32` and MinGWx64 soft installed to `/mingw64`. Simply run 'configure' without parameters!
However, if special prefix is specified as configure parameter, it will not be overridden by 'config.site', so all current scripts will continue to work as expected.

This is initial version. If it will be decided, we can modify CLFAGS and LDFLAGS, but in this case grep/sed processing will be required to avoid duplication of flags which can slowdown every configure run.

PS To force 'configure' to ignore site script use `CONFIG_SITE=/dev/null ../srcpath/configure`


Karlson2k:
Updated checksums. CRLF->LF translation was required.

PR #646, opened 2016-06-28
Fix to missing closing paran in FindBoost.cmake

djmott:
patch stripped a needed closing paran causing cmake builds that use boost to fail


fracting:
I think you need to increase pkgrel=2 in PKGBUILD to trigger a CI build. :D

PR #647, opened 2016-06-29
subversion: avoid stack corruption in script bindings

dscho:
Subversion's script bindings (which are used in git-svn) can cause
segmentation faults simply by corrupting the stack under memory
pressure.

This bug was already fixed in Subversion's trunk, but that fix did not
make it into Subversion 1.9.4. Let's forward-port that patch manually.

See http://article.gmane.org/gmane.comp.version-control.git/298427 for
full details.

Reported by Ioannis Kappas.


dscho:
Thanks!

PR #648, opened 2016-06-30
DO NOT MERGE: Fix msys2-runtime upgrade issue by split core-update and compile into…

fracting:
… two shells.

Hi @renatosilva 

This patch could solve a "race-condition" when Tea CI mirror is upgraded by the teaci/msys32 and teaci/msys64 docker image have not upgraded to latest msys2-runtime.

See https://tea-ci.org/fracting/MSYS2-packages/114 for an example.

Could you comment on it?

I would be glad to here any suggestion for better naming than "core-update" and "compile" :)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,11 +1,19 @@
 build:
-  image: teaci/msys$$arch
-  pull: true
-  shell: msys$$arch
-  commands:
-    # TODO: remove this option when not anymore needed
-    - if [ $$arch = 32 ]; then export DISABLE_QUALITY_CHECK=true; fi
-    - ./ci-build.sh
+  core-update:
+    image: teaci/msys$$arch
+    pull: true
+    shell: msys$$arch
+    commands:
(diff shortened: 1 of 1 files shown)


PiotrVV:
Someone crypt.h update to version 1.4 with Cygwin? Then the errors disappear in perl msys2-runtime, and openssl.

fracting:
> Someone crypt.h update to version 1.4 with Cygwin? Then the errors disappear in perl msys2-runtime, and openssl.

Hi @PiotrVV, I'm not sure what you mean, do you have a url to a build failure log? Not sure if this is the right place for your question.

PiotrVV:
I need both versions 1.3 and 1.4 file crypt.h. And the possibility of switching depends.

fracting:
@PiotrVV , I'm afraid this is the wrong place to ask, please ask in #msys2 irc channel on oftc, or create a separate issue, thanks.

PiotrVV:
@fracting thanks four you help.

lazka:
We no longer use  .drone.yml, so closing for now.

PR #651, opened 2016-07-05
bashrc: re-include $PS1 priority

PhilipOakley:
6e6310d (filesystem: New specific variable MSYS2_PS1., 2016-05-01)
introduced the MSYS2_PS1 prompt script proriority to allow multiple
device configurations.

Unfortunately it overwrote the $PS1, even when specifically set by the
users configuration, such as the Git-for-Windows SDK
see https://github.com/git-for-windows/git/issues/794#issuecomment-227401839
and the monkey patch fix
https://github.com/git-for-windows/build-extra/pull/122.

Introduce a priority order so that the user's $PS1 (if set) has an
intermediate priority between the MSYS2_PS1 and the default PS1.

Signed-off-by: Philip Oakley <philipoakley@iee.org>

### DIFF
--- a/filesystem/PKGBUILD
+++ b/filesystem/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=filesystem
 pkgver=2016.06
-pkgrel=1
+pkgrel=2
 pkgdesc='Base filesystem'
 arch=('i686' 'x86_64')
 license=('BSD')
@@ -49,7 +49,7 @@ source=('bash.bash_logout'
         'mingw32-config.site'
         'redirect-config.site')
 sha256sums=('6d651f6b0b2d173961a3fa21acd9d44c783ed9cd73a031687698c8b9ed1f6dee'
-            'b985a05176a940bfd46c33ac1a1f78b3b09508d8b7c28f8e23e542975e09e0e2'
+            '82fb61ad0e5625fe29d4fd2ef4d191cb76b24be17967076a45ba9941f855eb48'
             '99eae6e37081edd73b399009c85f4a67a0c14481241ee4937ab45c4178b540fb'
             'ed08a8bd0919ba29ec5dd2d0c74ba74324ca2bcbf7852354b0b6c523809029a0'
             '4330edf340394d0dae50afb04ac2a621f106fe67fb634ec81c4bfb98be2a1eb5'
--- a/filesystem/bash.bashrc
+++ b/filesystem/bash.bashrc
@@ -39,11 +39,14 @@ unset _warning_prefix
 unset _warning_file
 unset _warning
 
-# If MSYS2_PS1 is set, use that as default PS1, otherwise set a default prompt
+# If MSYS2_PS1 is set, use that as default PS1;
+# if a PS1 is already set, use that;
+# otherwise set a default prompt
 # of user@host, MSYSTEM variable, and current_directory
 if test -n "${MSYS2_PS1}"
     then PS1="${MSYS2_PS1}"
-    else PS1='\[\e]0;\w\a\]\n\[\e[32m\]\u@\h \[\e[35m\]$MSYSTEM\[\e[0m\] \[\e[33m\]\w\[\e[0m\]\n\$ '
+    elif test -z "${PS1}"
+        then PS1='\[\e]0;\w\a\]\n\[\e[32m\]\u@\h \[\e[35m\]$MSYSTEM\[\e[0m\] \[\e[33m\]\w\[\e[0m\]\n\$ '
 fi
 
 # Uncomment to use the terminal colours set in DIR_COLORS


PhilipOakley:
@Alexpux I was not sure if the sha256sums and the package vers are to be updated by the maintainers (you?, @elieux ..? ), or whether I should have done that.
The MSYS2 wiki on contributing https://sourceforge.net/p/msys2/wiki/Contributing%20to%20MSYS2/ doesn't break out those little details.

@renatosilva - does this change releflect your intitial intentions?
Philip

fracting:
Yes, please update shasum and pkgrel, thanks!

PhilipOakley:
> Yes, please update shasum and pkgrel, thanks!

Could you clarify a couple of points?

1. What is the sha256sum computed over? Is it just its (the bash.bashrc) 
file contents, in the same way Git does, or is it some larger entity (in 
some given format), and is there an easy to use (recommended) tool to do 
that checksum?

2. When do the `pkgver` and `pkgrel` values change - i.e. is there a 
particular workflow that decides which one increments when? - I'm presuming 
you would expect an increment to `pkgrel=2` ?
--
this is my first dip into the pacman world...
Philip

fracting:
> 1. What is the sha256sum computed over? Is it just its (the bash.bashrc)
file contents, in the same way Git does, or is it some larger entity (in
some given format), and is there an easy to use (recommended) tool to do
that checksum?

There is a builtin tool called sha256sum.exe:
```
$ which sha256sum.exe
/usr/bin/sha256sum.exe

$ pacman -Qo /usr/bin/sha256sum.exe
/usr/bin/sha256sum.exe is owned by coreutils 8.25-1
```

You just need to calculate the checksum of the file in source array:
```
$ sha256sum.exe bash.bashrc
b985a05176a940bfd46c33ac1a1f78b3b09508d8b7c28f8e23e542975e09e0e2 *bash.bashrc
```

> 2. When do the `pkgver` and `pkgrel` values change - i.e. is there a
particular workflow that decides which one increments when? - I'm presuming
you would expect an increment to `pkgrel=2` ?

Increase `pkgrel` whenever there is any non trivial change in the patch. If you fix a typo in a comment, it might not be worth to increase pkgrel, but if there is any functional change, it is considered as non trivial, so we increase `pkgrel` and trigger a real CI build. `pkgver` is usually matching upstream package version, but in the filesystem package case, Msys2 itself is the upstream, so it's just decided by Msys2 maintainers like Alexpux or others.

If you see "[MSYS2 CI] SUCCESS: No changes in package recipes." that means no change in PKGBUILD file is detect so the CI server does nothing.

I typed to slow and found you just made it ;)

PhilipOakley:
Thanks for the feedback. It's good to know I made reasonable guesses ;-)

Feedback on the patch itself is welcomed. I currently only have a lttle XP Sp3 netbook for home/mobile hacking, so the patching was tested there.

mati865:
@fracting There is an easier way.
```
cd <PKGBUILD/dir>
updpkgsums
```
It will automatically calculate sha256 for files contained in `source()` and update PKGBUILD.

PhilipOakley:
I have updated the commit message (only) to correct a spelling mistake and clarify the wording.

PhilipOakley:
@Alexpux Is there anything I need to do to help move this forward? What is the typical process time for changes like this (Summer holidays may have an impact)?, regards Philip

PhilipOakley:
Many Thanks!

Karlson2k:
After updating to new bash and filesystem-2016.02-2 my default prompt changed to `bash-4.3$`
Before this update my unmodified prompt includes current path.

    bash-4.3$ declare -p PS1
    declare -x PS1="\\s-\\v\\\$ "

Alexpux:
I think this is problem of new filesystem package. See commit:
https://github.com/Alexpux/MSYS2-packages/pull/651

PhilipOakley:
@Karlson2k when was your previous update the Filesystem? was it before 6e6310d (filesystem: New specific variable MSYS2_PS1., 2016-05-01) introduced the MSYS2_PS1 prompt script proriority to allow multiple device configurations? i.e. did you ever install 2016.02-1 - maybe re-install that and see if it fixes it, or not.

That patch 6e6310d  had a bug in that it overwrote the users PS1 (as decribed above). This meant that the Git for Windows SDK's bash prompt was corrupted, and my patch was a minimal fix for the G4W SDK case. As a mainly windows person, I'm not sure what the fix for you would be. I'll have a google about, given your description to see if I can understand what happened, but if you can try a few debug things that would be great.

PhilipOakley:
Angelo Graziosi wrote:
> Wrong prompt with your patches. Upgrading to filesystem-2016.06-2 gives:
>   bash-4.3$
> which is very bad!
> Before I had:
>   the_user@MY_PC MSYS ~/work
> To see it newly I had to downgrade to filesystem-2016.06-1
> 
> Angelo

Thank you for the confirmation.
I'm not sure of all the processing sequences surrounding how the PS1 can be set in many places.
The filesystem-2016.06-1 is broken for me (using the Git SDK install, https://git-for-windows.github.io/#contribute https://github.com/git-for-windows/build-extra/releases/tag/git-sdk-1.0.3).

So at the moment I'm at a loss as to the appropriate fix that accepts everybodies differing ways of setting the PS1 for the different purposes (both bash window title, and bash prompt).

The very original source of the PS1 setting in G4W (and the SDK, which is different) is detailed in https://github.com/git-for-windows/build-extra/pull/122#issuecomment-228997341 

It all looks a bit of a mess. 

Also, jsut seen (but not read) https://github.com/Alexpux/MSYS2-packages/issues/653

Karlson2k:
@Alexpux See #654 for fix.

PR #654, opened 2016-07-14
bash.bashrc: Update PS1 if PS1 wasn't exported.

Karlson2k:
Should fix regression introduced by #651

bash use some default value for PS1, but this value is not exported. This allow us to distinguish between user-defined PS1 and bash automatic default PS1.

### DIFF
--- a/filesystem/PKGBUILD
+++ b/filesystem/PKGBUILD
@@ -3,8 +3,8 @@
 # Contributor: Karlson2k (Evgeny Grin) <k2k@narod.ru>
 
 pkgname=filesystem
-pkgver=2016.06
-pkgrel=2
+pkgver=2016.07
+pkgrel=1
 pkgdesc='Base filesystem'
 arch=('i686' 'x86_64')
 license=('BSD')
@@ -49,7 +49,7 @@ source=('bash.bash_logout'
         'mingw32-config.site'
         'redirect-config.site')
 sha256sums=('6d651f6b0b2d173961a3fa21acd9d44c783ed9cd73a031687698c8b9ed1f6dee'
-            '82fb61ad0e5625fe29d4fd2ef4d191cb76b24be17967076a45ba9941f855eb48'
+            '81706dc0674d06ddd98b630774816810e7dc25c1441e9805a4b6b81e1beaf3f1'
             '99eae6e37081edd73b399009c85f4a67a0c14481241ee4937ab45c4178b540fb'
             'ed08a8bd0919ba29ec5dd2d0c74ba74324ca2bcbf7852354b0b6c523809029a0'
             '4330edf340394d0dae50afb04ac2a621f106fe67fb634ec81c4bfb98be2a1eb5'
--- a/filesystem/bash.bashrc
+++ b/filesystem/bash.bashrc
@@ -40,14 +40,13 @@ unset _warning_file
 unset _warning
 
 # If MSYS2_PS1 is set, use that as default PS1;
-# if a PS1 is already set, use that;
+# if a PS1 is already set and exported, use that;
 # otherwise set a default prompt
 # of user@host, MSYSTEM variable, and current_directory
-if test -n "${MSYS2_PS1}"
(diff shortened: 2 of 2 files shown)


PhilipOakley:
@Karlson2k Many thanks. That does appear to fix the G4W SDK (I've only done a `pacman -Syu` the appropriate number of times to get all the changes, and then repeatedly started the SDK shortcut, and I'm getting the right $TITLEPREFIX that's generated by the git bash being properly shown (the difference is "SDK-32" vs "MINGW32" in the title bar)

Also the start paths differ (correctly) of "/" vs "~".

Many thanks.

PhilipOakley:
@Alexpux Just to say I spoke too soon, and while my initial monkey patch fix git-for-windows/build-extra#122. to G4W still works, reverting that fix doesn't (I reinstalled the SDK from scratch to check, which I hope was sufficient).

What's not clear to me is whether you require that PS1 settings should be exported by/from the bash.bashrc, or whether they should propagate naturally. There appears to be far more nuance in that part of the set up than I currently understand (and lots of stackexchange questions on the subject).

Half the web instructions for setting PS1 don't say `export` while others do, so it looks like a its quite confused.

Do you require/expect that all PS1 settings should be exported?

Karlson2k:
@PhilipOakley Sorry, didn't get.
Could you describe what exactly is not working on currently?

If you export your `PS1` variable before sourcing `bash.bashrc` then it will not be overridden. If you just set `PS1` to any value and **not** export it, it will be set to Msys's default value by `bash.bashrc`.

PhilipOakley:
@Karlson2k, thanks for getting back.

In git for windows, and the G4W SDK, the PS1 is set within the bash.bashrc file (inserted at the end by an [install patch](https://github.com/git-for-windows/build-extra/blob/master/git-extra/git-extra.install#L61-L65))  by dot sourcing a [git-prompt.sh](https://github.com/git-for-windows/build-extra/blob/master/git-extra/git-prompt.sh) script (the link is to the one with my monkey patch). 

That PS1 string does two things, first it sets a TITLEPREFIX for the window (so one can tell which window (SDK or MinGW) is which!), and second, adds the regular __git_ps1 promt function to also show the current checked out branch etc. Note that it doesn't export the PS1. I don't (personally) set the PS1 at all (to the best of my knowledge).

As I understand it, you/we shouldn't need to export the PS1 from the bash.bashrc (and files it dot sources) because all subsequent interactive shells will invoke it anyway, so the PS1 will be set anyway e.g. [this explicit comment](http://stackoverflow.com/questions/31662453/export-ps1-d-in-bashrc-does-not-work#comment51269597_31662453)

What I'm seeing is that my window title bars are not displaying the TITLEPREFIX, indicating 'something has gone wrong'. That was the root cause of all these issues way back at [dschos comment](https://github.com/git-for-windows/git/issues/794#issuecomment-227423648), which is where I noticed the 'fix' by [renatosilva](https://github.com/Alexpux/MSYS2-packages/commit/6e6310dd6). It's seems like we've been digging holes from that point.

While my monkey patch fix to G4W (mentioned above) fixed the issue on the surface, I tried to understand and fix the base issue (and **probably making some mistakes on the way**). At the moment I'm still not sure what the root cause was, and if it's fixed.

Additionally, I know that G4W does a liitle bit of a two-step when starting to get the mintty window going properly, and there are comments in the release notes about needing `winpty` for some apps, so there maybe a rece/cascade issue in setting the PS1 (e.g. the setting seen on the display was set by a previous instance during startup - one window close and another opens...)

So, all in all, things aren't as clear as they could be. 

For example, 

1. Should the file dot sourced from the end of the bash,bashrc have the final say on the PS1 value? [Y/n]
2. Should the PS1 set in the bash.bashrc, and it's dot sourced files, be exported (or not!). [y/N]
3. What should the bash.bashrc file do if it detects a previously exported PS1?
4. And what about an exported MSYS2_PS1 seeking to gain priority?
5. Ss the G4W install patch too assertive in how it modifies the PS1?

(for the first two I've shown what I think my reseach tells me)

PR #656, opened 2016-07-16
msys2_shell.cmd updates and fixes

Karlson2k:
Currently msys2_shell.cmd has several issues:
* You cannot pass parameter to shell, if you have directory with the same name. For example, command `msys2_shell.cmd -here -c 'bash -c ls; sleep 5'` will not work if you have directory named `-c`.
* You cannot use directory names like '-help', '-mingw64' and so on - they will be interpreted as parameters. Moreover, last parameter will be used as directory name. For example command `msys2_shell.cmd -mingw64 -here -msys -defterm scripts` will launch **MSys** shell instead of **MinGW64** shell in directory **`scripts`** instead of directory **`-msys`** (even if `-msys` directory is present).
* If you mistyped name of working directory, it will be silently passed as bash parameter. For example `msys2_shell.cmd -mingw64 -here my_driectory` will pass `my_driectory` as bash parameter.

If you try to use `-where` instead of `-here`, result will be exactly the same.

This update will fix those issues and remove ambiguous meaning of `-here` parameter: now `-here` means current directory and `-where` means specified directory. If specified directory doesn't exist, error will be displayed and error status returned by script.

This update also remove deprecation warnings - let users use parameters which are convenient for them.
Additionally, help message is clarified and some minor fixes and improvements are included.

### DIFF
--- a/filesystem/PKGBUILD
+++ b/filesystem/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=filesystem
 pkgver=2016.07
-pkgrel=1
+pkgrel=2
 pkgdesc='Base filesystem'
 arch=('i686' 'x86_64')
 license=('BSD')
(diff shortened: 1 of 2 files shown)


Rinat84:
I do not know where to write
startup icon mintty
add msys2_shell.cmd 
`rem Setup proper  icon
if "%MSYSTEM%" == "MINGW32" (
  set "MICON=MINGW32"
) else if "%MSYSTEM%" == "MINGW64" (
  set "MICON=MINGW64"
) else (
  set "MICON=MSYS2"
)`

and change
`start "%CONTITLE%" "%WD%mintty" -i /%MICON%.exe /usr/bin/bash --login %1 %2 %3 %4 %5 %6 %7 %8 %9
`

PR #661, opened 2016-07-28
cmake  - 3.6.1 Update to new version

JPeterMugaas:
Remove 3.5.2-case-sensitivity.patch.  It's now in the msys patch
rekey cpuinfo and msys patches.


Alexpux:
````__CYGWIN__```` is defined for MSYS2 so simplify patches

dran7505:
Update to new version

PR #666, opened 2016-08-09
Adding zlib support to Python3 configure command

jonoomph:
Related issue: https://github.com/Alexpux/MSYS2-packages/issues/665


jonoomph:
Researching this some more...

jonoomph:
Please disregard, zlib is already included as a module

PR #675, opened 2016-08-28
update git-flow package to use stable version 1.10.1

anb0s:
To be used in Git for Windows


anb0s:
Just tested the 1.10.1 in my MSYS2 with feature and release branch. My workflow still working like it was with 1.9.1.

anb0s:
I'm asking for integration, so then it can be used in next Git for Windows release, see reference.
Thanks!

anb0s:
Thank You!

PR #684, opened 2016-09-19
Expat, and other prerequesits for cmake

JPeterMugaas:


Alexpux:
Also update curl to 7.50.3

PR #686, opened 2016-09-21
readline - Update to 7.0

JPeterMugaas:


readline/PKGBUILD
```diff
@@ -51,8 +35,11 @@ prepare() {
     msg "applying patch readline${_basever//./}-$(printf "%03d" $p)"
     patch -p0 -i $srcdir/readline${_basever//./}-$(printf "%03d" $p)
   done
-  patch -p1 -i $srcdir/readline-6.3-cygwin.patch
+  msg2 "$srcdir/readline-6.3-cygwin.patch"
```
Alexpux: replace 6.3 with 7.0 then or better with ${_basever}

DavidEGrayson:
Did you make sure that your new PKGBUILD provides `/usr/bin/msys-readline6.dll`?  Otherwise, you'll break tons of packages that were compiled against readline 6, right?

Alexpux:
Yeah new library name will break most MSYS2 packages.

readline/PKGBUILD
JPeterMugaas: It probably will break some packages because of that. /usr/bin/msys-history7.dll /usr/bin/msys-readline7.dll

I'm not sure if there is a fix for that though.

DavidEGrayson:
If this is merged, I hope it is only merged after there is a plan to fix all the things it will break.

Alexpux:
Well currently I have no plans to rebuild nearly all MSYS2 packages so this pull request I will merge when have time to rebuild everything

PR #700, opened 2016-10-07
Winpty, mintty, and ncurses

JPeterMugaas:


JPeterMugaas:
Thanks for catching that.  I deleted that file.  The file was originally created in an experimental dir and slipped into GIT.  I would use that file just to paste a diff I created for one file into a larger patch.

PR #706, opened 2016-10-14
add ported lua and lua51 packages

s-ol:
almost identical to the original arch PKGBUILDs except for how `readline` is replaced by `readline-devel`.

### DIFF
--- a/lua/LICENSE
+++ b/lua/LICENSE
@@ -1,0 +1,30 @@
+License
+
+Lua is free software distributed under the terms of the MIT license
+reproduced below; it may be used for any purpose, including commercial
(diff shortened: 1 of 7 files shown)


mati865:
Is it really need to have emulated version of lua?
If so then reset `pkgrel=5` to `pkgrel=1`, replace `md5sums` with `sha256sums` and regenerate them.

s-ol:
@mati865 the mingw64 version of lua is very broken; it mixes unix and window-style paths all over the place.

Really I am trying to install `luarocks`, and I had a working PKGBUILD for mingw64, but lua wouldn't load the packages because of the path issues.

Alexpux:
I will not add this packages to msys-packages

PR #711, opened 2016-10-20
emacs-nox for msys2

RickLeaf:
emacs without x tool kit version for MSYS2

### DIFF
--- a/emacs-nox/PKGBUILD
+++ b/emacs-nox/PKGBUILD
@@ -1,0 +1,62 @@
+# Maintainer: Ricky <rickleaf.wu@gmail.com>
+_realname="emacs"
+pkgname="emacs-nox"
+pkgver=25.1
+pkgrel=1
+pkgdesc="The extensible, customizable, self-documenting, real-time display editor (msys2)"
+url="http://www.gnu.org/software/${_realname}/"
+license=('GPL3')
+arch=('any')
+depends=('ncurses')
+groups=('editors')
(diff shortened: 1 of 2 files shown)


mati865:
I think naming it `emacs` instead of emacs-nox would be better.
Also check CI error:
```
configure: error: You seem to be running X, but no X development libraries
were found.  You should install the relevant development files for X
and for the toolkit you want, such as Gtk+ or Motif.  Also make
sure you have development files for image handling, i.e.
tiff, gif, jpeg, png and xpm.
If you are sure you want Emacs compiled without X window support, pass
  --without-x
to configure.
```

RickLeaf:
This package is for MSYS2 environment like vim.

RickLeaf:
msys2 do not include x tool kit, if we use it with GUI, we must use WIN32 native GUI solution. But we have full emacs for win32 version in mingw.

RickLeaf:
makepkg --skippgpcheck   in your MSYS

elieux:
@RickLeaf, the mingw version is named `mingw-w64-i686-emacs` or `mingw-w64-x86_64-emacs` in pacman.

RickLeaf:
@elieux yes, I knew. I have used it for a long time.

elieux:
Then why not use the name `emacs` for msys2 emacs?

Anyway, that's just a nitpick. More important is to solve the CI error. I think you just need to add `--without-x` to the `./configure` line.

RickLeaf:
Ok , I rename it

PR #728, opened 2016-10-26
file: Update to 5.29

Chocobo1:


Chocobo1:
The CI error seems to be the same as in https://github.com/Alexpux/MSYS2-packages/pull/723#issuecomment-255720900.

PR #729, opened 2016-10-26
procps-ng: add 3.3.12

Chocobo1:
When installing this package, it will replace the older `procps`.


Chocobo1:
The CI fails because it conflicts with `/usr/bin/ps` from `msys2-runtime` package.
Should I disable building `ps` (in procps-ng)?

PR #730, opened 2016-10-28
ag grepper: New package

forgottenswitch:
"A better grep" and an addon package.
Distributions name it the_silver_searcher.


ag/PKGBUILD
```diff
@@ -0,0 +1,33 @@
+# Maintainer: Mihail Konev <k.mvc@ya.ru>
```
Chocobo1: IMO this folder should be named `ag-git`. 
There should be another folder `ag` which installs stable release version.

The same applies to `cgvg`.

ag/PKGBUILD
```diff
@@ -0,0 +1,33 @@
...
+
+pkgname=ag
+srcname=the_silver_searcher
+pkgver=git
```
Chocobo1: A fixed `pkgver` disables the ability to update this package (via pacman) in the future.
See https://aur.archlinux.org/packages/?O=0&K=-git for a better `pkgver` scheme.

Also if you have time: 
https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/the_silver_searcher
https://wiki.archlinux.org/index.php/Arch_packaging_standards
https://wiki.archlinux.org/index.php/PKGBUILD

mati865:
`pkgver` is wrong. For git packages it should use commit number like here https://github.com/Alexpux/MSYS2-packages/blob/master/gyp-git/PKGBUILD

forgottenswitch:
This cgvg currently does not have config file, like the original ...
https://github.com/uzi/cgvg/blob/master/cgvgrc-example

mati865:
`pkgver` is still wrong, look here: https://github.com/Alexpux/MSYS2-packages/blob/master/gyp-git/PKGBUILD#L24

ag-git/PKGBUILD
```diff
@@ -0,0 +1,35 @@
...
+srcname=the_silver_searcher
+pkgver=git
+pkgrel=1
+pkgdesc="A better grep"
+arch=('i686' 'x86_64')
+license=('Apache')
+url="https://github.com/ggreer/the_silver_searcher"
+groups=('base' 'base-devel')
+provides=('ag')
+conflicts=('ag')
```
mati865: Remove `provides=('ag')`, `conflicts=('ag')` and `groups=('base' 'base-devel')`

ag-git/PKGBUILD
```diff
@@ -0,0 +1,35 @@
+# Maintainer: Mihail Konev <k.mvc@ya.ru>
+
+pkgname=ag
```
mati865: change `pkgname=ag` to `pkgname=ag-git`

cgvg-git/PKGBUILD
```diff
@@ -0,0 +1,29 @@
+# Maintainer: Mihail Konev <k.mvc@ya.ru>
+
+pkgname=cgvg
```
mati865: change `pkgname=cgvg` to `pkgname=cgvg-git`

cgvg-git/PKGBUILD
```diff
@@ -0,0 +1,29 @@
...
+pkgver=git
+pkgrel=1
+pkgdesc="Grep with numerical goto"
+arch=('any')
+license=('MIT')
+url="https://github.com/forgottenswitch/cgvg"
+groups=('base' 'base-devel')
+provides=('cgvg')
+conflicts=('cgvg')
```
mati865: remove `groups=('base' 'base-devel')`, `provides=('cgvg')`, `conflicts=('cgvg')`

ag-git/PKGBUILD
forgottenswitch: https://github.com/Alexpux/MSYS2-packages/blob/master/mintty-git/PKGBUILD#L11

forgottenswitch:
The cgvg is also implement here:
https://github.com/sampson-chen/sack

mingwandroid:
We already have ag in MINGW-packages so I'm against adding it here too. This came up already about a year ago.

forgottenswitch:
But in mingw it clashes with stdin being emulated with named pipes.

forgottenswitch:
There indeed...

mingwandroid:
The named pipe issue can be fixed with some tests for the pipe names. Git for Windows does this.

Slightly related, until the pipe handling is fixed, to prevent MINGW-packages ag from thinking it's in pipe mode pass `.` as an argument to indicate the current working directory.

forgottenswitch:
Put this into header and patch ag, nodejs, ffmpeg, everything `stdin is not a tty`?
https://github.com/git-for-windows/git/pull/102/commits

mingwandroid:
Yes, but we need to be careful of licenses. @kblees, would you consider relicensing this code under BSD so we can use it in GPL and BSD software in MSYS2? If not then we'll need to rewrite it.

forgottenswitch:
As long as we have this grep, no need to copypaste.

mingwandroid:
Sure.

forgottenswitch:
Minimal working example:
https://github.com/forgottenswitch/mingw-isatty

forgottenswitch:
Didn't tested it properly, just copied `iscygtty.c` to `/mingw64/x86_64.../include`,
and built ag from source. Tty detection worked.
https://github.com/Alexpux/mingw-w64/compare/master...forgottenswitch:cygtty
https://github.com/Alexpux/MINGW-packages/compare/master...forgottenswitch:ag-tty
Ag patch was formatted from:
https://github.com/ggreer/the_silver_searcher/compare/master...forgottenswitch:cygtty

PR #742, opened 2016-11-08
libtorrent: New package 0.13.6

tsmolka:
Added new package, inspired from https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/libtorrent


Alexpux:
Don't see reason adding it to msys-packages

PR #744, opened 2016-11-08
yajl: New package 2.1.0

tsmolka:
Similar to pull request from MINGW-packages (https://github.com/Alexpux/MINGW-packages/pull/1489)


Alexpux:
Don't see reason to add it to msys packages

PR #750, opened 2016-11-15
autoconf-archive: updated to 2016.09.16.

KrullKorg:


JPeterMugaas:
This seems to be more critical than people would think.  I ran into an issue that is described at:

 http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=commit;h=d343892fbd0605a9375ff1ae0467aa320a404b12

When attempting to build something for MSYS.

PR #759, opened 2016-11-25
Add the core Styrene repacker tool

achadwick:
I wrote a developer tool for bundling up combinations of local native builds and binary MSYS packages into convenient app bundles shaped for regular Windows users. "All" developers need to do is maintain their normal .desktop files and write a config file for their app; the Styrene tool takes care of all the repackaging and installer-making details.

https://github.com/achadwick/styrene

Would it be OK to incorporate this early draft of the tool into MSYS2-packages in -git form initially?

### DIFF
--- a/styrene-git/PKGBUILD
+++ b/styrene-git/PKGBUILD
@@ -1,0 +1,43 @@
+# Maintainer: Andrew Chadwick <a.t.chadwick@gmail.com>
+
+_realname=styrene
+pkgname=${_realname}-git
+pkgver=0.1.0.4.g06accfb
+pkgrel=1
+pkgdesc="Repacks MSYS2 packages into installer .exes and standalone .zips."
+url="https://github.com/achadwick/styrene"
+arch=('i686' 'x86_64')
+provides=("${_realname}")
+conflicts=("${_realname}")
+makedepends=('git')
+depends=(
+    'python3'
+    'python3-pyalpm'
+    'zip'
+)
+license=('GPLv3+')
+source=("${_realname}::git+https://github.com/achadwick/styrene.git")
+sha256sums=('SKIP')
+
+pkgver() {
+    cd "${srcdir}"/${_realname}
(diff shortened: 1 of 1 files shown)


achadwick:
This tool expects to be run from the MINGW{64,32} environments, despite being an MSYS package. That's because it runs the native compiler and stuff from native binutils at runtime in order to make its output bundles' launcher .exe files, with custom icons and command line handling. Also, is uses native nsis to make the installers☺

It does appropriate runtime checks for what it needs and advises its users appropriately. However if dependency-only packages in mingw{32,64} are required, I can put together a PKGBUILD for those.

achadwick:
That CI error, "Fatal Python error: Py_Initialize: Unable to get the locale encoding" is a well-known WINE Python bug which Styrene itself works around by building only the msys32 side: TeaCI/tea-ci#11

mati865:
I think it should be in MINGW-packages.

achadwick:
@mati865 
No. It imports pyalpm.

mingwandroid:
I have considered porting as much if pacman as possible to mingw-w64.. mostly libalpm. I still might one day.

achadwick:
Is there any good technical or organizational reason to put it in MINGW-packages? Because I can retool Styrene's [package version comparison](https://github.com/achadwick/styrene/blob/06accfbcd129600c0e3bff040d35cfc1a447b1bd/styrene/bundle.py#L422) to actually use `/usr/bin/vercmp` pretty easily, on reflection.

Right now it would seem to be unnecessary duplication, although being able to use image-manipulation libraries to check the icon sizes appeals to me.

Just let me know why there's this preference :slightly_smiling_face:

mingwandroid:
Just the general 'MSYS2 is about native software' principle.

If you had time to make enough of libalpm work that way that would be great I think.

I can dig out my old patches of you were interested, though they weren't really hard to come up with.

achadwick:
Okay, I'll withdraw this one then, and investigate what's needed to make it run as a pure native-prefix tool.

PR #767, opened 2016-12-07
nano-syntax-highlighting-git: bump to 275.d2d84b8

rorgoroth:
* r115 --> r275
* add new syntax's to sample conf.


Chocobo1:
Not sure why `nanorc.sample` should be in the repo.
It can be generated on-the-fly as in: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=nano-syntax-highlighting-git#n26

rorgoroth:
I'm fine with it being there, it is nice having a header for each rather than a huge wall of text to fish through since not all of us dump the whole thing in our nano config.

PR #773, opened 2016-12-14
Add glib and libhogweed to dependencies for msys/emacs.

bustercopley:
Add glib and libhogweed to dependencies for msys/emacs, since "/usr/bin/emacs -Q" won't start without them.


bustercopley:
It looks like the "--without-x" configure flag is needed for the tea CI build to succeed, but that's a preexisting problem.

PR #797, opened 2017-01-28
[WIP] Admin prompt

elieux:
Vary the prompt between regular '$' and bold '#'.

Show '#' only if we can detect that the current process is elevated.  It requires `group: db` enabled in /etc/nsswitch.conf.

Some files are wrong, I need to fix them first.


elieux:
@Alexpux, can you verify it and merge? Tea CI is building it in a wrong order.

PR #805, opened 2017-02-09
Perlminary work on Python 3.6.  The build does not completely work be…

JPeterMugaas:
…cause there's a segfault.  I'm checking it in for reviews or for anyone to a better job than I.  Further work will continue on this at my end.  See https://docs.python.org/3.6/whatsnew/3.6.html for more information about this release.

python3 - python -3.6.0 - Updated to latest version
Drop unneed 016-3.4-ctypes-error-formatting patch.  That is in this new version
Rekey some other patches and change the name for the new Python3 version.


PiotrVV:
Maybe this file will help to build python 3.6.1. I attach a zip file with patch.
[011-3.6-thread-cygwin64.zip](https://github.com/Alexpux/MSYS2-packages/files/985994/011-3.6-thread-cygwin64.zip)

PR #810, opened 2017-02-14
Fix missing bracket in cmake FindBoost

kolewu:
The patch used in PKGBUILD for cmake missed a bracket. This is a regression and had been already fixed in #646.


mingwandroid:
Thanks. Merging because I am soon going to start work on preparing for 3.8.0.

PR #833, opened 2017-03-03
Update pass to 1.7 and add libqrencode package

micbou:
Pass 1.7 dropped the pwgen dependency and added an optional dependency: [libqrencode](http://fukuchi.org/works/qrencode/).

---
This change is [Reviewable](https://reviewable.io/reviews/alexpux/msys2-packages/833)

### DIFF
--- a/libqrencode-git/PKGBUILD
+++ b/libqrencode-git/PKGBUILD
@@ -1,0 +1,31 @@
+# Maintainer: micbou <contact@micbou.com>
+
+pkgname=libqrencode-git
+pkgver=v3.4.3.r238.ge3410bf
+pkgrel=1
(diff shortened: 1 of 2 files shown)


micbou:
Is there something blocking the PR?

mati865:
Usually release versions are preferred over vcs but latest release is almost 3 years old and git repo had last commit 10 days ago.

@Alexpux what's the decision?

Alexpux:
@micbou  never do ````make install```` without ````DESTDIR````. In this case you directly install into system instead creating package.

micbou:
@Alexpux Thanks for merging and sorry for the mistake.

PR #853, opened 2017-03-30
python-yaml - 3.12 - New Package.that may be needed for build processes.

JPeterMugaas:


lazka:
Is this still needed? If yes I'd suggest removing the python2 package from it at least since we are trying to get rid of python2.

JPeterMugaas:
Okay.  That package might be old )the current version is 3.13)  I had originally built it so that duktape (https://duktape.org/) could be built.  This package has been sitting for over 2 years and I think someone had made python-yaml available since this package.  So I'm closing this.

lazka:
Thanks for checking.

PR #854, opened 2017-03-31
Taskbar-pinnable mintty

forgottenswitch:
Adds a shell selection dialog that pops up when the terminal is launched without command line arguments, as well as duplicated with `Alt-F2`. For the latter use case, this also fixes `cd ~` by `/etc`.

### DIFF
--- a/filesystem/PKGBUILD
+++ b/filesystem/PKGBUILD
@@ -4,8 +4,8 @@
 # Contributor: Alethea Rose <alethea@alethearose.com>
 
 pkgname=filesystem
-pkgver=2017.02
-pkgrel=4
+pkgver=2017.03
+pkgrel=1
 pkgdesc='Base filesystem'
 arch=('i686' 'x86_64')
 license=('BSD')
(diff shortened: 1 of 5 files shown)


elieux:
I'm okay with the `PS1` patch. The `CHERE_INVOKING` one is kind of required for this to make sense, so I won't object to it as well.

The launcher patch is a big change though and I'm not convinced of its usefulness yet. Have you tried the `msys2-launcher` package? It's three launchers (one for each shell type) and it's pinnable.

forgottenswitch:
If the reason for separate launcher package to exist is maintainability, this is closed.

forgottenswitch:
Added shell interpreter selection and separate taskbar items.
Cannot see what's wrong with the first version apart from that.

- packaged: https://github.com/forgottenswitch/msys2-packages/tree/mintty_launcher-v2
- squashed: https://github.com/forgottenswitch/mintty/tree/msys2-launcher-v2
- raw: https://github.com/forgottenswitch/mintty/tree/msys2-launcher

elieux:
@forgottenswitch, what do you mean? I thought we already agreed on the maintenance burden being the problem.

(Sorry for the delays.)

forgottenswitch:
Packaged the single-repository version at (not going to be updated)
https://github.com/forgottenswitch/msys2-packages/tree/mintty-msys2launcher
If there should have been a comma before the "apart" - a "features missing" was intended.
The change might be rather amounting - but it doesn't alter the code in too many places.

PR #857, opened 2017-04-08
Fix perl to find xs packages

PiotrVV:


Alexpux:
You need first get latest scripts. We already have 5.24 perl in repo.

PiotrVV:
I see and I have many changes to your latest version

PR #864, opened 2017-04-13
Update Upx to 3.93

yujincheng08:


Alexpux:
Why you create so many pull requests for one package? I will ban you if you will do it again

PR #866, opened 2017-04-13
Trying to add openbsd-netcat

yujincheng08:


yujincheng08:
It seems it rquires libbsd, I will try again.

PR #867, opened 2017-04-13
Try to add libbsd

yujincheng08:


yujincheng08:
It seems it need fixing confliction of funopen

PR #869, opened 2017-04-13
Add openbsd-netcat

yujincheng08:
Sorry for my so many tries. ><
I love using openbsd-netcat more than gnu-netcat because it's more powerful.
So I try a day to make it complied on msys2!
Hope it can be merged!


mati865:
Maintainer field has to be fixed and Contributor field removed.

bobpaul:
The status says this is merged, but it doesn't show up when I `pacman -Ss netcat`. Do I just need to wait longer? I did a fresh install today and ran `pacman -Syu` already.

PR #880, opened 2017-04-23
libarchive: Update to 3.3.1

srjek:
~~I needed some packages on my system to build libarchive, so I added them to the makedepends.~~
Most of the previous patches are now in upstream, which is awesome.


srjek:
> The makedepends you added should be removed as they are part of base-devel.
Ah, that explains the difference between my msys install and the build bots, I pushed a new patch that with them removed.

PR #882, opened 2017-04-24
msys2-runtime: Allow unprivileged symlink

Einbert-Xeride:
Tested on Windows 10 Pro (15063.138) x64 and Windows 7 Pro x64.

Added a patch to enable new behavior of symbolic link introduced in Windows 10 Creators Update. (https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/)


elieux:
(Continuing in #883)

PR #883, opened 2017-04-24
msys2-runtime: Allow unprivileged symlink

Einbert-Xeride:
Tested on Windows 10 Pro (15063.138) x64 and Windows 7 Pro x64.

Added a patch to enable new behavior of symbolic link introduced in Windows 10 Creators Update. (https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/)

### DIFF
--- a/msys2-runtime/0024-path.cc-Allow-native-unprivileged-symlink.patch
+++ b/msys2-runtime/0024-path.cc-Allow-native-unprivileged-symlink.patch
@@ -1,0 +1,28 @@
+From 07dc2f408e5da9615c2eb6db07979a1a7592ef78 Mon Sep 17 00:00:00 2001
+From: Einbert_Xeride <einbert-xeride@outlook.com>
+Date: Tue, 25 Apr 2017 02:39:07 +0800
+Subject: [PATCH 24/24] path.cc: Allow native unprivileged symlink
+
+---
+ winsup/cygwin/path.cc | 5 +++--
+ 1 file changed, 3 insertions(+), 2 deletions(-)
+
+diff --git a/winsup/cygwin/path.cc b/winsup/cygwin/path.cc
+index 13f4e39f5..e3aaae3f6 100644
+--- a/winsup/cygwin/path.cc
++++ b/winsup/cygwin/path.cc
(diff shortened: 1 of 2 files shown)


msys2-runtime/0024-path.cc-Allow-native-unprivileged-symlink.patch
```diff
@@ -0,0 +1,28 @@
...
++++ b/winsup/cygwin/path.cc
+@@ -1813,8 +1813,9 @@ symlink_native (const char *oldpath, path_conv &win32_newpath)
+     }
+   /* Try to create native symlink. */
+   if (!CreateSymbolicLinkW (final_newpath->Buffer, final_oldpath->Buffer,
+-			    win32_oldpath.isdir ()
+-			    ? SYMBOLIC_LINK_FLAG_DIRECTORY : 0))
++				(win32_oldpath.isdir ()
++				? SYMBOLIC_LINK_FLAG_DIRECTORY : 0)
++				| 0x2 /* SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE */ ))
```
mati865: Are you sure there is works on both Win 7 and Win 10 for directories?
If `win32_oldpath.isdir()` is true you will get `0x1 | 0x2` which results in 0x3.

Einbert-Xeride: Yes. It works on Win 7 as if this flag (0x2) is not set.

elieux:
Setting this flag unconditionally will cause "Invalid argument" errors on older Windows ([source](https://www.mail-archive.com/cygwin@cygwin.com/msg152399.html)).

Regardless, I don't think we're in any hurry to have this feature, so I'm inclined to close this PR and wait for upstream (which should have it pretty soon). Thanks for your effort, @Einbert-Xeride, and I hope you won't be discouraged to contribute to MSYS2 in the future.

Einbert-Xeride:
@elieux OK. Thank you. Maybe I've done something wrong while testing on Win7. I may investigate it further.

PR #892, opened 2017-05-10
improve sourceware.org vcs protocols

vszakats:
https://sourceware.org/git/cygwin-apps/windows-default-manifest.git doesn't seem to work, but it does with unencrypted Git protocols, which is an improvement over CVS.

### DIFF
--- a/crypt/PKGBUILD
+++ b/crypt/PKGBUILD
@@ -10,7 +10,7 @@ arch=('i686' 'x86_64')
 license=('custom')
 url="https://cygwin.com/"
 makedepends=('gcc' 'git' 'make')
-source=(${pkgbase}-${pkgver}::git://sourceware.org/git/cygwin-apps/crypt.git#tag=${pkgbase}-${pkgver//./_}-release
+source=(${pkgbase}-${pkgver}::git+https://sourceware.org/git/cygwin-apps/crypt.git#tag=${pkgbase}-${pkgver//./_}-release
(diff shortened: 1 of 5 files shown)


vszakats:
Patch seems to be correct, not sure why it fails, but it seems like pre-existing/unrelated issues.

Alexpux:
Try run updpkgsums with this changes

vszakats:
I'm not using Windows for these updates, so `updpkgsums` is not readily available (tried, but it requires further tools). Anyhow the alternate tool I use shows no checksum problem after the patch. For the updated packages that is.

The problem seems to be a timeout for this Git repo:
```
Cloning into bare repository '/c/projects/msys2-packages/mingw-w64-cross-binutils-git/binutils'...
error: Connection timed out after 300007 milliseconds (curl_result = 28, http_code = 0, sha1 = 274404641f6157fa450c5a878f3b31989aa5ae6a)
error: Unable to find 40e5240e1f32b1cc36c45551eb3f59253c40f317 under https://sourceware.org/git/binutils-gdb.git
```

Maybe a shallow clone could help. Or, TLS/SSL can be disabled for this particular repo.

vszakats:
The graver problem is that the local patches for `mingw-w64-cross-binutils-git` won't apply anymore to the repo tip. I'm going to remove this patch and see.

vszakats:
Giving up on this, sorry!

PR #921, opened 2017-06-04
Update Python3 to 3.6.1 version

PiotrVV:


Alexpux:
Why so many new patches?

PiotrVV:
Because I write packages that are most compatible with many different platforms.

PiotrVV:
I only added compatibility with the arm platform.

PiotrVV:
Prepares python for the new msys-runtime version with many fixes for the arm platform

Alexpux:
In few days 3.6.2 will be released

gerryqd:
An issue related to this upgrade is that, the  python3-setuptools is not upgraded yet.
That causes problems!

Please help to fix it.

PiotrVV:
Correct order of building modules:
python-appdirs
python-colorama
python-packaging
python-pygments
python-pyparsing
python-six
python-setuptools

PR #922, opened 2017-06-05
Add device tree compiler

agross-linaro:
This patch adds a package description for the device tree compiler

Signed-off-by: Andy Gross <andy.gross@linaro.org>

### DIFF
--- a/dtc/PKGBUILD
+++ b/dtc/PKGBUILD
@@ -1,0 +1,34 @@
+# Maintainer: Andy Gross <andy.gross@linaro.org>
+
+pkgname=dtc
+pkgver=1.4.4
+pkgrel=1
+pkgdesc="Device Tree Compiler"
+arch=('i686' 'x86_64')
+url="https://git.kernel.org/pub/scm/utils/dtc/dtc/git"
+license=('GPL2')
+groups=('base')
+
+source=("https://git.kernel.org/pub/scm/utils/dtc/dtc.git/snapshot/${pkgname}-${pkgver}.tar.gz"
+        fpic.patch)
+sha256sums=('2f2c0bf4d84763595953885bdcd2159b0b85410018c8ba48cc31b3d6e443e4d8'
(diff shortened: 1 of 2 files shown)


carlescufi:
CI is failing because, in order to build on MSYS2, DTC requires this commit:

https://git.kernel.org/pub/scm/utils/dtc/dtc.git/commit/?id=d990b8013889b816ec054c7e07a77db59c56c400

Which is not part of 1.4.4

dtc/PKGBUILD
```diff
@@ -0,0 +1,27 @@
+# Maintainer: Andy Gross <andy.gross@linaro.org>
+
+pkgname=dtc
+pkgver=1.4.4
```
carlescufi: Is there a way to start with master instead of a released version? 1.4.4 won't build on MSYS2.

mati865: AppVeyor didn't fail so normal build should be fine.

agross-linaro: yeah i noticed that.  The odd thing is locally, it builds fine.  We need to stick with tagged versions.  I am modifying the patch to fix the FPIC stuff and will add it to the build once I get the right format on it.

mati865: I think Tea enforces `-Werror` and that's the reason.

agross-linaro: That makes it hard to test locally.  This behavior changed from a few weeks ago (the last time I compiled dtc).  I don't see warnings during build.  Seems odd to me.

mati865: Tea doesn't really matter here, passing 1 CI is fine enough.

agross-linaro: I see.  Well i just worked up a patch to make this equivalent to the master dtc branch.  I'll go ahead and push that so this will work in all cases.  When the new version of dtc is tagged we can drop the patch.

agross-linaro:
OK well, not sure what environment the tea is using, but the fpic patch is in place and verified to work in msys.  So this should be good to go.  Sorry for the noise

PR #926, opened 2017-06-08
w32api-headers - update to latest version

JPeterMugaas:
w32api-runtime - Update to latest version


Ede123:
Any reason for partially reverting both #837 and #838 in the process?

I also noticed that in https://github.com/Alexpux/MINGW-packages/pull/2533 which reverts https://github.com/Alexpux/MINGW-packages/pull/2269 and https://github.com/Alexpux/MINGW-packages/pull/2270

PR #935, opened 2017-06-29
pacman: Force makepkg to produce .tar files without extended attributes

dscho:
Extended attributes are not necessary for pacman to work. Extended
attributes may not even be supported by the file system to which we
extract the files.

And worst of all: pacman cannot handle extended attributes, but simply
stops extracting files when it encounters an extended attribute.

This is particularly nasty when creating packages on a system configured
to cache, say, file integrity information about .exe files in extended
attributes, as the created packages will fail to install any .exe files.

The fix is very easy: simply pass the -no-xattrs option to bsdtar.

### DIFF
--- a/pacman/0006-makepkg-avoid-creating-.tar-files-with-extended-attr.patch
+++ b/pacman/0006-makepkg-avoid-creating-.tar-files-with-extended-attr.patch
@@ -1,0 +1,39 @@
+From 3757cecebbf77f05742f0226c858edaad0e9654c Mon Sep 17 00:00:00 2001
+From: Johannes Schindelin <johannes.schindelin@gmx.de>
+Date: Thu, 29 Jun 2017 11:54:48 +0200
+Subject: [PATCH] makepkg: avoid creating .tar files with extended attributes
+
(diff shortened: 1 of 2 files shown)


dscho:
Why does [Tea fail](https://tea-ci.org/Alexpux/MSYS2-packages/961)? I do not get this failure here.

mati865:
Tea is not maintained by MSYS2 and is outdated for about year.
Just ignore it.

dscho:
> Tea is not maintained by MSYS2 and is outdated for about year.
> Just ignore it.

Okay. Maybe it should be disabled, then, as it is annoying to see that a PR fails when it should not fail.

In any case, AppVeyor passes, and I think my reasoning is sound (it took a day to figure out what was going wrong here, it's good that it was that easy a fix).

Do you agree?

mati865:
POSIX attributes aren't supported on Windows anyway so this PR should help indeed.

dscho:
> POSIX attributes aren't supported on Windows anyway so this PR should help indeed.

Well, NTFS supports alternate streams, and that is exactly how Cygwin's `getfattr` accesses extended attributes on Windows...

PR #943, opened 2017-07-04
ncurses: Add missing "links" for compatibility

orgads:
Required for enabling compilation of sipp.

### DIFF
--- a/ncurses/PKGBUILD
+++ b/ncurses/PKGBUILD
@@ -5,7 +5,7 @@ pkgname=('ncurses' 'ncurses-devel')
 _basever=6.0
 _date=20170527
 pkgver=${_basever}.${_date}
-pkgrel=1
+pkgrel=2
 pkgdesc="System V Release 4.0 curses emulation library"
 arch=('i686' 'x86_64')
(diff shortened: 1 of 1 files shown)


ncurses/PKGBUILD
```diff
@@ -92,9 +92,12 @@ package_ncurses-devel() {
   rm -rf ${pkgdir}/usr/lib/terminfo
 
   cp -rf ${pkgdir}/usr/include/ncursesw ${pkgdir}/usr/include/ncurses
+  cp -f ${pkgdir}/usr/include/ncursesw/*.h ${pkgdir}/usr/include
```
Alexpux: I disagree with this changes

orgads: This is also done in cygwin (see [here](https://github.com/cygwinports/mingw64-x86_64-ncurses/blob/master/mingw64-x86_64-ncurses.cygport#L37). The headers are installed in /usr/include, and links are created in ncurses and ncursesw). I don't know how many project include these headers without the `ncurses/` prefix, but at least sipp does.

PR #972, opened 2017-08-16
gtk-doc: updated to 1.26.

KrullBorg:


lazka:
fyi, gtk-doc 1.26 was ported from perl to python2/3

PR #973, opened 2017-08-16
curl - 7.55.1 - Update to latest version and add http2 support through nghttp2

JPeterMugaas:
…h nghttp2

jansson - 2.10 - Add package for libnghttp2
libnghttp2 - 1.24.0 - add so curl can support HTTP/2
libssh - 1.80 - Update to altest version

### DIFF
--- a/libnghttp2/PKGBUILD
+++ b/libnghttp2/PKGBUILD
@@ -1,0 +1,91 @@
+# Maintainer: J. Peter Mugaas <jpmugaas@suddenlink.net>
+
+pkgbase=libnghttp2
+pkgname=("nghttp2" "${pkgbase}" "${pkgbase}-devel")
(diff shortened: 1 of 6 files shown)


libnghttp2/PKGBUILD
```diff
@@ -0,0 +1,91 @@
...
+pkgrel=1
+pkgdesc='Framing layer of HTTP/2 is implemented as a reusable C library'
+arch=(i686 x86_64)
+url='https://nghttp2.org/'
+depends=("jansson"
+         #"jemalloc"
+         "libevent"
+         "libxml2"
+         "openssl"
+         "libcares-devel"
```
Alexpux: Here need to be ````libcares```` without devel

PR #990, opened 2017-09-10
Ca-certificates: update to 20170717, p11-kit update to 0.23.7

holgern:
I adopted the pkgbuild from mingw-w64


Alexpux:
Why you not download sources from server but instead add certdata here?

holgern:
I adopted the mingw-w64-ca-certificates/PKGBUILD. There, the certdata is also checked into the git repo.

holgern:
I could use the approach from http://pkgs.fedoraproject.org/cgit/rpms/ca-certificates.git/tree/
There the certdata is get by 

cvs -d :pserver:anonymous@cvs-mirror.mozilla.org:/cvsroot   \      co -p mozilla/security/nss/lib/ckfw/builtins/certdata.txt   \      > certdata.txt

But, the certdata.txt is also checked into the git repo

Alexpux:
"http://ftp.debian.org/debian/pool/main/c/${pkgname}/${pkgname}_${pkgver}.tar.xz"

holgern:
Ok, thanks.
I will use this download-link.

holgern:
Both packages should be working now

PR #992, opened 2017-09-11
libidn2: Fix dependency on libunistring

jtanx:
After upgrading msys2 today, gnupg was broken, because of a transitive dependency on libcurl -> libidn2, which depends on libunistring, but I didn't have that installed.

image


jay:
I had the same problem this morning, this breaks libcurl which now depends on libidn2, which explains why git https is broken. Workaround is install it manually: `pacman -S libunistring`

Ref: #993

PR #996, opened 2017-09-13
Msys2 runtime fix strnstr

holgern:
revert changes in strnstr
in order to fix p11-kit


mattn:
I upgraded msys2-runtime yesterday, git.exe and gcc.exe seems crash and close cmd.exe. Is this issue related the bug?

dscho:
> I upgraded msys2-runtime yesterday, git.exe and gcc.exe seems crash and close cmd.exe. Is this issue related the bug?

Quite possibly.

This PR simply reverts the change in https://github.com/cygwin/cygwin/commit/f22054c94d3f77ba834d55123ffa9fe4aef6f93c, and I spent some time identifying the *real* culprit (as the code itself seemed fine). The root cause is that the code is compiled without the `_GNU_SOURCE` constant defined, and as a consequence the prototype of `memmem()` is not declared. GCC then assumes a return type of `int` (which on 64-bit is crucially *smaller* than the required size of `char *`), hence the upper 32 bits get cut off, and we're in the typical "it works in 32-bit systems but not in 64-bit systems" hell.

I will prepare a patch for Cygwin.

PR #1001, opened 2017-09-15
Publicsuffix list

JPeterMugaas:


Alexpux:
wget - increase pkgrel
Why you remove libnghttp2-devel from dependencies ?

JPeterMugaas:
It was a mistake that I fixed.

PR #1005, opened 2017-09-16
lbxml2 - 2.9.5 - Update to latest version.  Use ICU package (that wor…

JPeterMugaas:
…ks better in the tests).  catalog-revert patch dropped because it's in the source-code.

libxslt - 1.3.0 - Update to latest version and rebuild with ICU-dependant libxml2


libxml2/PKGBUILD
```diff
@@ -1,42 +1,68 @@
...
+apply_patch_p2_with_msg() {
+  for _patch in "$@"
+  do
+    msg2 "Applying $_patch"
+    patch -Nbp2 -i "${srcdir}/$_patch"
+  done
+}
+
+del_file_exists() {
+  for _fname in "$@"
```
jtanx: What's wrong with calling `rm -rf ` or `rm -f` always?

JPeterMugaas: Sometimes, that might trigger an error if the file or dir doesn't exist.

jtanx: That only happens without the `-f` flag.

libxslt/libxslt-1.1.29-msys2.patch
```diff
@@ -1,17 +1,17 @@
...
+@@ -781,6 +781,9 @@
      i*:CYGWIN*:*)
  	echo ${UNAME_MACHINE}-pc-cygwin
  	exit ;;
 +    i*:MSYS*:*)
 +	echo ${UNAME_MACHINE}-pc-msys
 +	exit ;;
-     *:MINGW64*:*)
- 	echo ${UNAME_MACHINE}-pc-mingw64
+     *:MINGW*:*)
```
Alexpux: This is wrong. Return as it was

PR #1013, opened 2017-09-27
pacman/libalpm: ignore file conflicts for foo.exe -> foo renames

lazka:
In case a file "foo.exe" gets renamed to "foo" the msys2 .exe
interpolation makes pacman think an untracker file is replaced.
In case foo and foo.exe are the same file and foo.exe existed
in the old package version we can be sure it's handled by the
old package and ignore the conflict.


lazka:
Might be overkill, ymmv.

lazka:
Build fails due to #1012

PR #1020, opened 2017-09-30
Bump icu dependency requirements to the newest version.

lazka:
It was reported on IRC that installing wget on appveyor wont
update icu which results in icu not being found due to an ABI bump/dll rename.
Bumping the version requirements of icu users should fix this.

### DIFF
--- a/libpsl/PKGBUILD
+++ b/libpsl/PKGBUILD
@@ -3,13 +3,13 @@
 pkgbase='libpsl'
 pkgname=("${pkgbase}" "${pkgbase}-devel")
 pkgver=0.18.0
-pkgrel=1
+pkgrel=2
 pkgdesc='Public Suffix List library'
 url='https://github.com/rockdaboot/libpsl'
 arch=('i686' 'x86_64')
 license=('MIT')
-depends=('icu>=55.1-1')
-makedepends=('libxslt' 'python3' 'publicsuffix-list')
+depends=('icu>=59.1')
+makedepends=('libxslt' 'python3' 'icu-devel' 'publicsuffix-list')
 source=(https://github.com/rockdaboot/libpsl/releases/download/${pkgname}-${pkgver}/${pkgname}-${pkgver}.tar.gz
         0001-win32-fix.patch
         0.16.0-no-undefined.patch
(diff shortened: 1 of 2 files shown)


mati865:
I think CI should be fully upgraded. Maybe adding one more `pacman -Su` could be better solution?

Ede123:
(I was the reporter on IRC when I tried to figure out why wget was failing)

The reason I want to avoid to do a full system upgrade is that it takes over 5 min on AppVeyor to do so (they seem to already have a set of packages pre-installed) which is a lot and mostly unnecessary as I use almost none of those packages.

Unfortunately there does not seem to be a good possibility to update individual packages *and their dependencies* using pacman, so a full system upgrade might be the most reliable solution after all. :-/

mati865:
Ah, that might be problem.
I think AppVeyor will upgrade MSYS2 on request since it's a bit dated installation.

Ede123:
They do occasionally, [last request](https://github.com/appveyor/ci/issues/1736) is still open, though.

BTW, I found a downstream bug while searching for the above issue: [wget doesnt work in MSYS2 anymore](https://github.com/appveyor/ci/issues/1822)

Ede123:
For what it's worth I found two possibilities to shave a few minutes off the upgrade time on AppVeyor.

To avoid upgrading completely useless packages (in my case I do a matrix build for i686 and x86-64 separately):
```
pacman -Quq | grep -v mingw-w64- | xargs pacman -S --needed --noconfirm --noprogressbar
pacman -Quq | grep ${MINGW_PACKAGE_PREFIX} | xargs pacman -S --needed --noconfirm --noprogressbar
```
First command will *only* upgrade MSYS2 packages; second command will upgrade MINGW-packages, but only those matching the systems architecture (this assumes it's run on a MinGW shell).

As I noticed pacman's "checking available disk space" takes quite some time, too:
```
sed -i 's/^CheckSpace/#CheckSpace/g' /etc/pacman.conf
```
(will disable the disk space check)

lazka:
This is was arch does: https://wiki.archlinux.org/index.php/System_maintenance#Partial_upgrades_are_unsupported

Seems sensible, on the other hand this is an easy fix for a probably not so uncommon broken wget on appveyor. So feel free to either close or merge :)

Ede123:
Two additional hints derived with help from #archlinux-pacman:

For upgrading one package and all dependencies:
```
pactree -l wget | sort -u | xargs pacman -S [...]
```

For upgrading a whole group plus all dependencies of all packages:
```
pacman -Sgq base-devel | while read pkg; do pactree -l $pkg; done | sort -u | xargs pacman -S [...]
```

PR #1036, opened 2017-10-16
iperf3: New package (3.2)

Equim-chan:


Alexpux:
Really I dont want to add this package as MSYS2 mostly created as systwem for building MINGW packages. 
For other things use Cygwin instead.

PR #1037, opened 2017-10-16
iperf: New package (2.0.10)

Equim-chan:


sskras:
Can someone please explain, why it went into MSYS2-packages ant into MINGW-packages ?

Alexpux:
I don't want to increase MSYS2 packages list without any good reason. MSYS2 goal is to build MINGW packages

sskras:
@Alexpux , it's very wise decision.  I fixed my question.  What are the reason to put iperf2 into MSYS2?  I guess no package in the world does need iperf.  Correct me if I am wrong, please.

Alexpux:
@sskras where you see iperf2 package? I dont add it.

sskras:
Ah, OK then.  I mixed up a commit from in PR (from @Equim-chan) with an official commit (in public repo).  Thanks @Alexpux for clarification.  I still need to get used to GitHub:)

PR #1040, opened 2017-10-18

h-c-mueller:


h-c-mueller:
It passes AppVeyor x86_64-pc-msys but hangs on tea during configure:

  checking what kind of threads to support... pthreads
  err:seh:setup_exception stack overflow 7344 bytes in thread 01c6 eip 000000007bc99c63 esp 00000000ffff8950 stack 0xffe00000-0xffffa000-0x100000000

I assume this due to a tea vm issue but not the guile package under test. How to proceed?

mati865:
Tea is basically Cygwin under Wine, ignore it.

PR #1041, opened 2017-10-19
URL and protocol updates

vszakats:


vszakats:
As for the failure in `mingw-w64-cross-binutils-git`: I've double checked and the URL update doesn't change the source and that particular patch even applies correctly on my system — so, maybe it's an EOL issue. I've remove that protocol upgrade to silence it.

vszakats:
Moving the CVS -> Git update to a separate PR, so this one should be alright now.

vszakats:
AppVeyor passes!

vszakats:
@Alexpux: AppVeyor tests are passing fine now, tea builds failed due to an unrelated download error — Is there any concern to be addressed with the patch? (If there isn't, a merge would be appreciated here as well!)

PR #1045, opened 2017-10-20
mingw-w64-cross-binutils-git: secure URL

vszakats:


vszakats:
Updated one of the patches to apply on the current master. It should be fine now.

vszakats:
Giving up.

PR #1046, opened 2017-10-23

h-c-mueller:

compile option --with-threads reverted to --without-threads since it fails
on Cygwin under wine (CI/tea)


ErichDonGubler:
Any reason this got suddenly closed? :\

h-c-mueller:
With this pull request (#1046) I tried if CI/tea passes with compile option --without-threads instead of 
--with-threads. In fact if fails also for --without-threads on CI/tea.

So I opened a new pull request #1047 and reverted to use --with-threads.

ErichDonGubler:
@h-c-mueller: No problem. :) Just wanted to understand!

PR #1047, opened 2017-10-23

h-c-mueller:


h-c-mueller:
AppVeyor passes.

PR #1049, opened 2017-10-25
Update `vim` to v8.0.1213

ErichDonGubler:
I've done a very basic smoke test opening `vim` and ensuring that there's no issues loading up my rather large [`vimrc`](https://github.com/erichdongubler-dotfiles/vim/blob/0b53811fe036dedce79be64f74ce466ba31f620d/.config/vim/vimrc). If you'd prefer the community did some more testing, I'm more than happy to help make that happen. :)

The main feature of this update is the addition of the `CmdlineLeave` and `CmdlineEnter` `autocmd` triggers.

### DIFF
--- a/vim/PKGBUILD
+++ b/vim/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=vim
 _topver=8.0
-_patchlevel=0606
+_patchlevel=1216
 _versiondir="${pkgname}${_topver//./}"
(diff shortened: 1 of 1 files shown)


ErichDonGubler:
Gah, I just realized that I chose a patch that hasn't passed all CI checks for `vim` itself -- I'll change the version to the highest one that does before v8.0.1216, which is v8.0.1213.

PR #1050, opened 2017-10-25
Add autoconf to vim's build dependencies

ErichDonGubler:
While creating a [PR updating `vim`](https://github.com/Alexpux/MSYS2-packages/pull/1049), I noticed that `autoconf` wasn't specified as a build dependency and caused my build to fail -- I've added it to the `PKGBUILD` build dependencies.

I don't know of other missing dependencies, but I hope that this PR will help others who maintain the `vim` `PKGBUILD` down the road. :)

### DIFF
--- a/vim/PKGBUILD
+++ b/vim/PKGBUILD
@@ -11,7 +11,7 @@ license=('custom:vim')
 url="http://www.vim.org"
 depends=('ncurses')
 groups=('editors')
-makedepends=('gawk' 'perl' 'python2' 'python3' 'ruby' 'libiconv' 'ncurses-devel' 'libcrypt-devel')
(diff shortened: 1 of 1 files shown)


ErichDonGubler:
Actually...I'm not sure if this pull request is necessary...from the [ArchWiki](https://wiki.archlinux.org/index.php/PKGBUILD#makedepends):

> The group base-devel is assumed to be already installed when building with makepkg. Members of this group should not be included in makedepends array.

I can close this, if desired, but I'm not sure if this project follows the same convention as in the ArchWiki.

mati865:
> I can close this, if desired, but I'm not sure if this project follows the same convention as in the ArchWiki.

Yes, it allows to make `makedepends` more readable.

ErichDonGubler:
Got it. Closing, then! :) Thanks!

PR #1058, opened 2017-11-07
Add libpng package (version 1.6.34)

Adsun701:
This adds libpng as a MSYS2 package. It is needed by the plotutils utility on Linux.


ErichDonGubler:
`zlib-devel` should only be needed in `makedepends`, shouldn't it?

Adsun701:
All fixed. Thanks!

Alexpux:
I dont want to increase packages in MSYS2 repo. Use cygwin for such p[rograms if needed or create MINGW package.
MSYS2 priority is MINGW subsystem.

PR #1059, opened 2017-11-07
Add package moreutils-0.61

ErichDonGubler:


PiotrVV:
[Check the package and conflicting files from MSYS distribution modeled on ArchLinux ](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/moreutils)

ErichDonGubler:
@PiotrVV: I'm...not sure what you're referring to, in this case. In the linked package (which I did originally derive the PKGBUILD I'm PRing here from), there are no noted conflicts, and the `replaces` entry doesn't exist in the current non-`mingw-w64` MSYS2 repo. Am I missing something?

If you're trying to help me troubleshoot the failed build...it's kind of odd, actually. For some reason, the `moreutils` repo is failing to clone on `arch32`, but not `arch64`...I'm not sure how I'd go about reproducing that on my 64-bit machine.

ErichDonGubler:
This was indeed a spurious failure. I have forced a reversion back to the original commit, with an updated timestamp.

PiotrVV:
You do not install the package without changing the filenames as in archlinux
ts.1.gz conflicts with bash package 
parallel.1 parallel conflicts with parallel package

ErichDonGubler:
Ah, understood. You're totally right -- I'll fix that. Thank you! :)

PiotrVV:
You're welcome.

Alexpux:
Why we need this package?

ErichDonGubler:
@Alexpux: I appreciate the need to justify a package, and I think I can do it! :) To quote the [introduction page on the wiki relating to packages](https://github.com/msys2/msys2/wiki/MSYS2-introduction#packages):

> When looking for msys2 packages or deciding to create a new one, keep in mind that MSYS2 doesn't intend to compete with Cygwin or duplicate their efforts. The set of things that belong to the msys2 subsystem is pretty small:
>
> ....
>
> highly useful stuff: `mc`, `ssh`, `rsync`, `lftp`, ...

My justification for you is that the tools provided by `moreutils` DO qualify as "highly useful stuff" -- take, for instance, some of the tools I have found valuable in my personal and professional usage:

* > sponge: soak up standard input and write to a file

    This lets users redirect to and from a file WITHOUT having to have an intermediate file:
    ```bash
    cat something | sort | uniq | sponge something
    # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
    # This example is nice because `cat something ... > something` would NOT do what is obviously intended here.
    ```
* > vipe: insert a text editor into a pipe

    `vipe` is an awesome tool that encapsulates some boilerplate that's hard to get right. It's nice to be able to make a one-off edit for those times when you need to quickly cobble a not-quite-perfect command pipeline.
    ```bash
    # Note how we added `vipe` here.
    cat something | sort | uniq | vipe | sponge something
    # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^ Now...we can edit to our heart's content here with minimal effort.
    ```
* > vidir: edit a directory in your text editor

    `vidir` can make batch operations for files extremely easy -- mass deletes and renames can be exposed to your editor, and then file management need only be limited by your editing prowess. 

Here's a question to make sure I understand your concerns...I'm still trying to understand where exactly things fit between the base MSYS2 and the MINGW package repo. If this package were sufficiently valuable, where do you think it *should* go? I've used MSYS2 for a long time, but I'm new to the MSYS2 community, and am still learning how things are organized.

PR #1064, opened 2017-11-10
secure checksums and urls

vszakats:
It'd be nice to deprecate `md5sum` to avoid future issues.


vszakats:
BTW, `python-pip`'s previous checksum was wrong. Moreover, both `md5sums` and `sha256sums` were present, and both were wrong.

PR #1067, opened 2017-11-14
libffi-git: Add -git suffix to packages

lazka:
And add corresponding "provides"/"conflicts".


lazka:
Reason being that it confuses my [version tracking code](https://quodlibet.duckdns.org/msys2/queue) since two PKGBUILD files produce a libffi package.

PR #1072, opened 2017-11-20
ninja-git: add package

yuui-tanabe:
This add ninja build system's git version package


Alexpux:
We already have release package for ninja.

PR #1088, opened 2017-12-21
mintty: Update to 2.8.2

k-takata:
Also install themes.


k-takata:
Thank you for merging. However, mintty 2.8.3 has been released ;-)

PR #1099, opened 2018-01-04
Change npth Package

yuui-tanabe:
I deleted the patch because I built the package with vanilla source successful.


Alexpux:
You remove lots of code. Return splitting package and patch

Issue #1124, opened 2018-01-30
Arrow keys and backspace not working in bash after recent update


avih:
Today I ran `pacman -Syu` which upgraded the following packages (from the end of `/var/log/pacman.log`):
```
[2018-01-30 11:56] [ALPM] upgraded mpfr (3.1.5.1-3 -> 4.0.0.1-3)
[2018-01-30 11:56] [ALPM] upgraded gcc-libs (6.4.0-1 -> 6.4.0-2)
[2018-01-30 11:56] [ALPM] upgraded ncurses (6.0.20170708-2 -> 6.1.20180127-1)
[2018-01-30 11:56] [ALPM] upgraded gawk (4.1.4-2 -> 4.2.0-1)
[2018-01-30 11:56] [ALPM] upgraded isl (0.16.1-1 -> 0.18-1)
[2018-01-30 11:56] [ALPM] upgraded mpc (1.0.3-1 -> 1.1.0-1)
[2018-01-30 11:56] [ALPM] upgraded msys2-w32api-headers (5.0.0.4961.e09dd4ca-1 -> 6.0.0.5097.36d7b92b-1)
[2018-01-30 11:56] [ALPM] upgraded msys2-w32api-runtime (5.0.0.4961.e09dd4ca-1 -> 6.0.0.5097.36d7b92b-1)
[2018-01-30 11:56] [ALPM] upgraded gcc (6.4.0-1 -> 6.4.0-2)
[2018-01-30 11:56] [ALPM] upgraded ncurses-devel (6.0.20170708-2 -> 6.1.20180127-1)
[2018-01-30 11:56] [ALPM] upgraded wget (1.19.2-1 -> 1.19.3-1)
```

After this update the arrow keys and backspace don't work correctly at the shell input line, in a weird way.

I tried the following with and without a `~/.bashrc` file:

Pressing few chars and then a backspace seems to print a space. Up arrow works but then the shell doesn't clean up the line with history inputs and just appends the history items to the line.

Left and right arrows at the shell prompt don't skip chars as expected, but `ctrl-v` and then pressing an arrow key does print a correct (I think) string, e.g. ctrl-v and then right arrow prints `^[[C` but the shell doesn't seem to re-render the line correctly when used without it.

Up/down/left/right arrows while in `less` (/usr/bin/less) does seem to work correctly.

It's like the shell reads the keys correctly but not able to re-render the input line correctly. Maybe the ncurses update is at fault?

avih:
[...]

PR #1128, opened 2018-02-01
bash: rebuild. See #1124

lazka:


lazka:
hm, looks like that's not working :/

mati865:
It's really weird `make[1]: *** No rule to make target 'install-', needed by 'install'.  Stop.`

PR #1139, opened 2018-02-08
Drop the catgets library and facility from packages being offered.  m…

JPeterMugaas:
…sys2-runtime 2.10 now has it's own facility and indicate that msys2-runtime replaces catgets

Todo: figure out what to do about the "man" package as that package depends upon the catgets library and will to link with catgets.  I would suggest replacing "man" with "man-db" for consistanstancy with Cygwin and Archlinux.


msys2-runtime/PKGBUILD
```diff
@@ -21,6 +21,7 @@ makedepends=('cocom'
              'gettext-devel'
              'libiconv-devel'
              'diffutils')
+replaces=('catgets')
```
Alexpux: Remove this line as this must be in package procedure.

msys2-runtime/PKGBUILD
```diff
@@ -4,7 +4,7 @@
 pkgbase=msys2-runtime
 pkgname=('msys2-runtime' 'msys2-runtime-devel')
 pkgver=2.10.0
-pkgrel=1
+pkgrel=3
```
Alexpux: Why 3 not 2

Alexpux:
man-db is already in repo

msys2-runtime/PKGBUILD
JPeterMugaas: I went ahead and fixed it.

JPeterMugaas:
Okay.  Then the "man" package has to be rebuilt to remove the catgets dependency.  It was linking to it statically. I haven't gotten to that because I'm working on Ruby 2.5.  But ruby 2.4 needs to be rebuilt at the least because that too depended upon the libssp in gcc-libs.

Alexpux:
You dont fix one more comment

Issue #1141, opened 2018-02-09
Updating MSYS2 unattended is broken (MSYS2 locks up on full system upgrade)


Ede123:
Hi all,

the latest update seems to have made it impossible to do a full system upgrade of MSYS2 unattendedly (i.e. via CI on AppVeyor).

1. The removal of catgets makes it hard to even have pacman run an update (MSYS2's own CI [is affected](https://ci.appveyor.com/project/Alexpux/mingw-packages/build/1.0.4718/job/cfn12kli82cvxya4)) for which I found a workaround, though (`yes | pacman`).
2. Even with the workaround it's still not possible to complete the update as [MSYS2 seems to lock up](https://ci.appveyor.com/project/Ede123/inkscape/build/master-501/job/36n9kdepjc916iao) when warning about<br/>`warning: terminate MSYS2 without returning to shell and check for updates again`<br/>`warning: for example close your terminal window instead of calling exit`

Does anybody know a way how to work around this issue?

Also I seem to recall that it worked previously (MSYS2 did not lock up and CI could continue). This seems not to be true any longer. :-/

Alexpux:
@Ede123 there are no solution for it. This warning is always when upgrade core packages.
You must close MSYS2 console, reopen it and continue upgrade packages. There are no other way

Ede123:
Yes, I know the message, but previously the console could be closed and there was no locked process reported by mintty. Nowadays a process remains that needs to be killed.

Also note that for this specific reason I chose to [run pacman once directly on Windows console](https://gitlab.com/inkscape/inkscape/blob/master/.appveyor.yml#L11-12) before starting a bash and continuing the update. This still allowed to workaround the issue previously, [allowing CI to succeed](https://ci.appveyor.com/project/Ede123/inkscape/build/master-350/job/c61uhhh1vyn8u8km) despite upgrades of msys2-runtime.

Ede123:
@Alexpux I found the issue:
* if pacman is run with `--noconfirm` it does not wait but return immediately after full system upgrade, see [here](https://github.com/Alexpux/MSYS2-packages/blob/5a0d2a9c31e50a8f61530aa7351a9db4f6983920/pacman/0000-pacman-msysize.patch#L1553)
* however if pacman is run with `--noconfirm` it assumes "no" for the question `$1 and $2 are in conflict. Remove $2? [y/N]`
* in my last change I removed to `--noconfirm` to account for #1139 which in turn caused the blocking behavior

[...]

PR #1145, opened 2018-02-10
pacman: change default answer of all queries to yes

Ede123:
See #1141 and upstream PR in the main repo: https://github.com/Alexpux/MSYS2-pacman/pull/52

While the change seems trivial, pacman does not build for me right now, so I'm unable to test unfortunately.
The build log is attached.

[pacman-5.0.1-4-x86_64-prepare.log](https://github.com/Alexpux/MSYS2-packages/files/1713818/pacman-5.0.1-4-x86_64-prepare.log)
[pacman-5.0.1-4-x86_64-build.log](https://github.com/Alexpux/MSYS2-packages/files/1713817/pacman-5.0.1-4-x86_64-build.log)


Ede123:
I already tried rebuilding libpsl but it did not fix the issue.

Alexpux:
What about increasing ````pkgrel````

Ede123:
@Alexpux Done. I usually do that at the very end when I'm happy with the result and forgot in this case because of the build issue...

Issue #1107, opened 2018-01-16
default `/etc/profile` redirect me to $HOME (while using tmux)


eromoe:
When I using tmux, it always opened new pane at $HOME ,

With the help of nicm , I found the problem is by  `/etc/profile` 
https://github.com/tmux/tmux/issues/1225#issuecomment-357892584

But I don't understand which line cause the problem .

avih:
> When I using tmux, it always opened new pane at $HOME ,

I assume the issue is that you want a new pane to open at the current `$PWD` ? If yes, you can add this to your `~/.tmux.conf`:
```
# cygwin/msys2 needs this to open a new shell in the current dir
set-environment -g CHERE_INVOKING 1
```

The solution[s] which were pushed recently to msys2 make a normal shell not open at the home dir anymore - either by running mintty.exe directly or by running the "official" launchers `mingw32.exe` or `mingw64.exe` or `msys2.exe`. I'm 100% sure the direct-mintty was opening at the home dir, and I don't recall the startup PWD with the launchers - but I assume it was home too - and now it isn't.

mintty.exe now opens in `/usr/bin` and the launchers now open at the root dir (`/`).

niklasholm:
Should be fixed now.

PR #1151, opened 2018-02-13
Fix issue #1107

niklasholm:
Instead of always changing directory to `$HOME` during shell startup, only do it when started from msys2_shell.cmd as to not break the behavior of `tmux -c` (issue #1107).

### DIFF
--- a/filesystem/05-home-dir.post
+++ b/filesystem/05-home-dir.post
@@ -252,7 +252,8 @@ maybe_create_home ()
   # Make sure we start in home unless invoked by CHERE
   if [ ! -z "${CHERE_INVOKING}" ]; then
     unset CHERE_INVOKING
-  else
+  elif [ "${CD_HOME}" = "yes" ]; then
+    unset CD_HOME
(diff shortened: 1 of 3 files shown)


Alexpux:
Change filesystem package version

avih:
I think this has caused new shell to start at `/usr/bin/` instead of at the home dir. How should one make the shell start at the home dir like before?

niklasholm:
@avih, you're absolutely right, I'll be upping a new pull request shortly with a fix.

avih:
> @avih, you're absolutely right, I'll be upping a new pull request shortly with a fix.

@niklasholm note that to fix #1107 it's enough to use CHERE_INVOKING to `.tmux.conf`, as mentioned here https://github.com/Alexpux/MSYS2-packages/issues/1107#issuecomment-365604283 .

niklasholm:
@avih: If possible I prefer solutions that work out of the box and doesn't require user configuration to make it work as expected.

PR #1154, opened 2018-02-14
Fix cd to $HOME on startup only

niklasholm:
New solution for issue #1107 and reverts parts of the old solution (#1151) that introduced a regression.

I think this solution is prettier and probably what I've should have done in the first place. However, it might not work for shells that doesn't increment the `$SHLVL` variable.


avih:
Confirm it fixes the issue for me. Thanks.

PR #1159, opened 2018-02-15
gnupg: Don't nuke pacman's keyring in post-install

elieux:
It seems this is obsolete and only makes things more difficult for Git-for-Windows and people with their own private pacman repositories. What do you think, @Alexpux?

### DIFF
--- a/gnupg/PKGBUILD
+++ b/gnupg/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=gnupg
 pkgver=1.4.22
-pkgrel=2
+pkgrel=3
 pkgdesc='Complete and free implementation of the OpenPGP standard'
 url='https://gnupg.org/'
 license=('GPL')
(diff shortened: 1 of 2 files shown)


Alexpux:
This is initializing of key ring for MSYS2 Pacman. I will not accept thos

elieux:
Why is this in the GnuPG package though? There is `/etc/post-install/07-pacman-key.post` for this purpose.

Alexpux:
@elieux postinstall scripts are executed only once after MSYS2 install. But keyring package can be updated in one day and then need reinit keyring with new keys.

elieux:
Updating keys is solved in `msys2-keyring.install`.

Alexpux:
Hmm, maybe then this is extra

elieux:
Do you want to check anything else, or can we merge?

Alexpux:
Increase pkgrel?

elieux:
Okay. Force-pushed.

elieux:
Thanks.

PR #1165, opened 2018-02-17
ncurses: Fix double-free and prevent crash

elieux:
Work-around for https://github.com/Alexpux/MSYS2-packages/issues/1124, https://github.com/git-for-windows/git/issues/1473.


Alexpux:
I think need ask ncurses devs instead this ugly hack

elieux:
I agree, this is just an ugly hack, but it should solve the crashes for now.  Unless there are issues with the hack (maybe a big memory leak), I think we should apply it.

I'm talking to the ncurses developer here: https://lists.gnu.org/archive/html/bug-ncurses/2018-02/msg00019.html

avih:
Is the MSYS2 patch `0001-CRITICAL-fix-return-value-of-drv_CanHandle-on-unknow.patch` which is titled "ncurses: fix crash when TERM refers to an unknown terminal" and is from mid-2015 still relevant today? Maybe it's causing/related to this new crash?

elieux:
@avih:  I tried reverting it and it did not make a difference.  I assume this fix does nothing since the ncurses update, but I'd like a confirmation from @dscho, since he's the one who investigated that issue.  See https://lists.gnu.org/archive/html/bug-ncurses/2015-06/msg00047.html.

ncurses/0002-fix-double-free-on-unknown-terminal.patch
```diff
@@ -0,0 +1,13 @@
+--- a/ncurses/tinfo/lib_setup.c.orig	2017-07-01 20:24:50.000000000 +0200
++++ a/ncurses/tinfo/lib_setup.c	2018-02-17 17:53:49.304266500 +0100
+@@ -722,8 +722,8 @@
+ 	    termp->Filedes = (short) Filedes;
+ 	    termp->_termname = strdup(tname);
+ 	} else {
+-	    _nc_free_termtype2(&TerminalType(termp));
+-	    free(my_tcb);
++	    //_nc_free_termtype2(&TerminalType(termp));
++	    //free(my_tcb);
```
dscho: Maybe we can set `termp = NULL;` after it was released (same for `my_tcb`) and then avoid calling `TerminalType(termp)` altogether if `termp` is `NULL` (or probably just guard the first line behind `if (termp)`)?

dscho:
> I tried reverting it and it did not make a difference. I assume this fix does nothing since the ncurses update, but I'd like a confirmation from @dscho, since he's the one who investigated that issue. See https://lists.gnu.org/archive/html/bug-ncurses/2015-06/msg00047.html.

@elieux wow, what a blast from the past. I frankly had forgotten about this. Is there a public Git repository for ncurses in the meantime? (I was a bit put off by "I did it differently" without any hint how, where I could see it, and why it was a good thing to do it differently rather than using a full, public review process).

ncurses/0002-fix-double-free-on-unknown-terminal.patch
elieux: IIUC, this won't work.  The first `free` happens a few calls deep and only the pointer value is passed, so the pointers in the calling function here can't be affected by it.

elieux: It's worsened by all the aliasing (the above-allocated TERMINAL_CONTROL_BLOCK is pointed to by `termp`, `my_tcb` *and* `TCB` inside this function). :/

elieux:
@dscho, there is https://github.com/ThomasDickey/ncurses-snapshots/ which is kept up-to-date by the maintainer.  It's still super-shallow (one commit for each weekly snapshot), but it should be enough to easily investigate what was changed at that time.

ncurses/0002-fix-double-free-on-unknown-terminal.patch
avih: It appears that there's a fix for this issue upstream: https://github.com/ThomasDickey/ncurses-snapshots/commit/dae8e7bc .

elieux: It's eerily similar to my work-around. :D

elieux:
Okay, we can abandon my fix and just update to the latest release of ncurses (v6.1.20180217) instead.

avih:
> Okay, we can abandon my fix and just update to the latest release of ncurses (v6.1.20180217) instead.

Would be nice to also remove the old patch for the same issue if it's not longer required.

elieux:
@avih, can you check [ncurses-6.0-20150627](https://github.com/ThomasDickey/ncurses-snapshots/commit/562dd9f53dc2a6eb589b8e2bc6fcba9526360372) faster than me?  It's killing my browser to view it on GitHub.

elieux:
It's where the fix for 0001 is supposed to be.

avih:
> @avih, can you check https://github.com/ThomasDickey/ncurses-snapshots/commit/562dd9f53dc2a6eb589b8e2bc6fcba9526360372 faster than me? It's killing my browser to view it on GitHub.

For some reason this commit removes all the files from the repo, but the next commit/snapshot https://github.com/ThomasDickey/ncurses-snapshots/commit/42bcfdc823e89181dc4f11dd0134847ce0e60642 which has the same title restores them all.

The 0001-.. patch doesn't appear to be there (the later commit with the same date), at least not as is: https://github.com/ThomasDickey/ncurses-snapshots/blob/42bcfdc823e89181dc4f11dd0134847ce0e60642/ncurses/tinfo/tinfo_driver.c#L116

But it does have some addition of `#define SETUP_FAIL FALSE` which doesn't look like the base commit of the patch has. Maybe it fixes the same issue as the 0001 patch, no clue as I'm completely unfamiliar with ncurses code.

avih:
Regardless of patch analysis though, the 0001 patch and your patch and the latest upstream snapshot all seem to be addressing the exact same symptom: don't crash when TERM is set to unknown value.

IMO if the upstream snapshot without the 0001 patch and/or your patch doesn't crash with unknown TERM value, my guess would be that it's fixed upstream and no local patches are required for this issue.

elieux:
Pull request updated.  I tried re-building bash with these recipes and no crash seems to occur.

elieux:
@Alexpux, please build and upload.

Alexpux:
Remove patch as it not used anymore

elieux:
Oops.  Sorry.  Removed in https://github.com/Alexpux/MSYS2-packages/commit/57707900f0b5bb7dc5ffb03708d1d08cac20a874.

PR #1169, opened 2018-02-20
filesystem: Add -shell parameter to msys2_shell.cmd

jd28:
Allow setting the login shell via command line parameter.  e.g, ```msys2_shell -shell fish -mintty```. 
The default remains bash.

For those that use this script there seems to be no obvious way to change the default login shell besides editing the file.  See https://superuser.com/questions/961699/change-default-shell-on-msys2

The file however gets overwritten when the package is upgraded and modifications lost.

### DIFF
--- a/filesystem/msys2_shell.cmd
+++ b/filesystem/msys2_shell.cmd
@@ -3,6 +3,7 @@ setlocal
 
 set "WD=%__CD__%"
 if NOT EXIST "%WD%msys-2.0.dll" set "WD=%~dp0usr\bin\"
+set "LOGINSHELL=bash"
 
 rem To activate windows native symlinks uncomment next line
 rem set MSYS=winsymlinks:nativestrict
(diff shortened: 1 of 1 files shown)


elieux:
Hi! Thanks for addressing this issue. I won't try to review for now, because I don't know if specifying a custom shell is a good feature.

I've got an idea for the issue of default shell being wrong though:  Let's use `getent passwd $(id -u) | cut -d: -f7` (of course only if those commands are available).

I think we can also replace `%1 %2 %3 %4 %5 %6 %7 %8 %9` with `%*` at some point.

(I can send both as a separate PR once this one gets reviewed.)

jd28:
Heya, thank you for response!

That does does sound like a much nicer solution.  I wasn't aware that msys2 had an ``/etc/passwd``.

I also considered getting the value of the ``SHELL`` value in the selected subsystem ini file, if present, or an env variable might be a reasonable alternative or fallback.

elieux:
It should all be controlled from the passwd database (which can incidentally also recognize the value of `$SHELL` under certain circumstances), but that doesn't mean the launcher can't have a feature to override the default shell using an argument.

Alexpux:
Update checksums and increase pkgrel

PR #1176, opened 2018-02-22
Add new lcov package

lazka:
See https://github.com/Alexpux/MINGW-packages/issues/3156

### DIFF
--- a/lcov/PKGBUILD
+++ b/lcov/PKGBUILD
@@ -1,0 +1,25 @@
+# Maintainer: Christoph Reiter <reiter.christoph@gmail.com>
+
+pkgname=lcov
+pkgver=1.14
+pkgrel=1
+pkgdesc="front-end for GCC's coverage testing tool gcov"
+arch=('i686' 'x86_64')
+url="http://ltp.sourceforge.net/coverage/lcov.php"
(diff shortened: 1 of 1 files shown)


lazka:
@yumetodo why the "confusion" ?

yumetodo:
To use for mingw32/mingw64, how to use this patch?

In my understanding, the answer is nothing.

I cannot understand why adding lcov as msys2 package, not mingw32/mingw64, resolve https://github.com/Alexpux/MINGW-packages/issues/3156.

lazka:
> To use for mingw32/mingw64, how to use this patch?

`pacman -S lcov` and then use lcov as usual, once it is in the repo.

> In my understanding, the answer is nothing.

It works for gcov coverage files produced by mingw gcc. What do you think wouldn't work?

lazka:
@Alexpux any objections?

lazka:
Updated to 1.14 which adds gcc8 support

lazka:
1.14 doesn't support gcc9 and upstream doesn't plan to do a release supporting gcc9 [0], so closing.

[0] https://github.com/linux-test-project/lcov/issues/58#issuecomment-522967473

PR #1180, opened 2018-02-26
Awk: Fix stripping of CR

orgads:
Fixes https://github.com/git-for-windows/git/issues/1524

### DIFF
--- a/gawk/PKGBUILD
+++ b/gawk/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=gawk
 pkgver=4.2.0
-pkgrel=1
+pkgrel=2
 pkgdesc="GNU version of awk"
 arch=('i686' 'x86_64')
 url="https://www.gnu.org/software/gawk/"
@@ -12,10 +12,17 @@ depends=('sh' 'mpfr' 'libintl' 'libreadline')
 makedepends=('gettext-devel' 'mpfr-devel' 'libreadline-devel')
 provides=('awk')
 install=gawk.install
-source=(https://ftp.gnu.org/pub/gnu/${pkgname}/${pkgname}-${pkgver}.tar.gz{,.sig})
+source=(https://ftp.gnu.org/pub/gnu/${pkgname}/${pkgname}-${pkgver}.tar.gz{,.sig}
+        fix-cr.patch)
 sha256sums=('c88046c6e8396ee548bcb941e16def809b7b55b60a1044b5dd254094f347c7d9'
-            'SKIP')
+            'SKIP'
+            'cdd352564fa2da7762c899986dd74bfbc7fc8741113076be0274d3deb0eaade3')
 validpgpkeys=('D1967C63788713177D861ED7DF597815937EC0D2')
+
+prepare() {
+  cd ${srcdir}/${pkgname}-${pkgver}
+  patch -p1 < ${srcdir}/fix-cr.patch
+}
 
 build() {
   cd ${srcdir}/${pkgname}-${pkgver}
--- a/gawk/fix-cr.patch
+++ b/gawk/fix-cr.patch
@@ -1,0 +1,42 @@
+From 34e05f3e05e958a77a219c457bc5a47bb754a6ae Mon Sep 17 00:00:00 2001
+From: Orgad Shaneh <orgads@gmail.com>
+Date: Mon, 26 Feb 2018 08:10:42 +0200
+Subject: [PATCH] Revert "Update posix/gawkmisc.c - remove cygwin code."
+
+It broke stripping of \r on Windows.
+
+This reverts commit 5db38f775d9ba239e125d81dff2010a2ddacb48e.
+---
+ posix/gawkmisc.c | 19 +++++++++++++++++++
+ 1 file changed, 19 insertions(+)
+
+diff --git a/posix/gawkmisc.c b/posix/gawkmisc.c
+index ecf5aff9..40950826 100644
+--- a/posix/gawkmisc.c
++++ b/posix/gawkmisc.c
+@@ -289,3 +289,22 @@ void
+ init_sockets(void)
+ {
+ }
++
++#ifdef __CYGWIN__
++void
++cygwin_premain0(int argc, char **argv, struct per_process *myself)
++{
++	static struct __cygwin_perfile pf[] = {
++		{ "", O_RDONLY | O_TEXT },
++		/*{ "", O_WRONLY | O_BINARY },*/
++		{ NULL, 0 }
++	};
++	cygwin_internal(CW_PERFILE, pf);
++}
++
++void
++cygwin_premain2(int argc, char **argv, struct per_process *myself)
++{
++	setmode(fileno (stdin), O_TEXT);
++}
++#endif
+-- 
+2.16.2.windows.1
+


dscho:
@orgads excellent analysis and work on fixing the issue! I do think, though, that you will want to squash git-for-windows/MSYS2-packages@e65e181a145e940fa00185bc4534be350f535e5a in ;-)

orgads:
Done. I don't have much experience with makepkg, as you can see ;)

Alexpux:
@orgads move configure to build step

orgads:
Done

orgads:
I opened a discussion in awk and cygwin mailing lists. Looks like this was a deliberate change:

http://www.cygwin.com/ml/cygwin/2018-02/msg00280.html

dscho:
> move configure to build step

Why, oh why? 😦

dscho:
> Why, oh why? 😦

I guess it is inherited from Pacman's peculiar interpretation of the words "prepare" and "build": https://wiki.archlinux.org/index.php/Creating_packages#prepare.28.29 & https://wiki.archlinux.org/index.php/Creating_packages#build.28.29 ☹️ ☹️ ☹️

mati865:
@dscho it's Arch standard and makes easier to quickly test changes.
`makepkg -e` will skip `prepare()` so one can modify source and then build it.

Alexpux:
@dscho also when you build for multiple arches you can run prepare only once

svnpenn:
This is NOT a bug - and this pull should be reverted ASAP

http://cygwin.com/ml/cygwin/2017-08/msg00033.html

svnpenn:
@Alexpux this is NOT a bug - and this pull should be reverted ASAP

http://cygwin.com/ml/cygwin/2017-08/msg00033.html

dscho:
> it's Arch standard and makes easier to quickly test changes.

@mati865 hah... well, I don't know about you, but on my machine, the `./configure` run often takes way longer than the *entire* `make` run.

> also when you build for multiple arches you can run prepare only once

@Alexpux okay, that's a valid point. Except that we don't use that feature, do we? When we call `makepkg-mingw`, we still call `makepkg` twice... or am I wrong?

BTW @Alexpux thanks for the quick turnaround on unbreaking our users.

svnpenn:
@Alexpux this is NOT a bug - and this pull should be reverted ASAP

http://cygwin.com/ml/cygwin/2017-08/msg00033.html

orgads:
@svnpenn: Since @dscho locked the issue, I'll reply here.

I do agree that the applied changes in cygwin are "correct". Still, let me try to explain why this should *not* be reverted ASAP.

This kind of change can be accepted, but it must come after a _decent_ period of time, after notifying potential users about the implications, and letting them adapt to the change.

I spent a lot of time investigating this issue, and I really doubt I'm the only one that got hit by this.

At work, we have an awk script that in order to be portable does something like:
```
len = length(substr(string, 2))
if (win) len += 1 # win = match(ENVIRON["OS"], "Windows")
var = substr(string, 2, len-1)
```

I _know_ this is really bad scripting, but I didn't write this script, and it's there - works as expected.

Following this behavior change, I replaced this with proper stripping of `\r`.

BUT we have in this repository about 5-6 active official branches, so we need to cherry-pick or merge this change to all of them. Apart from that, there are about 7-8 collaboration branches. We have users that work on all of these branches, each of them typically has about 2-5 local branches, based on different commits from different branches.

Now I can cherry-pick the change to all of the branches, and announce this change and tell all the users to rebase their local branches. Most of them will not do this immediately, and even if they do, some of them will only rebase the current branch and will forget this issue when they switch to another branch. Each user that will face this problem might spend about 30-40 minutes trying to build, rebuild, call me, pull, update git. I'm talking about ~30 users I'm aware of.

Now let's consider the gerrit issue. The commit-msg hook did a seemingly right assumption that `/^$/` is an empty line. This is broken with the new behavior, because RS was not set. Gerrit's commit-msg hook is installed once, and used for every commit. Users don't update this hook on a regular basis. There was an update to the qt-project gerrit hook about *5 years ago*, and I _still_ notice users with the previous version from time to time (see for example [here](https://codereview.qt-project.org/#/c/211633/1//COMMIT_MSG), and this is one of the primary maintainers of the project. Probably used an old laptop or something).

I applied the needed fix in gerrit today, for version 2.14. It can take *a few years* until users will upgrade to this version (the qt-project for example is stuck with 2.7, and unlikely to upgrade any time soon...). Even after they will upgrade to gerrit, they must notify all the users to update their commit-msg hooks, or they'll get bad commit messages. I assume that there are at least tens of thousands of gerrit users around the globe, not sure how many on Windows. All of them must update the commit-msg to get correct commit messages.

And these 2 small examples are only those that I personally hit. Who knows how many scripts out there rely on this so-called broken behavior?

To summarize, I'm not opposed to the behavior change, but I strongly disagree with just throwing it at the users without proper announcement (and the cygwin mailing list is not enough. Release notes of MSYS2 and Git for Windows must contain this). So :-1: to ASAP. After announcing it, you should give the users _at least_ 6 months (or even a year) to adapt their scripts. Then you can revert this change, and hope for good.

And if the same should be done for sed and grep, so be it.

svnpenn:
@orgads using this command:

    printf 'hello world\r\n' | awk 1

previously stripped carriage returns, which you were happy with and relied on.
however what about people who werent happy with that, and didnt rely on that? If
one didnt want Awk to do that, you had to use a command like this:

    $ unset POSIXLY_CORRECT
    $ printf 'hello world\r\n' | awk -vBINMODE=1 1

and this command doesnt work for Awk scripts, as BINMODE is not recognized even
in the BEGIN section. This error I describe was fixed by the change you are
complaining about.

The pull request here that you tricked @Alexpux into committing reintroduces
this error. Now Awk users have no way to write portable scripts because of your
actions. so yes, while it is certainly inconvenient to do this:

~~~sh
something | tr -d '\r' | awk something
~~~

it is better than breaking portability altogether, which is what you have done.

mati865:
This discussions is going to nowhere.

For now there should be environment variable to let you switch between old vs portable behaviour without using binary mode.

- It could be set to old behaviour by default with some sane deadline announced (it was considered as feature for long time and big amount of tools depending on that shouldn't unexpectedly get broken).

- Alternatively MSYS2 and Git for Windows could define variable to maintain old behaviour by default  for some time.

orgads:
@mati865: An environment variables makes sense.

By the way, `tr -d '\r'` (and also `RS = "\r?\n"`) is also not portable, because `\r` is used as line break (without `\n`) on macOS. I think that the assumption that a script works on the same binary file on all platforms is too inclusive.

svnpenn:
@orgads you keep putting `RS = "\r?\n"`, and maybe you are not getting that is not portable either. POSIX awk doesnt support RegEx for RS, and it doesnt support multi character RS

svnpenn:
> By the way, `tr -d '\r'` (and also `RS = "\r?\n"`) is also not portable, because
> `\r` is used as line break (without `\n`) on macOS. I think that the assumption
that a script works on the same binary file on all platforms is too inclusive.

@orgads you keep putting `RS = "\r?\n"`, and maybe you are not getting that is
not portable either. POSIX awk doesnt support RegEx for RS, and it doesnt
support multi character RS

true portability is not something that can be reached. it is a goal, it is
something everyone should strive for. what youve done here is say "no thats not
important, what is important is that i dont have to rewrite any non-portable
scripts ever". youve walked away from that goal, shame on you.

svnpenn:
> By the way, `tr -d '\r'` (and also `RS = "\r?\n"`) is also not portable, because
> `\r` is used as line break (without `\n`) on macOS. I think that the assumption
that a script works on the same binary file on all platforms is too inclusive.

@orgads you keep putting `RS = "\r?\n"`, and maybe you are not getting that is
not portable either. POSIX awk doesnt support RegEx for RS, and it doesnt
support multi character RS

true portability is not something that can be reached. it is a goal, it is
something everyone should strive for. its the whole reason the POSIX
(**Portable** Operating System Interface) was created.

what youve done here is say "no thats not important, what is important is that i
dont have to rewrite any non-portable scripts ever". youve walked away from that
goal, shame on you.

dscho:
> [...] shame on you.

No, shame on you for completely missing the opportunity to realize how your rudeness makes everything worse.

This is *not* helpful. Stop it.

svnpenn:
@dscho you stop it? Youre responding to a comment from 2 days ago that doesnt even exist in the form that you are quoting. This isnt your repository and I dont like being bosses around. My arguments are strong and are backed by sources and merit.

If you wish to continue off topic bickering that is your mistake.

svnpenn:
https://github.com/Alexpux/MSYS2-packages/issues/1138

dscho:
@svnpenn Your arguments constantly ignore the fact that you would break tons of users' setups, and thereby cause no little harm in terms of time and money.

Also, a very important thing to remember is that text (which is what gawk processes, not binary files) is platform dependent, due to the end-of-line conventions. This has *also* already been pointed out by @orgads.

If you want to turn this into a sober discussion again, do not even post those incorrect claims, and also refrain from insulting people (even if you then delete the comment, to pretend that it never happened).

svnpenn:
@dscho you dont have the authority to tell me to post and what not to post. Stop
pretending like you do. This issue has been raised and rejected by the Cygwin
community:

- http://cygwin.com/ml/cygwin/2018-02/msg00280.html
- http://cygwin.com/ml/cygwin/2017-09/msg00130.html
- http://cygwin.com/ml/cygwin/2017-08/msg00032.html
- http://cygwin.com/ml/cygwin/2017-06/msg00040.html
- http://cygwin.com/ml/cygwin/2017-05/msg00161.html

It was likely passed here because @orgads shielded the history of the change
from @Alexpux. Notice carefully that he made no mention in either repository of
the extensive discussion i have linked above. he either didnt care about it or
didnt know about it.

i am guessing the latter because he posted a new thread on Cygwin when at least
4 other threads already existed.

the final word is this. while it may be a (big) hinderance for some to go back
and change scripts to santize input going into Grep/Sed/Awk, its a necessary
evil. You can sacrifice portability in such a blatant and obvious way as this
pull request does with the only justification being that people can keep their
old scripts.

orgads:
@svnpenn You guessed right, I wasn't aware of these discussions. But neither were @dscho or @Alexpux. Like I said, this change can be accepted, but it will take (a lot of) time. You must give users enough time to adapt to it, and even after accepting it, there has to be a way to force the current behavior. An environment var, like suggested by @mati865, makes a lot of sense. Its default can be changed when decided. After all, this portability issue existed since ever, and you lived with it.

You're right that we can't tell you what to do, but with your attitude you can expect very low cooperation. I consider myself a cooperative person (as you can see from over 2,600 commits I contributed to Qt Creator, and my activity here on github), but because of your attitude I have very low interest in helping out.

I suggest that _you_ propose a pull request that introduces an environment variable which you can use, and we'll consider this issue closed.

dscho:
The argument that MSYS2 should follow Cygwin is probably based on a misunderstanding: for Cygwin, the line ending is LF-only. For MSYS2, which aims to integrate better with Windows, it is the Windows line ending: CR/LF.

I guess that @Alexpux has already come to that conclusion.

svnpenn:
I think some people also misunderstand the importance of portability in scripts

PR #1199, opened 2018-03-10
Please add the following optimizations to /etc/post-install/* and /etc/profile.d/* scripts

mehrdadn:
These changes shave about **220 ms** (**~30%**) off my MSYS2 shell startup times. Thank you!


PiotrVV:
It's not that easy because the first if is from windows but the second third fourth is from cygwin. Such a testing mix will cause many problems in building correctly perl and ruby, and tests that run erroneously unnecessarily will be a small problem. Because the errors will be from unknown problems cut by the windows system

mehrdadn:
I'm very confused, there are many files here. Are you referring to `01-devices.post`? Are you saying the test for `/proc/self/fd/0` is different from `/proc/self/fd/1` and `/proc/self/fd/2`? I'm not sure I understand. But definitely feel free to make additional changes if needed.

PiotrVV:
You do not understand tests in windows built on different rules with another "profile". In Cygwin, they try to change it and some tests do not work so metaphorically because in a cygwin thanks to such a "mix if" the profile is missing.

mehrdadn:
Sorry—I still don't understand whether you're talking about `01-devices.post`, `05-home-dir.post`, or something else. Or if you're saying these are all impossible to modify.

PiotrVV:
Your test is a mix of two systems, which causes many problems because we do not have control over which profile will be loaded.

mehrdadn:
Are you talking about `if /usr/bin/test ! -h /dev/stdin -o ! -h /dev/stdout -o ! -h /dev/stderr -o ! -h /dev/fd; then`?

PiotrVV:
Do not test with windows if sometimes it will load the whole thing wrong because there is no information what inside of cygwin is going on. stdout stderr will be empty

PiotrVV:
Use tests only from cygwin.

mehrdadn:
I'm afraid I'm very confused. I still don't even understand what you mean by `test`. Are you referring to `/usr/bin/test`? Or are you referring to my testing of this commit?

PiotrVV:
Delete test "if / usr / bin / test! -h / dev / stdin -o! -h / dev / stdout -o! -h / dev / stderr -o! -h / dev / fd; then"

mehrdadn:
OK but I still don't understand why. Can we split it into two at least? One for `stdin`, one for the rest?

PiotrVV:
My tests show that these two stdout stderr have sometimes strange unwanted effects like changing the font or changing the encoding. But not only because sometimes the text that lands there is not from what it should be. What in the transmission of information causes instability in many places

mehrdadn:
I see. Is this specific to a Windows version? Can we perhaps detect this with built-ins (e.g. with environmental variables) and take the faster path when we're sure it will work? Or is there no way to detect this at all?

PiotrVV:
It tests on many systems from windows xp to windows 10 and unfortunately there are no rules. But in Windows 10, the role of restriction from the ntfs file system is playing harder.

mehrdadn:
Okay I've undone the change.

PiotrVV:
Thank you. It looks simple but unfortunately the differences in the systems are large and often only tests give some results.

mehrdadn:
By the way, do the `chmod`/`umask`/`mkdir -m 755` etc. calls in `01-devices.post`, `05-home-dir.post`, etc. make any sense on Windows with MSYS2? They slow the scripts down, and they seem to have no effect because MSYS2 can't store Unix file permissions on Windows. Can they be changed to make things faster?

PiotrVV:
Problems are transferring programs from windows xp to windows 10. And where from cygwin to windows. In cygwinie you need support for this in the form of templates mapping in part action is different to work for someone who will change them every moment while changing in cygwinie .

Alexpux:
Update checksums and version

mehrdadn:
I'm confused, would you like me to update something? I'm not sure what to update where.

Alexpux:
@mehrdadn you change files but not update sha256sums and pkgver/pkgrel in PKGBUILD

mehrdadn:
Oops sorry, it's my first time doing this, I had no idea :) I fixed it!

mehrdadn:
Oops sorry forgot the version -- let me change that too.

mehrdadn:
Fixed!

PR #1232, opened 2018-04-26
msys2-runtime: Cherry-pick gcc 7 suport from cygwin

TBBle:
This cherry-picks commit 67609efeb0bcb198463a952a6a214813794a9c2b from cygwin upstream, see https://www.mail-archive.com/cygwin-patches@cygwin.com/msg06820.html


TBBle:
Noticed while starting to ramp-up on looking into #682

TBBle:
@Alexpux I just noticed https://github.com/Alexpux/Cygwin. Should I be making pull requests against that, rather than adding patches here? It looks like the msys2-master branch there is the source of the patchset here.

Alexpux:
@TBBle my Cygwin repo is for other purposes

PR #1274, opened 2018-05-31
Add data package directory to PKG_CONFIG_PATH in makepkg-mingw

cbrt64:
See https://github.com/Alexpux/MINGW-packages/pull/3871


Alexpux:
I'm already do this changes

PR #1279, opened 2018-06-04
pacman-mirrors: Fix sourceforge mirror, avoid redirection

Ratio2:


Alexpux:
Fix checksums and increase pkgrel

Ratio2:
@Alexpux Done

PR #1286, opened 2018-06-08
tar: support Windows paths

jeroen:
Here is a patch we have been using locally for years, which would be nice in msys2. By default tar does not support windows paths:

```sh
tar xzvf C:/Users/Jeroen/Documents/base/R-devel.tar.gz
# tar (child): Cannot connect to C: resolve failed
# gzip: stdin: unexpected end of file
# tar: Child returned status 128
# tar: Error is not recoverable: exiting now
```

This simple patch fixes that. It works great.

### DIFF
--- a/tar/allow-drives.patch
+++ b/tar/allow-drives.patch
@@ -1,0 +1,17 @@
+--- tar-1.28/lib/rmt.h.orig	2018-04-04 14:13:27.434833600 +0200
++++ tar-1.28/lib/rmt.h	2018-04-04 14:19:17.243518800 +0200
+@@ -35,10 +35,14 @@
+    Distributed File System (DFS).  However, when --force-local, a
+    filename is never remote.  */
+ 
++/*
+ #define _remdev(dev_name) \
+   (!force_local_option && (rmt_dev_name__ = strchr (dev_name, ':')) \
+    && rmt_dev_name__ > (dev_name) \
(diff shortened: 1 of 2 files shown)


tar/allow-drives.patch
```diff
@@ -0,0 +1,17 @@
+--- tar-1.28/lib/rmt.h.orig	2018-04-04 14:13:27.434833600 +0200
++++ tar-1.28/lib/rmt.h	2018-04-04 14:19:17.243518800 +0200
+@@ -35,10 +35,14 @@
+    Distributed File System (DFS).  However, when --force-local, a
+    filename is never remote.  */
+ 
++/*
```
JL2210: Not a maintainer of this, but wouldn't it be better if you used `#ifndef _WIN32`?

lazka:
What about using mingw bsdtar.exe instead? (part of mingw-w64-x86_64-libarchive for example)

jeroen:
I don't know, but everything we use in R has been configured to use `tar.exe`.

Do you think there are any drawbacks from this patch?

lazka:
> Do you think there are any drawbacks from this patch?

Not sure.

Looking at the logic this replaces I'd guess `tar xzvf C:/Users/Jeroen/Documents/base/R-devel.tar.gz --force-local` works too.

lazka:
Since there are two workarounds and you can also use cygpath I'm going to close this. These types of patches are very MSYS2 specific and I don't want to keep supporting them.

PR #1288, opened 2018-06-08
Rebuild MSYS 2 - version 2.1.10

JPeterMugaas:
Rebuild with 7.1.0 compiler and toolchain
Add some patches from Cygwin including an extra one after the April release
Update RTL used to build the .DLL to the latest as well as associated libs and tools.


Alexpux:
@JPeterMugaas please create one pull request per package. Dont add msys2-runtime patches - we will wait Cygwin 2.11.0 release

PR #1321, opened 2018-06-26
Automatically check for (unintended) dll dependencies

jeroen:
A little script that shows dll dependencies in the CI script. This way we can manually inspect if there are any unlisted msys2 dependencies.

I modified the `curl` `PKGBUILD` only for testing this script :) We don't have to merge that part.


peterbud:
Is there any way to prevent failing the build script in case there is no dll/exe in the package? Right now all perl package builds fail as ther is no dll or exe in the package:

```
MSYS2 DLL dependencies:
ldd: missing file arguments
Try `ldd --help' for more information.
[MSYS2 CI] FAILURE: perl-URI: Checking dll depencencies failed.
Command exited with code 1
```

jeroen:
Ah sorry about that. I'll try to fix that!

jeroen:
OK I sent a PR: https://github.com/Alexpux/MSYS2-packages/pull/1348

niklasholm:
Did you test this? Looks to me it still fails since the function "inherits" the status code of the last command, which is still falsy. Fix in #1379.

PR #1336, opened 2018-07-12
Fixed dep names in recipes

iscgar:

PR #1340, opened 2018-07-13
Comment out missing check and optional dependencies

iscgar:
As suggested on #1336


lazka:
Some of these deps are now in the repo, so this needs updating. Also increase pkgrel when changing the package.

PR #1375, opened 2018-08-07
better support for Win10 OpenSSH server

rkitover:
If msys2 bash is set as the DefaultShell for the Win10 OpenSSH server,
there is no way to pass -l to make it a login shell, so /etc/profile is
not sourced, PATH is not set, and the bashrc produces errors.

Detect this case in /etc/bash.bashrc, by checking for the SSH_CONNECTION
environment variable and PATH not having msys2 elements, and source
/etc/profile.

### DIFF
--- a/filesystem/PKGBUILD
+++ b/filesystem/PKGBUILD
@@ -4,7 +4,7 @@
 # Contributor: Alethea Rose <alethea@alethearose.com>
 
 pkgname=filesystem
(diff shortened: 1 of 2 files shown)


mati865:
You need to update `pkgver`.

rkitover:
Updated pkgver and sha256sum for the file, thanks!

PR #1412, opened 2018-09-15
gc: Update to 8.0.0

peterbud:


h-c-mueller:
I want just to inform about #1454, which I noticed for i686 (not tested for x86_64).
For me it sounds if 8.0.0 is still experimental stage. On next update please have a look if guile is working too. Thanks.

PR #1453, opened 2018-10-10
Add libtool patches that fixes using clang with compiler-rt and lld

mstorsjo:


mstorsjo:
If this is ok, I can do a similar PR for MINGW-packages to add the same two patches there as well.

mstorsjo:
Actually, one of the patches might potentially be causing issues (response files with binutils ld might need more escaping of backslashes than what is needed in linker scripts), so I'll close this PR until I've resolved my concerns around that.

mstorsjo:
Turns out everything is fine here wrt these patches.

Alexpux:
I think we need add this patches to mingw libtool, not msys2 one.

mstorsjo:
I sent the pull request for the MINGW-packages repo now. Although I thought it might be valuable to have the patches here as well (as they shouldn't be harmful afaik), in case someone only has the msys version of the package installed, in use for mingw targeting (which otherwise mostly works fine for libtool, as far as I've seen).

PR #1464, opened 2018-10-21
Add tup

cmcaine:
In Arch this would be a tup-bin package, I suppose.


lazka:
Sorry this never got looked at.

This package only packages pre-built executable, which we don't allow normally. If you are still interested you can make a package request issue for a package which builds "tup".

PR #1476, opened 2018-11-03
Rebuild

yuui-tanabe:


Alexpux:
I’m provide only release version of clang in repo. In any case I will not build this package

PR #1482, opened 2018-11-09
New package: zstd v1.3.8

OTLabs:
New package: zstd v1.3.7

Zstandard - Fast real-time compression algorithm
https://facebook.github.io/zstd/

### DIFF
--- a/zstd/PKGBUILD
+++ b/zstd/PKGBUILD
@@ -1,0 +1,38 @@
+# Maintainer: Oleg Titov <oleg.titov@gmail.com>
+
+pkgname='zstd'
+pkgver=1.3.7
+pkgrel=1
+pkgdesc="Zstandard - Fast real-time compression algorithm"
+arch=('i686' 'x86_64')
+url='https://facebook.github.io/zstd/'
+license=('BSD')
+makedepends=('gcc' 'make' 'cmake')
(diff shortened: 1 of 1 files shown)


Alexpux:
@OTLabs why we need it in MSYS-packages? We already have it under mingw

OTLabs:
@Alexpux It's a very good question.

I see that there are packages for xz under mingw and msys same time, tar and gzip are msys packages only. I thought that it would be appropriate to have this compression utility in msys too.

I was thinking that mingw and msys are two separate systems so we need to have packages in each of them.

I would quite appreciate your comments about how to maintain same package in msys and/or mingw.

zstd/PKGBUILD
```diff
@@ -0,0 +1,57 @@
...
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_INSTALL_PREFIX="/usr" \
+    ../${pkgname}-${pkgver}/build/cmake
+
+  make install DESTDIR="${srcdir}/dest"
+}
+
+package_zstd() {
+  groups=('libraries')
+  depends=('gcc-libs' "zstd=${pkgver}")
```
Alexpux: packages cant depends on itself. I think you mean libzstd

OTLabs: Done! Thank you for the tip.

PR #1489, opened 2018-11-14
Added 'msys2-runtime' to 'winpty' depends

podsvirov:
- Update sha256sum for 'path_conv.cc' source file
- Increace pkgrel to 2


podsvirov:
Cc: @elieux

Alexpux:
@podsvirov no need to do it as msys2-runtime is always installed

podsvirov:
Ok, sorry for noise.

PR #1496, opened 2018-11-29
Update PKGBUILD

yuui-tanabe:


mati865:
The patches were made for 3.12 version. You have to check which one doesn't apply and recreate it. Then test if package is generally working.

PR #1513, opened 2018-12-06
pigz: New Package

yuui-tanabe:


yuui-tanabe:
This PR adds multi-threading gzip package.

PR #1549, opened 2019-01-13
pacman: fix bash_completion script

plotasse:
The bash_completion script uses \w and \s character classes in its regex, but these do not seem to be implemented in msys2-runtime (for the =~ operator, bash uses the POSIX regcomp/regexec functions which are implemented there). This makes it impossible to auto-complete package names, and probably other things.

This patch replaces them which [[:alnum:]_] and [[:space:]] classes that are implemented and match the same characters.

Not sure if this should be reported upstream since glibc does implement these classes, so the bug does not happen in Arch Linux.

### DIFF
--- a/pacman/0011-bash-completion-fix-regex.patch
+++ b/pacman/0011-bash-completion-fix-regex.patch
@@ -1,0 +1,31 @@
+Unlike glibc, msys2-runtime does not seem to implement \w and \s.
+This replaces them with equivalent classes.
+--- a/scripts/completion/bash_completion.in
++++ b/scripts/completion/bash_completion.in
+@@ -24,7 +24,7 @@
+ }
+ 
+ _arch_incomp() {
+-  local r="\s-(-${1#* }\s|\w*${1% *})"; [[ $COMP_LINE =~ $r ]]
++  local r="[[:space:]]-(-${1#* }[[:space:]]|[[:alnum:]_]*${1% *})"; [[ $COMP_LINE =~ $r ]]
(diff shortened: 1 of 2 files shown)


eli-schwartz:
> since glibc does implement these classes, so the bug does not happen in Arch Linux.

pacman tries to be cross-platform, so this is a valid bug. Moreover, I guess it would probably break on Arch Linux as well, if bash were to be compiled against a different libc.

Please do submit bugs like this to us. :) MSYS2 is the primary entity with a vested interest in pacman aside for Arch Linux itself, and I think it is reasonable for us to collaborate.
(I cannot promise that MSYS2-specific fixes or especially features will be accepted ;) but changes that improve our portability are changes that I will ACK.)

I've rebased these changes onto pacman master and submitted them as https://lists.archlinux.org/pipermail/pacman-dev/2019-June/023465.html

allanbrokeit:
Pulled patch into my queue - will hit master soon.

I see the MSYS2 package is carrying a number of patches to pacman.  I'll review those for anything we should push upstream.

eli-schwartz:
Speaking of which... https://git.archlinux.org/users/eschwartz/pacman.git/commit/?h=queue2&id=52431cfc9c4f1dff26f3c8641e8a07d155f6e022 is intended to make sure libintl links correctly on systems without ngettext in the libc.

However I don't know whether the meson component of this will work, or indeed how to make sure that libintl is fully detected everywhere that it is needed.

PR #1552, opened 2019-01-15
fish: update to 3.0.0-2

### LINKED ISSUE #1546: fish package needs prebuilt /etc/fish/config.fish similar to /etc/profile
Without a file like it fish cannot be used as the login shell, as the paths are all completely broken. I'll see if I can cook something up and open a PR.

orthoxerox:
I am not sure if I've made any mistakes in PKGBUILD, but these fish config files let you use fish as a login shell. Tagging @mashir43 for review. Fixes #1546

### DIFF
--- a/fish/PKGBUILD
+++ b/fish/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=fish
 pkgver=3.0.0
-pkgrel=1
+pkgrel=2
 epoch=
 pkgdesc="a smart and user-friendly command line shell"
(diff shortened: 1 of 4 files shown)


Alexpux:
@orthoxerox resolve conflicts. Dont downgrade fish version

orthoxerox:
Ah, I see, my fork was stale. Will do.

orthoxerox:
Since 3.0.0 doesn't work right now, does it make sense to not downgrade?

PR #1553, opened 2019-01-18
Add git-crypt package, version 0.6.0

Krakonos:
A git plugin for transparent file encryption; works well in msys environment.

### DIFF
--- a/git-crypt/PKGBUILD
+++ b/git-crypt/PKGBUILD
@@ -1,0 +1,22 @@
+# Maintainer: Ladislav Laska <krakonos@krakonos.org>
+
+pkgname="git-crypt"
+pkgver="0.6.0"
+pkgrel="1"
+pkgdesc="Transparent file encryption in git"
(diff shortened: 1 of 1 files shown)


git-crypt/PKGBUILD
```diff
@@ -0,0 +1,20 @@
+pkgname="git-crypt"
```
Alexpux: Add header

Krakonos: Amended. I hope the header is just the note about maintainer, which I put myself in (I'm willing to take care of the package in the future).

Krakonos:
Bump. I have updated the patch quite some time ago. Could you take another look?

Thanks,
Ladislav

Alexpux:
@Krakonos I dont see where you specify prefix for installing?

Krakonos:
@Alexpux You are right. 

I have amended the commit to set correct prefix. (Sorry it took me so long, I missed your comment).

PR #1593, opened 2019-03-15
pacman: Update to 5.1.3

lazka:
(I get doxygen segfaults during the build locally.. calling make again makes the build finish at least, also with the old pacman)


lazka:
ok, the same error here. I guess disabling doxygen might help.

lazka:
added a hacky double make..

peterbud:
@lazka  have you  tried to downgrade doxygen to `1.8.14-3`? See https://github.com/msys2/MINGW-packages/issues/4959

PR #1609, opened 2019-03-30
new package: rlwrap-0.43

rkitover:
This is a very useful little utility that gives you readline editing for
any program with a prompt, and it works perfectly for native windows
programs as well (e.g. rlwrap powershell works fine.)

Can be a nice alternative to winpty for some things.

Signed-off-by: Rafael Kitover <rkitover@gmail.com>

### DIFF
--- a/rlwrap/PKGBUILD
+++ b/rlwrap/PKGBUILD
@@ -1,0 +1,32 @@
+# Contributor: Rafael Kitover <rkitover@gmail.com>
+# Anyone is welcome to update this package, for new versions or patches or w/e.
+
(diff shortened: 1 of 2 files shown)


rlwrap/PKGBUILD
```diff
@@ -0,0 +1,28 @@
+pkgname=('rlwrap')
```
Alexpux: Add header

rkitover:
Hi Alex, thanks for looking at this!

Added header and post-install message.

rkitover:
ping, I think this would be a useful package for some people.

PR #1631, opened 2019-05-04
brotli: drop python2-brotli

lazka:
Nothing depends on it and there is python-brotli for Python 3.

@OTLabs any objections?


OTLabs:
@lazka Agree!

lazka:
The i686 build hangs in CI :/ no idea

OTLabs:
`/usr/lib/python3.7/distutils/dist.py:274: UserWarning: Unknown distribution option: 'test_suite'`

lazka:
Adding a rebaseall run in CI makes things work again, not sure if this is the right fix though.

PR #1641, opened 2019-05-14
Various imrpvements to ruby package - 2.6.3

JPeterMugaas:
- Added some patches from Cygwin
- added a list of provides for default ruby gems and bbundled ruby gems
- added an updpkgprovs for updating the provides in the PKGBUILD that works similar to Perl.

### DIFF
--- a/ruby/PKGBUILD
+++ b/ruby/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=('ruby' 'ruby-docs')
 pkgver=2.6.3
-pkgrel=2
+pkgrel=3
 arch=('i686' 'x86_64')
 url='https://www.ruby-lang.org/en/'
 license=('BSD' 'custom')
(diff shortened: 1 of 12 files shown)


Alexpux:
@JPeterMugaas for what reason you add these patches?

JPeterMugaas:
I think there were several reasons why CygWin had those patches.  One of them is that they package ruby and the bunbled gems separately (perhaps we need to consider that as well) and I think the gems in a 2.6 dir instead of 2.6.0 which we do (even for 2.6.3).  In addition, some things were probably having problems with version info processing so they attempted to fix that (I don't understand it completely).  Then there was some  changes with OpenSSL 1.1.

lazka:
Based on the patch authors it looks like the patches come from fedora.

It's a bit hard to judge them without know what they fix/improve and with not many devs knowing about ruby internals here.

I'm fine with syncing patches if specific problems arise.

PR #1645, opened 2019-05-17
sqlite - 3.28.0 - - update  from CygWin

JPeterMugaas:
Split package up with subpackages for various SQL extensions.


Alexpux:
@JPeterMugaas  fix build errors

JPeterMugaas:
Alright, I have taken the liberty of fixing the problem by adding tcl to the makedepends.  This is so that the tcl extension, tea could also be installed if someone wants that.

PR #1679, opened 2019-06-19
Let the Azure Pipeline know that we moved from Alexpux to msys2

dscho:
I was wondering why the artifacts were not published properly in https://dev.azure.com/msys2/mingw/_build/results?buildId=2600... turns out that the script verifies that the current account is `Alexpux` (when it really is `msys2`).

I can't say that I understand how that CI script works at all, though: it fetches `master` from https://github.com/msys2/MSYS2-packages and compares the commits between the fetched revision and the current revision. I have no idea why they are different, though, but it seems to work. @Alexpux do _you_ know why it works?

In any case, this should fix the problem where the build artifacts are not published in the CI build.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -33,7 +33,7 @@ jobs:
           CHERE_INVOKING: yes
           BUILD_URL: dummy
           DEPLOY_PROVIDER: bintray
-          BINTRAY_ACCOUNT: Alexpux
+          BINTRAY_ACCOUNT: msys2
           PACMAN_REPOSITORY_NAME: ci.msys
       - task: PublishBuildArtifacts@1
         displayName: Publish packages


dscho:
> @Alexpux do _you_ know why it works?

Oh, I guess it is this line: https://github.com/msys2/MSYS2-packages/blob/ee9c8f02d6e5487bedbaf5e817fd00624b8fdcb3/ci-library.sh#L52

In addition to `upstream/master..` (which is now empty because Alexpux/MSYS2-packages is no longer a repository separate from msys2/MSYS2-packages but redirects there instead), it also enumerates `HEAD^..`.

I guess that works, as long as you are disciplined and don't ever push directly to `master`, and as long as you always merge or squash PRs (and instead of using GitHub's "Rebase" feature).

PR #1682, opened 2019-06-21
Fix pacman to work on Vista, fixes #1674, fixes #1181

### LINKED ISSUE #1181: Current pacman refuses to run on Vista
Upgrading from latest binary [MSYS2](https://sourceforge.net/projects/msys2/files/Base/i686/msys2-i686-20170918.exe) package to current breaks pacman on Windows Vista due to undefined `BCryptDeriveKeyPBKDF2` entry point in `bcrypt.dll` which has been [introduced](https://msdn.microsoft.com/en-us/library/windows/desktop/dd433795(v=vs.85).aspx) in Windows 7.

### LINKED ISSUE #1674: GPG setup fail on Windows 2008 / Vista
Initiating msys2 on Windows 2008 or Vista always give me the error below and then `pacman` does not work. Every time I start the shell, it does this again.

It seems to be unable to find the `trustdb.gpg` which should be in `/etc/pacman.d/gnupg/`. What could be causing this?

msys2-vista

jeroen:
The issue is described in https://github.com/msys2/MSYS2-packages/issues/1674.

Libarchive uses Windows 7 API `BCryptDeriveKeyPBKDF2` to encrypt zip files. By disabling this, pacman works again on Windows Vista and Windows 2008.

### DIFF
--- a/libarchive/PKGBUILD
+++ b/libarchive/PKGBUILD
@@ -12,7 +12,7 @@ depends=('gcc-libs' 'libbz2' 'libiconv' 'libexpat' 'liblzma' 'liblz4' 'liblzo2' 
 makedepends=('libbz2-devel' 'libiconv-devel' 'libexpat-devel' 'liblzma-devel' 'liblz4-devel' 'liblzo2-devel' 'libnettle-devel' 'libxml2-devel' 'zlib-devel')
 options=('!strip' 'debug' 'libtool')
 # provides=('libarchive.so')
-source=("https://libarchive.org/downloads/$pkgname-$pkgver.tar.gz"
+source=("https://github.com/libarchive/libarchive/archive/v${pkgver}.tar.gz"
         'libarchive-3.3.2-bcrypt-fix.patch'
         'libarchive-3.3.1-msys2.patch')
 sha256sums=('c160d3c45010a51a924208f13f6b7b956dabdf8c5c60195df188a599028caa7c'
@@ -36,7 +36,11 @@ build() {
     --enable-shared \
     --enable-static \
     --without-libiconv-prefix \
-    --without-xml2
+    --without-xml2 \
+    --without-cng
+
+# CNG breaks pacman on Vista due to BCryptDeriveKeyPBKDF2
+
   make
   make DESTDIR="${srcdir}/dest" install
 }
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=pacman
 pkgver=5.1.3
-pkgrel=2
+pkgrel=3
 pkgdesc="A library-based package manager with dependency support (MSYS2 port)"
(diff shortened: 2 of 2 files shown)


Alexpux:
@jeroen some months ago we switch minimum target fro MSYS2/MINGW to Win7 so I dont see the reason to support Vista. You can do it in you local repo if needed

libarchive/PKGBUILD
```diff
@@ -12,7 +12,7 @@ depends=('gcc-libs' 'libbz2' 'libiconv' 'libexpat' 'liblzma' 'liblz4' 'liblzo2'
 makedepends=('libbz2-devel' 'libiconv-devel' 'libexpat-devel' 'liblzma-devel' 'liblz4-devel' 'liblzo2-devel' 'libnettle-devel' 'libxml2-devel' 'zlib-devel')
 options=('!strip' 'debug' 'libtool')
 # provides=('libarchive.so')
-source=("https://libarchive.org/downloads/$pkgname-$pkgver.tar.gz"
+source=("https://github.com/libarchive/libarchive/archive/v${pkgver}.tar.gz"
```
dscho: That's an unrelated change, I would back it out.

libarchive/PKGBUILD
```diff
@@ -36,7 +36,11 @@ build() {
     --enable-shared \
     --enable-static \
     --without-libiconv-prefix \
-    --without-xml2
+    --without-xml2 \
+    --without-cng
```
dscho: Is there a downside to doing this?

dscho:
> some months ago we switch minimum target fro MSYS2/MINGW to Win7 so I dont see the reason to support Vista.

Does Cygwin still support Vista? My guess is yes...

> You can do it in you local repo if needed

But then pacman still would be broken after every update, no?

Besides, if it does not break anything, it would be a really well-contained change, easy to maintain. So why not?

libarchive/PKGBUILD
mati865: According to https://github.com/libarchive/libarchive/blob/a163558c388bb2ccedef683e8031c11193e6875c/CMakeLists.txt#L210 libarchive would no longer be able to handle encrypted archives.

jeroen: I think this only applies to windows zip archives. Other formats are handled by the respective delegate library (bz2, lzma, etc).

I think there is little need to support encrypted zip files in msys2 runtime (of course they still work in the native `mingw-w64-libarchive`)

jeroen:
> Does Cygwin still support Vista? My guess is yes...

Yes it does, I have tested this myself with the 3.0.7 runtime and the above patch and all works now:

screenshot

libarchive/PKGBUILD
jeroen: I changed it because the `https://libarchive.org/downloads` url is 404.

PR #1687, opened 2019-07-02
msys2-runtime: 0026-mkdir-always-check-for-existence.patch

bwijen:
Temporary patch until cygwin 3.8.x is integrated.


bwijen:
Fixes #1689: pacman --root: error: failed to initialize alpm library

strace revealed:
When alpm builds the directory structure it explicitly relies on mkdir returning EEXIST (See lib/libalpm/util.c: _alpm_makepath_mode)
However mkdir didn't return EEXIST for drives (only directories)
This has been fixed in cygwin 3.8.x

Alexpux:
@bwijen increase pkgrel

bwijen:
@Alexpux pkgrel increased

PR #1695, opened 2019-07-13
mintty - 3.0.2 - Update to llatest version

JPeterMugaas:
include sounds
build with a patch for GCC 9.1.0.


Alexpux:
@JPeterMugaas mintty-3.0.2 not work correctly. It doesnt show all messages, for example, progress of downloading sources or errors of msys2

PR #1698, opened 2019-07-16
mintty: Update to 3.0.2

k-takata:
Also apply a patch from
https://github.com/mintty/mintty/commit/8be3ff0d918058b56ba0d1bc26eea25b875ab929
to be able to compile with gcc 9.x.

This patch should be removed when updating to the next version (3.0.3?).


k-takata:
I didn't aware #1695. Closing as duplicate.

PR #1704, opened 2019-07-25
imrpved cmake build

JPeterMugaas:
cmak3 3.15.0 - r2
clarifed some dependencies mentioned by cmake build process
fix bug where files created by patch were not removed when rerunning makepkg
make new packages for emacs and vim support
make an optimal docs package


Alexpux:
@JPeterMugaas we dont have sphinx package for MSYS python

PR #1770, opened 2019-10-28
unicode-character-database - 12.1.0 - new package.  Used to build guc…

JPeterMugaas:
…harmap and some testing we may encounter.

### DIFF
--- a/unicode-character-database/LICENSE
+++ b/unicode-character-database/LICENSE
@@ -1,0 +1,33 @@
+COPYRIGHT AND PERMISSION NOTICE
+
+Copyright © 1991-2017 Unicode, Inc. All rights reserved.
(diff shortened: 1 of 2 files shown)


lazka:
If it's for gucharmap, shouldn't it be a mingw package then?

JPeterMugaas:
It's possible but I also think it might be used by several other things.  There is a test in utf8proc that uses the database.  I thought it would be more elegant if it was in MSYS rather than MINGW because in MINGW, it require two copies of the unicode database.  We not talking about a program but rather, data that could be used for building and testing other things.  I do have a mingw-w64 package that I can provide.

lazka:
Both gucharmap and utf8proc would be mingw packages, so closing in favor of  https://github.com/msys2/MINGW-packages/pull/5916

PR #1786, opened 2019-11-15
pacman: Translate MSYS2 specific messages into Japanese

k-takata:
`0000-pacman-msysize.patch` adds some translatable messages.
However those messages have not been translated. This PR translates them into Japanese.

### DIFF
--- a/pacman/0011-update-ja-po-for-msys2.patch
+++ b/pacman/0011-update-ja-po-for-msys2.patch
@@ -1,0 +1,30 @@
+--- pacman-5.2.1-orig/src/pacman/po/ja.po	2019-11-15 21:15:19.934578400 +0900
++++ pacman-5.2.1/src/pacman/po/ja.po	2019-11-15 21:32:57.921215600 +0900
+@@ -1439,6 +1439,27 @@
+ msgstr ""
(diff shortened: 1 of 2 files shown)


Alexpux:
@k-takata this must be reported upstream to Arch Linux pacman, rather then there

k-takata:
No, these messages are added by this project. Not by upstream.

here:
https://github.com/msys2/MSYS2-packages/blob/b709a4d6f3406f26569ce82525e1bf0602dba4e0/pacman/0000-pacman-msysize.patch#L1226
here:
https://github.com/msys2/MSYS2-packages/blob/b709a4d6f3406f26569ce82525e1bf0602dba4e0/pacman/0000-pacman-msysize.patch#L1253
and here:
https://github.com/msys2/MSYS2-packages/blob/b709a4d6f3406f26569ce82525e1bf0602dba4e0/pacman/0000-pacman-msysize.patch#L1256-L1257

So, they do not exist in upstream. Such changes would not be accepted by upstream.
In another word, `0011-update-ja-po-for-msys2.patch` can be a part of `0000-pacman-msysize.patch`.

If you actually think that this should be reported to upstream, `0000-pacman-msysize.patch` must be merged into upstream first.

PR #1807, opened 2019-12-22
gawk(5.0.1): bug fix - set shared object extension to `dll`

jannick0:
gawk is able to load extensions, i.e. shared objects, which on Windows
have extension dll. However, gawk sets it to `so`.

The patch added in this commit sets the extension to `dll` on Windows
systems (recognized if EXEEXT equals `.exe`).

### DIFF
--- a/gawk/PKGBUILD
+++ b/gawk/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=gawk
 pkgver=5.0.1
-pkgrel=1
+pkgrel=2
 pkgdesc="GNU version of awk"
 arch=('i686' 'x86_64')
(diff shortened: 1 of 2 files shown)


lazka:
lgtm :+1:

jannick0:
Thanks - but please put this on hold **until the next gawk release** expected to come out early this year which might have a solution for this issue.  Then this PR could be ignored - and currently nobody else complains about it anyways. :)

lazka:
https://git.savannah.gnu.org/cgit/gawk.git/commit/?id=6b7b05f011d15d8b12ee497a26d6dc69cbf445ed

Thanks for pushing this upstream. Yes, I think we better wait for the next release then.

PR #1808, opened 2019-12-22
fmt: new C++ formatting library, v6.1.1

bbolli:
This library might also help the MSYS2 users.

### DIFF
--- a/fmt/PKGBUILD
+++ b/fmt/PKGBUILD
@@ -1,0 +1,57 @@
+# Maintainer: Beat Bolli <dev@drbeat.li>
+
+pkgname=fmt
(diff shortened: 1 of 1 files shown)


lazka:
Since upstream supports Windows I wonder if this should built with mingw instead.

bbolli:
This is already part of the MINGW-packages repo.

lazka:
Ah, right, never mind then.

bbolli:
Obviously no interest to add this.

lazka:
We try keep the cygwin parts small and focus on native packages, sorry :/

bbolli:
No problem; I assumed as much :-)

PR #1815, opened 2020-01-05
perl-PadWalker: add version 2.3

jannick0:


jannick0:
@lazka Hi, could I do anything to get this merged into the repo? Thanks.

lazka:
The current rules are to only add things which are needed to build mingw packages, with some exceptions. Can you add an explanation why you want this package in, or why it's usefull?

jannick0:
@lazka I am using it to help a perl debugger IDE.  Maybe this is not too compelling to get it in here.  However, it might be a bit challenging to manually install perl modules which could cause updates of (outdated) "msys2's perl modules".  I believe this would be a nightmare.

What is the appropriate approach to 'manually' install perl modules without interfering with those packaged by msys2? Guidance would be appreciated.

Thanks.

PR #1816, opened 2020-01-05
perl: move 'libcrypt-devel' from makedepends to depends

jannick0:
Many perl packages, if not all, require the header file <crypt.h>
which is part of 'libcrypt-devel'.  Hence move it from makedepends to
depends.

### DIFF
--- a/perl/PKGBUILD
+++ b/perl/PKGBUILD
@@ -9,8 +9,8 @@ arch=(i686 x86_64)
 license=('GPL')
 url="https://www.perl.org/"
 groups=('base-devel')
-depends=('db' 'gdbm' 'libcrypt' 'coreutils' 'msys2-runtime' 'sh')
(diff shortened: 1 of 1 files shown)


lazka:
Why can't packages that depend on libcrypt-devel for their build process not depend on libcrypt-devel directly?

jannick0:
You bump into this issue when using, e.g., cpanm to install perl modules.  This is different from perl modules pre-packaged for msys. So libcrypt-devel is a dependee of perl itself I would conclude.  It is annoying to find out that cpanm breaks because `<crypt.h>` cannot be found which belongs to crypt-devel. BTW: This was my use-case which caused me to file this PR.

Alexpux:
@jannick0 building modules needs devel packages, but using perl no needed this headers. So this commit is wrong really.

PR #1817, opened 2020-01-06
zstd: some fixes/improvements

lazka:
* zstd: remove unneeded patch
* zstd: install man pages -> man pages!
* zstd: fix core count detection -> faster compression!
* zstd: define _POSIX_C_SOURCE -> no more compiler warnings!


lazka:
I'v proposed both changes upstream:

* https://github.com/facebook/zstd/pull/1941
* https://github.com/facebook/zstd/pull/1948

lazka:
Both patches have been merged upstream and they also now run CI with cygwin that builds and runs the tests.

PR #1823, opened 2020-01-14
u-boot-tools: Add the missing $ before {pkgver}

lbmeng:
Add $ before {pkgver} that was missing in previous commit.


lbmeng:
I noticed this when I was looking at the Azure build logs. I missed this in the previous commit. Sorry for the inconvenience.

PR #1824, opened 2020-01-15
pacman: switch from xz to zstd for package compression

lazka:


jeroen:
I didn't see this coming. Is there a flag at build time to build a `.pkg.tar.xz` that is compatible with older versions of pacman?

lazka:
> I didn't see this coming. Is there a flag at build time to build a `.pkg.tar.xz` that is compatible with older versions of pacman?

pacman should be able to handle these for a few months now and this change only affects mingw packages which you can't update before you've updated pacman anyway.

Running  `PKGEXT='.pkg.tar.xz' makepkg` should work.

> This also broke your CI scripts that look for `*.pkg.tar.xz` like here:
> 
>     * https://github.com/msys2/MINGW-packages/blob/master/ci-build.sh#L32-L33
> 
>     * https://github.com/msys2/MINGW-packages/blob/master/ci-library.sh#L164

Nice catch, thanks, I'll have a look.

jeroen:
> pacman should be able to handle these for a few months now and this change only affects mingw

Yeah I know, I run a special production infrastructure where the users are frozen on a version of pacman from a while back, without zstd support. Going to try with `PKGEXT='.pkg.tar.xz'`

jeroen:
@lazka do you know what this option does? Does that define the default compression in pacman?

https://github.com/msys2/MSYS2-packages/blob/b28a1ac9830f2b15f21093b733a1ce50a926130a/pacman/PKGBUILD#L147

lazka:
yes, I think it sets the default, which is then overwritten by the value in makepkg.conf

expipiplus1:
Hi, this change has made it quite difficult to upgrade older versions of pacman, ones so old that they are unable to read the zst compressed archives.  Sadly these are still able to be found in the wild, in particular as part of msys2-20180531 distributed by the Stack (haskell) tool

lazka:
This change only affects mingw packages, not pacman etc.

mati865:
Latest Msys packages are also archived with zst (see http://repo.msys2.org/msys/x86_64/).
@expipiplus1 assuming that `msys2-runtime` package is not too old you can just upgrade pacman first and then proceed with other packages:
```
pacman -Sy pacman
pacman -Su
```

jeroen:
FYI I'm also running into this. As of today, the preinstalled msys2 on appveyor and azure pipelines dies when trying to upgrade itself:
 
 - https://ci.appveyor.com/project/jeroen/rtools-packages/builds/32960527
 - https://github.com/r-windows/rtools-packages/pull/115/checks?check_run_id=686583591

I think it would be better to keep using xz for msys packages.

lazka:
> I think it would be better to keep using xz for msys packages.

yes, see #1963

jeroen:
> @expipiplus1 assuming that `msys2-runtime` package is not too old you can just upgrade pacman first and then proceed with other packages:
> 
> ```
> pacman -Sy pacman
> pacman -Su
> ```

I'm bootstrapping my CI from `http://repo.msys2.org/distrib/msys2-x86_64-latest.tar.xz` and here the runtime is too old for the new pacman 😭

mati865:
@jeroen this link was not updated since 2018. msys2 website has been pointing to newer location http://repo.msys2.org/distrib/x86_64/ for a long time.
In any case the you can find updated installer here: https://github.com/msys2/msys2-installer/releases/tag/2020-05-17

jeroen:
Thanks. The bundle in the "newer location" is also a year old. But the GH link works!

jeroen:
Actually no, even the 2020-05-17 install segfaults after upgrading the msys2-runtime.

shlomif:
This `.zstd` vs. `.xz` mismatch is affecting my appveyor builds, too: https://ci.appveyor.com/project/shlomif/fc-solve/builds/33023327

Alexpux:
I wonder everyone write about switch to zstd problem. Pacman with zstd support in repo for about few months so if you update msys2 regularly and not once in a year, then everything will be ok.

jtanx:
The problem is bootstrapping off e.g. http://repo.msys2.org/distrib/x86_64/ (i.e. from ci) which hasn't been updated for a long time and doesn't have zstd support.

Also CI environments like appveyor provide out of date msys2 installations in the base image

mati865:
@jtanx there is new base available https://github.com/msys2/msys2-installer/releases/tag/2020-05-17
msys2.org website points to it.

jeroen:
FYI appveyor now has an image with an up-to-date preinstalled msys2:

```
image: Visual Studio 2019
```

shlomif:
> FYI appveyor now has an image with an up-to-date preinstalled msys2:
> 
> ```
> image: Visual Studio 2019
> ```

Thanks for the info! This may prove helpful.

PR #1830, opened 2020-01-19
sqlite: add extension uuid to sqlite extension package

jannick0:
uuid.c added since sqlite 3.30.0.


lazka:
The build failed.

jannick0:
Yes, thanks. `uuid.c` is not yet contained in the package `sqlite 3.30.1`. Too bad ...

Separate comment: I just realized that the commit would not work as such since incomplete: extension files (plus `Makefile.am` adjustments) are added through a patch file which is to be edited whenever an extension file changes, is added or removed.  Q: Why not copying the files from the src package upon `makepkg` time and using a separate `Makefile.am` in a sub-folder taking care of all the extensions?  Wouldn't that be a bit more transparent in terms of maintenance. ... just wondering. :)

PR #1835, opened 2020-01-25
meson: Update to 0.53.1

lazka:


lazka:
did a glib build with it and worked fine

PR #1836, opened 2020-01-25
Fix msys2_shell.cmd argument-handling with spaces

### LINKED ISSUE #1826: Issues with `msys2_shell.cmd` after today upgrade
I see that now `msys2_shell.cmd` has line endings in DOS convection (`CRLF`). 

I converted it in UNIX style (`LF`) with

`d2u msys2_shell.cmd `

and it has issues to start MSYS2.

Running from `cmd`,

`C:\msys64\msys2_shell.cmd`

starts MSYS2 but at DOS prompt I find 

`,*" non atteso.`

(`,* not expected`).

If, instead, I start

`C:\msys64\msys2_shell.cmd -msys2 -defterm -no-start`

MSYS2 DOES NOT start and it prints at DOS PROMPT:

`Impossibile trovare l'etichetta batch specificata - checkparams`

(`Impossible to find the specified label batch - checkparams`).

The previous version of `msys2_shell.cmd` worked just fine in UNIX style.

tanzislam:
Commit 11ba641bed28 intended to preserve [`cmd.exe` delimiter characters](https://ss64.com/nt/syntax-esc.html#delimiters) in the command line passed on to `bash`, but broke the handling of `-where` directories with spaces in them. Fix this by:

 - Incrementing `shiftCounter` with any quoted arguments treated as separate words (after unquoting), because the `FOR /F ... IN`-loop at the end cannot treat them (in the full command) as single parameters.
 - Using the full list of `cmd.exe` delimiters in the `FOR /F ... IN`-loop when skipping the initial tokens.

Tested this with:

```
C:\msys2\msys2_shell.cmd -defterm -no-start -mingw64 -where <DIR> -shell bash -c $@ -- echo hello there, --world=windows
```

where `<DIR>` is the following:

 - `C:\Users`
 - `C:\Progra~1`
 - `"C:\Program Files"`
 - `"C:\Program Files (x86)"`
 - `"C:\Users\T-17\Documents\Projects and Learning\test,dir"`

Also:

 - Prevent a stray warning if the script is run without any arguments. (This was also introduced in 11ba641.)
 - Ensure (using [delayed expansion](https://ss64.com/nt/delayedexpansion.html)) that the non-option arguments passed on to Bash are intact, even if they contain spaces, parentheses or [`cmd.exe` metacharacters](https://ss64.com/nt/syntax-esc.html#escape).

Fixes #1826

### DIFF
--- a/filesystem/msys2_shell.cmd
+++ b/filesystem/msys2_shell.cmd
@@ -4,7 +4,7 @@ setlocal EnableDelayedExpansion
 set "WD=%__CD__%"
 if NOT EXIST "%WD%msys-2.0.dll" set "WD=%~dp0usr\bin\"
 set "LOGINSHELL=bash"
-set /a shiftCounter=0
+set /a msys2_shiftCounter=0
 
 rem To activate windows native symlinks uncomment next line
 rem set MSYS=winsymlinks:nativestrict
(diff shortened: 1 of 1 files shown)


tanzislam:
This will need more work, as @angelog0 and @xvn found a couple of bugs in their testing (doesn't start without any arguments, and corrupted parentheses in Bash's command-line). Please hold off on approving this for now.

PR #1844, opened 2020-01-29
gmp 6.2.0

JPeterMugaas:
Upgrade to latest version
Use Cygwin patch
added separate msys patch.
use the helper macros


gmp/PKGBUILD
```diff
@@ -11,21 +11,45 @@ url="https://gmplib.org/"
...
+  apply_patch_p2_with_msg gmp-6.2.0-1.src.patch
+  apply_patch_with_msg gmp-6.2.0-msys2.patch
 
   autoreconf -fi
 }
 
 build() {
   cd "${srcdir}/${pkgname}-${pkgver}"
+# get rid of those pesky declspec warnings
+  CFLAGS+=" -Wno-attributes" \
```
lazka: Doesn't this hide potential bugs? What's the point?

JPeterMugaas: That's something Cygwin was doing.  We are probably not talking about errors but warnings.

PR #1846, opened 2020-02-01
ninja: Update to 1.10.0

Biswa96:


Biswa96:
Do we need emacs as a build dependecy?

ninja/PKGBUILD
```diff
@@ -2,18 +2,18 @@
...
+pkgrel=1
 _descr="Ninja is a small build system with a focus on speed"
 pkgdesc="${_descr}"
 arch=('i686' 'x86_64')
 url="https://ninja-build.org"
 license=('Apache')
 depends=()
 options=('strip' 'staticlibs')
-makedepends=("re2c" "python3")
+makedepends=("re2c" "python3" "emacs")
```
lazka: why?

Biswa96: `package_ninja-emacs` is not built without it.

lazka: ah, right, thanks!

Issue #1849, opened 2020-02-08
scons: Support Python 3


pal1000:
Mesa3D intends to drop Python 2 support with [this merge request](https://gitlab.freedesktop.org/mesa/mesa/merge_requests/3674) which will demolish Mesa3D build with Scons in MSYS2 due to [scons depending on Python 2](https://packages.msys2.org/package/scons?repo=msys&variant=x86_64). While it's true I migrated mingw-w64-mesa package to Meson, the migrated package is affected by [this problem](https://gitlab.freedesktop.org/mesa/mesa/issues/2012). While this is not a big deal when Mesa built with MSYS2 is used inside MSYS2, it becomes a problem when used outside.

lazka:
Sounds good.

lazka:
turns out this isn't that easy, things like nsis ship their own scons scripts that only work with Python 2.

lazka:
done

lazka:
new scons is now in the repo

pal1000:
I know. I am testing Mesa build with it and it looks like there won't be problems.

PR #1852, opened 2020-02-13
scons: build with Python 3

lazka:
See #1849


lazka:
packages using scons also depend on Python 3 API, so these need to be ported before this can be merged.

lazka:
This should be ready to merge now.

Issue #1847, opened 2020-02-04
Replacing bzr with breezy


lazka:
We could follow Arch here to get rid of the Python 2 dependency: https://lists.archlinux.org/pipermail/arch-dev-public/2020-February/029858.html

lazka:
Done in  #1854

PR #1854, opened 2020-02-15
Replace bzr with breezy

lazka:
See #1847

Tested by cloning the breezy repo itself.


lazka:
We have hardly any bzr using packages (I count one), so the risk/impact of this should be low..

PR #1861, opened 2020-02-21
libxml2: Fix dependencies (compile-time and run-time)

jannick0:
The post_install hook of the package calls 'xmlcatalog.exe' which is part
of the package.  This commit fixes run-time dependency errors which occur
upon installing and using the package.

* libxml2/PKGBUILD:
  - add required package dependencies, implied by make-dependencies
  - remove trailing white spaces

### DIFF
--- a/libxml2/PKGBUILD
+++ b/libxml2/PKGBUILD
@@ -7,7 +7,7 @@ pkgdesc="XML parsing library, version 2"
 arch=(i686 x86_64)
 license=('MIT')
 depends=('coreutils' 'icu>=59.1' 'liblzma' 'libreadline' 'zlib')
-makedepends=('python2' 'python3' 'icu-devel>=59.1' 'libreadline-devel' 'ncurses-devel' 'liblzma-devel' 'zlib-devel'
+makedepends=('python2' 'python3' 'icu-devel>=59.1' 'libreadline-devel' 'liblzma-devel' 'zlib-devel'
         'libcrypt-devel')
 url="http://www.xmlsoft.org/"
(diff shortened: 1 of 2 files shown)


Biswa96:
Don't we need to incrment pkgrel?

Biswa96:
Don't we need to increment `pkgrel`?

jannick0:
> Don't we need to increment `pkgrel`?

Absolutely. Thanks - will force-push shortly.

lazka:
The dependencies are set in package_libxml2() so your change doesn't do anything. If a dependency is missing add it there.

jannick0:
@lazka Just force-pushed new series of commits which pass all checks. They display changes in tiny steps which you could squash when merging into the repo if OK.  Otherwise I could squash them, too, in case you prefer. 

> The dependencies are set in package_libxml2() so your change doesn't do anything. If a dependency is missing add it there.

Here two issues which emerged after delving into this:
* for split packages the global `depends` array needs to be defined to be passed on to `makepkg`'s dependency checks (see reference in one of the commit messages)
* I removed indirect dependencies to `ncurses` and `ncurses-devel` from this package. BTW: There are far many more throughout the repo packages.  Do you intend to remove them for cleanup or are you happy as it is right now?  I am not sure how this impacts `pacman` which, however, identifies indirect dependencies by traversal of the dependency tree upon install/upgrade (as far as I can tell).

lazka:
I've created #3364 for the second commit.

For (1): we already makedepend on the devel packages, so this doesn't add anything.
(3) this will delete the catalog file and anything other packages like docbook-xml will write in there. While I don't know if the current approach is good, this is definitely wrong.

PR #1865, opened 2020-02-22
perl-Scalar-List-Utils: update to 1.54

Biswa96:


Biswa96:
This compiles in my PC but not in CI !?

Biswa96:
This compiles in my PC but not in CI !? Do we need `libcrypt-devel` as make depends?

lazka:
> Do we need `libcrypt-devel` as make depends?

yes, please try that

Biswa96:
Done.

PR #1866, opened 2020-02-23
Python 2 Removal - Round 1

lazka:


lazka:
Only the 32bit build fails atm, not sure why yet

Biswa96:
What about the build failure?

lazka:
Not sure, I was about to remove the gpgme change. But it likely failed to build before this because I didn't change anything there, so it's at least not making it worse.

PR #1868, opened 2020-02-24
Remove double quotes of the LOGINSHELL

hiroakit:
When set shell arguments, LOGINSHELL has double quotes.
In the default, L6: set "LOGINSHELL=bash". 
I followed the format it.

The task I encountered are shown below.
fig1

fig2


hiroakit:
Thank you for the merge.

PR #1875, opened 2020-02-26
setuptools: split package into Python 2 and Python 3 versions

lazka:
Also remove all dependencies, setuptools bundles all dependencies
by itself anyway.

python3-setuptools -> python-setuptools


lazka:
32bit segfaults, I can't reproduce locally so no idea

lazka:
The CI runner complains about DLL rebasing not succeeding, maybe that's the problem. I don't know enough about that though..

PR #1882, opened 2020-03-04
Make `rebaseall` more useful, especially with regards to `./configure`

dscho:
I've seen my share of failures of this form:

```
      0 [main] perl 5228 child_info_fork::abort: address space needed by 'Cwd.dll' (0x30000) is already occupied
Can't fork, trying again in 5 seconds at /usr/share/autoconf/Autom4te/General.pm line 307.
```

The reason seems that `Cwd.dll` was omitted when rebasing. Let's not omit it from now on, or for that matter, all other native Perl modules.

### DIFF
--- a/rebase/001-rebase-4.4.1-msys2.patch
+++ b/rebase/001-rebase-4.4.1-msys2.patch
@@ -243,7 +243,7 @@ diff -urN rebase-4.4.1.orig/rebaseall.in rebase-4.4.1/rebaseall.in
      done
 +	# some interpreters include a method for installing addons outside of the
 +    # package manager, such as CPAN and RubyGems.
-+    for d in /usr/lib/perl5/site_perl /usr/lib/py*/site-packages \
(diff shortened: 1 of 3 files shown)


dscho:
@mingwandroid you added this in d8a2175dfc18821f167b307887ff8530686c544b, could you have a look whether this change is correct from your point of view, too?

lazka:
I'm a bit confused, as far as I see the first loop (search for "for f in /usr/bin /usr/lib") already adds everything in /usr/lib to the list, so what does this change?

dscho:
@lazka hrm. You're right. And I _still_ get those problems even after this patch.

Funnily, what works around those issues is this:

```sh
 rebase -b 0x63070000 /usr/lib/perl5/core_perl/auto/*/{*,*/*}.dll
```

Any idea what is going on there? Why is `autorebase.bat` not enough to fix this problem?

dscho:
I'll just close this PR, the workaround seemed to be good enough in Git for Windows for the past two years.

PR #1876, opened 2020-02-27
Add China mirrors

taozuhong:

PR #1886, opened 2020-03-07
pacman-mirrors: Fix PRC mirrors

lhmouse:
Signed-off-by: Liu Hao <lh_mouse@126.com>


Biswa96:
If this is correct then this need some extra edits as stated here https://github.com/msys2/MSYS2-packages/pull/1876.

lhmouse:
Thanks. Updated.

PR #1905, opened 2020-03-27
perl-HTTP-Daemon: update to 6.06

Biswa96:


Alexpux:
@Biswa96 you need add new dependencies needed for HTTP-Daemon

lazka:
@Alexpux which one do you mean? `perl-module-build-tiny` which was added in Arch?

Biswa96:
`perl-ExtUtils-Helpers` package may need some patching. May you add that dependency?

Biswa96:
`perl-module-build-tiny` needs `perl-ExtUtils-Helpers` package which may need some patching. May you add that dependency?

Biswa96:
The dependencies of `perl-module-build-tiny` may need some patching because the source files only has `cygwin` mentioned, not msys. May you add that dependency?

hiroakit:
Ref #1914

Alexpux:
> 
> 
> The dependencies of `perl-module-build-tiny` may need some patching because the source files only has `cygwin` mentioned, not msys. May you add that dependency?

This need add bunch of new packages. I dont have time for it now

PR #1910, opened 2020-03-29
pacman-mirrors: add mirror.selfnet.de

carrotIndustries:


Biswa96:
Would it be better to use `https` links?

carrotIndustries:
> Would it be better to use https links?

Good idea, updated the PR accordingly.

PR #1912, opened 2020-04-01
git: Gui fixes

podsvirov:
Some fixes to allow using `git gui` command in `mingw` subsystems.

### DIFF
--- a/git/git-2.26.0-gui-fixes.patch
+++ b/git/git-2.26.0-gui-fixes.patch
@@ -1,0 +1,91 @@
+--- a/git-gui/git-gui.sh	2020-03-23 02:50:46.000000000 +0300
++++ b/git-gui/git-gui.sh	2020-04-01 12:24:15.278000000 +0300
+@@ -46,12 +46,41 @@
+ 
+ ######################################################################
+ ##
++## platform detection
++
++set _iscygwin {}
++
++proc is_Cygwin {} {
++	global _iscygwin
++	if {$_iscygwin eq {}} {
++		if {$::tcl_platform(platform) eq {windows}} {
++			if {[catch {set p [exec cygpath --windir]} err]} {
++				set _iscygwin 0
++			} else {
++				set _iscygwin 1
++			}
++		} else {
++			set _iscygwin 0
++		}
++	}
++	return $_iscygwin
++}
++
++######################################################################
++##
+ ## locate our library
+ 
+ if { [info exists ::env(GIT_GUI_LIB_DIR) ] } {
+ 	set oguilib $::env(GIT_GUI_LIB_DIR)
+ } else {
+-	set oguilib {@@GITGUI_LIBDIR@@}
++	if {[is_Cygwin]} {
++		set oguilib [exec cygpath \
++			--windows \
++			--absolute \
++			@@GITGUI_LIBDIR@@]
++	} else {
++		set oguilib {@@GITGUI_LIBDIR@@}
++	}
+ }
+ set oguirel {@@GITGUI_RELATIVE@@}
+ if {$oguirel eq {1}} {
+@@ -163,7 +192,6 @@
+ set _gitexec {}
+ set _githtmldir {}
+ set _reponame {}
+-set _iscygwin {}
+ set _search_path {}
+ set _shellpath {@@SHELL_PATH@@}
+ 
+@@ -266,26 +294,6 @@
+ 	return 0
+ }
+ 
+-proc is_Cygwin {} {
+-	global _iscygwin
+-	if {$_iscygwin eq {}} {
+-		if {$::tcl_platform(platform) eq {windows}} {
+-			if {[catch {set p [exec cygpath --windir]} err]} {
+-				set _iscygwin 0
+-			} else {
+-				set _iscygwin 1
+-				# Handle MSys2 which is only cygwin when MSYSTEM is MSYS.
+-				if {[info exists ::env(MSYSTEM)] && $::env(MSYSTEM) ne "MSYS"} {
+-					set _iscygwin 0
+-				}
+-			}
+-		} else {
+-			set _iscygwin 0
+-		}
+-	}
+-	return $_iscygwin
+-}
+-
+ proc is_enabled {option} {
+ 	global enabled_options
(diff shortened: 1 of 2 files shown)


podsvirov:
Hello @prati0100! Can your review this patch? Can we apply this changes to git-gui code base if it will work for others?

dscho:
Thank you for working on this!

However, this is unfortunately the wrong repository: MSYS2-packages *never* take mingw-w64-* dependencies.

Also, the Git GUI patch should be made against the `master` branch of https://github.com/prati0100/git-gui and be submitted using [GitGitGadget](https://gitgitgadget.github.io/).

podsvirov:
> Thank you for working on this!

It's working for you? Can you use `git gui` command from `mingw32/64` console afrer installing `git` and `mingw-w64-i686/x86_64-tk` packages?

> However, this is unfortunately the wrong repository: MSYS2-packages _never_ take mingw-w64-* dependencies.

It's no problem: we can drop optional dependencies from `PKGBUILD` file. But @Alexpux can we keep optional dependencies for user information.

> Also, the Git GUI patch should be made against the `master` branch of https://github.com/prati0100/git-gui and be submitted using [GitGitGadget](https://gitgitgadget.github.io/).

After testing in this distro, I will be create `PR` for appropriate repository.

dscho:
> It's working for you?

I haven't tried, due to serious lack of time (and I still think that this is the wrong repository, MINGW-packages seems much more appropriate).

podsvirov:
> > It's working for you?
> 
> I haven't tried, due to serious lack of time (and I still think that this is the wrong repository, MINGW-packages seems much more appropriate).

I understood (and it's not possible, because `git` available only in `MINGW-packages` :-) ).

dscho:
That's right, because `git` is an MSYS2 package. Meaning: it links to the MSYS2 runtime, a POSIX emulation layer based on the Cygwin runtime.

But you claim to be interested in MINGW, i.e. a Git package that does _not_ rely on that POSIX emulation layer.

What you want, therefore, is the `mingw-w64-git` package, in https://github.com/msys2/MINGW-packages.

podsvirov:
> That's right, because `git` is an MSYS2 package. Meaning: it links to the MSYS2 runtime, a POSIX emulation layer based on the Cygwin runtime.

All good.

> But you claim to be interested in MINGW, i.e. a Git package that does _not_ rely on that POSIX emulation layer.

No, because this `PR` about fixes to run `git gui` command. In this case `git` will run `git-gui` script, it's `tcl/tk` based application. To run `git-gui` script we need `wish` enterpreter, thats available in `mingw-w64-i686/x86_64-tk` package (in `mingw32/64` subsystem).

> What you want, therefore, is the `mingw-w64-git` package, in https://github.com/msys2/MINGW-packages.

No, I not want this now.

lazka:
We can't really depend on mingw packages. We could handle the cases when they are missing gracefully and suggest to user to install them.

podsvirov:
> We can't really depend on mingw packages. We could handle the cases when they are missing gracefully and suggest to user to install them.

Can you show examples of how to do this?

podsvirov:
The mingw `tk` package required for working the `git gui` command from `mingw` shell. And optionaly `aspell` package from mingw or msys additionaly can be used by `git gui` to spellcheck the commit message.

lazka:
Not sure. What happens if tk and aspell are missing and you call "git gui"?

podsvirov:
> Not sure. What happens if tk and aspell are missing and you call "git gui"?

Something like:
```bash
$ git gui
/usr/lib/git-core/git-gui: line 10: exec: wish: not found
```

podsvirov:
> We can't really depend on mingw packages.

The optional dependencies on mingw packages removed.

podsvirov:
More questions, reviews, feedback?

If not, I propose to merge as is and end the discussion.

git/git-2.26.0-gui-fixes.patch
```diff
@@ -0,0 +1,91 @@
+--- a/git-gui/git-gui.sh	2020-03-23 02:50:46.000000000 +0300
```
dscho, line 1: I am still of the rather strong opinion that this should be contributed to the Git GUI project by applying this patch on top of https://github.com/gitgitgadget/git's `git-gui/master`, then opening a PR there and submitting it to the Git mailing list for review via the `/submit` command in the PR.

The reason why I say this is that the MSYS2 maintainer (who could merge this patch here) can hardly be asked to review this patch touching idiosyncratic Tcl/Tk code.

And once it gets merged by @prati0100 (the Git GUI maintainer), it is just a matter of time for it to trickle down into v2.27.0 and then we do not need _any_ change here.

podsvirov: @dscho, done. Please, see https://github.com/gitgitgadget/git/pull/612. Can you help me to `/allow` on GitGitGadget?

dscho: Done.

prati0100:
Fixed upstream: https://github.com/prati0100/git-gui/commit/88db24d724619d6c2e602c3b56016f59326d31e3

git/git-2.26.0-gui-fixes.patch
```diff
@@ -0,0 +1,91 @@
...
++
++######################################################################
++##
+ ## locate our library
+ 
+ if { [info exists ::env(GIT_GUI_LIB_DIR) ] } {
+ 	set oguilib $::env(GIT_GUI_LIB_DIR)
+ } else {
+-	set oguilib {@@GITGUI_LIBDIR@@}
++	if {[is_Cygwin]} {
```
lazka, line 35: imo this should be something like:

```tcl
proc is_MingwTcl {} {
	# Handle MSys2 which is only cygwin when MSYSTEM is MSYS.
	if {[info exists ::env(MSYSTEM)] && $::env(MSYSTEM) ne "MSYS"} {
		return 1;
	}
	return 0;
}

if {[is_Cygwin] && [is_MingwTcl]} {
...
```

to not break real cygwin

lazka:
Commented on the upstream PR, and merged a simpler variant of this in #2886

so, closing.

podsvirov:
@lazka , thanks for review, update and apply this suggestion!

PR #1920, opened 2020-04-11
perl-PerlIO-gzip: Add version 0.20

marknsikora:
New dependency required by lcov


marknsikora:
Not sure what the correct solution to the missing crypt header is, the include is coming from a core Perl header and none of the other perl packages are declaring libcrypt-devel as a dependency.

Alexpux:
@marknsikora add libcrypt-devel to makedepends

PR #1923, opened 2020-04-17
dtc: Add libyaml to the dependency

lbmeng:
With a fresh installation of MSYS2, running dtc I got:

$ dtc
C:/msys64/usr/bin/dtc.exe: error while loading shared libraries:
msys-yaml-0-2.dll: cannot open shared object file: No such file or directory

Add libyaml to the dependency to fix it.

### DIFF
--- a/dtc/PKGBUILD
+++ b/dtc/PKGBUILD
@@ -2,13 +2,14 @@
 
 pkgname=dtc
 pkgver=1.6.0
-pkgrel=1
+pkgrel=2
 pkgdesc="Device Tree Compiler"
 arch=('i686' 'x86_64')
 url="https://git.kernel.org/pub/scm/utils/dtc/dtc/git"
 license=('GPL2')
 groups=('base')
-makedepends=('python' 'swig' 'libcrypt-devel')
+depends=('libyaml')
+makedepends=('python' 'swig' 'libcrypt-devel' 'libyaml-devel')
 source=("https://git.kernel.org/pub/scm/utils/dtc/dtc.git/snapshot/${pkgname}-${pkgver}.tar.gz"
         01-correct-shared-library-extension.patch
         02-pyext-install.patch)


dtc/PKGBUILD
```diff
@@ -8,6 +8,7 @@ arch=('i686' 'x86_64')
 url="https://git.kernel.org/pub/scm/utils/dtc/dtc/git"
```
lazka, line 8: increase pkgrel

dtc/PKGBUILD
```diff
@@ -8,6 +8,7 @@ arch=('i686' 'x86_64')
...
 license=('GPL2')
 groups=('base')
+depends=('libyaml')
 makedepends=('python' 'swig' 'libcrypt-devel')
```
lazka, line 12: needs libyaml-devel in makedepends

Biswa96:
The pkgbuild file in ArchLinux does not have that dependency. Also, did you check the dependencies of the program with a tool like Dependency Walker?

lbmeng:
> The pkgbuild file in ArchLinux does not have that dependency. Also, did you check the dependencies of the program with a tool like Dependency Walker?

Yes, Dependency Walker shows msys-yaml-0-2.dll as a dependency. Although the ldd does not complain.

```
$ ldd /usr/bin/dtc.exe
        ntdll.dll => /c/windows/SYSTEM32/ntdll.dll (0x76f40000)
        kernel32.dll => /c/windows/system32/kernel32.dll (0x76e20000)
        KERNELBASE.dll => /c/windows/system32/KERNELBASE.dll (0x7fefcce0000)
        msys-2.0.dll => /usr/bin/msys-2.0.dll (0x180040000)
```

Not sure if this is a bug of the "ldd" in MSYS2?

dtc/PKGBUILD
lbmeng: Actually without 'libyaml-devel' in makedepends, the dtc still can be built successfully, but the weird thing is that the executable requires msys-yaml-0-2.dll.

lazka:
> The pkgbuild file in ArchLinux does not have that dependency. Also, did you check the dependencies of the program with a tool like Dependency Walker?

Looking at the source it (optionally) checks for yaml-0.1.pc.

PR #1925, opened 2020-04-17
Python 3 porting:  libxml2, itstool, libxslt

lazka:
* Move libxml2 and itstool to Python 3
* Drop the python 2 bindings from libxslt, looks like nothing really needs them

### DIFF
--- a/itstool/PKGBUILD
+++ b/itstool/PKGBUILD
@@ -2,12 +2,12 @@
 
 pkgname=itstool
 pkgver=2.0.6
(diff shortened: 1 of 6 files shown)


jeroen:
I think you forgot:

https://github.com/msys2/MSYS2-packages/blob/85d54481612aea963af4ce822a6d9d98cb2a189b/libxml2/libxml2-python.install#L1-L2

lazka:
thanks

PR #1927, opened 2020-04-19
WIP: gdb: rebuild for Python 3.8

### LINKED ISSUE #1926: [BUG] Python update to 3.8.2 broke gdb
## Describe the bug

GDB wants a library named `msys-python3.7m.dll` which is no longer there after the Python update.

## Steps to Reproduce the Problem

Run
```sh
$ gdb --version
C:/msys64/usr/bin/gdb.exe: error while loading shared libraries: msys-python3.7m.dll: cannot open shared object file: No such file or directory
```

Expected behavior: the GDB version is printed.

lazka:
Fixes #1926

### DIFF
--- a/gdb/PKGBUILD
+++ b/gdb/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=gdb
 pkgver=8.2.1
-pkgrel=3
+pkgrel=4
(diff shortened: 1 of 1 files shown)


lazka:
Looks like we need to update first to et py3.8 support

PR #1940, opened 2020-05-04
u-boot-tools: Bug fixes for Flattened Image Tree generation by mkimage

lbmeng:
This adds a patch to fix the Flattened Image Tree (FIT) generation
on Windows. The same patch is already in the upstream U-Boot head
for next release, but not in the current v2020.04 release MSYS2 has.

While we are here, add dtc to the dependency as during FIT generation
the Device Tree Compiler (dtc) will be called by 'mkimage'.


lbmeng:
Ping?

PR #1953, opened 2020-05-16
Upgrade the MSYS2 runtime to v3.1.4

dscho:
This is based on msys2/Cygwin#4 and msys2/Cygwin#5 (and assumes that the latter will be merged unmodified). I verified that it works by initializing a fresh 64-bit MSYS2 environment and building the `msys2-runtime` package.


dscho:
The builds fail because https://github.com/msys2/MSYS2-packages/blob/master/ci-build.sh#L30 tries to install the `msys2-runtime` from within a Bash which already links to the MSYS2 runtime, so the failure is, like, totally expected.

piotrvv:
Is this a msys2-runtime bug? 
[perl.check.log](https://github.com/msys2/MSYS2-packages/files/4642353/perl.check.log) I just added a patch from cygwin perl 5.30.2

piotrvv:
Please remove the file from this package /usr/include/ndbm.h  - libgdbm-devel

dscho:
@Alexpux please note that v3.1.4 no longer comes with an `xlocale.h` and therefore `perl` needs to be rebuilt so that native Perl packages don't fail building (because they try to include `xlocale.h`).

PR #1956, opened 2020-05-17
bison: Update to 3.6.2

jannick0:


jannick0:
@lazka Apparently all checks have passed, but I am uncertain if the package checks were run, since I cannot find anything about them in the logs.  Are they run in the background or am I missing something here?

Thanks.

lazka:
@jannick0 we don't run the tests in CI, because for most packages they fail anyway

jannick0:
Many thanks for responding so quickly.
 
> @jannick0 we don't run the tests in CI, because for most packages they fail anyway

I was hoping to have CI run the tests more quickly than my Windows box.

> Do they work for you locally?

The local check run took very very long today.  Unfortunately, still one single test case fails for a rather simple reason as to how this test is currently configured I believe.  Let me check back with the bison team.

`bison` is such a fundamental application.  What we could do is to move `make check` into the `build` function to ensure that all the relevant checks really work. Agree?  I could do that for the next release (`> 3.6.2`) which presumably covers the last failing test.

lazka:
No, we shouldn't run check in build, that's not what it's there for. I've created #1958

PR #1964, opened 2020-05-19
msys2-runtime: demote the "pseudo console" support to opt-in

dscho:
To make it easier to experiment with it, we now also support `MSYS=enable_pcon` (in addition to the equivalent but unintuitive `MSYS=nodisable_pcon`).

This is a companion PR to https://github.com/msys2/Cygwin/pull/6


dscho:
As before, the build failed after the package was built successfully because it could not be installed (`msys-2.0.dll` is still in use when that Bash script is running).

Alexpux:
@dscho increase pkgrel

dscho:
> @dscho increase pkgrel

Done.

PR #1968, opened 2020-05-21
gperf 3.1: cleanup

jannick0:
This PR is meant to remove a redundant patch.

NOTE: For checking purposes ONLY `make check` is called in `build` (to force CI to run the checks, since CI does not call `check` currently and calling `makepkg` locally is broken for the time being, unfortunately).  

I will remove this tweak once all tests succeed, such that the PR ultimately reduces to the removal of the obsolete gperf patch.


jannick0:
PR ready for merge. 

Checks run successfully. Tweak removed (see above).

PR #1971, opened 2020-05-22
WIP: pacman: kill all msys processes after a core upgrade

lazka:
After a core upgrade instead of waiting endlessly and hoping that
the user closes all remaining msys processes ask the user if we should
kill them and then just do it.

This iterates over /proc/*/winpid and passes the list to taskkill
which kills all processes, except pacman itself which exits manually.

In case of --noconfirm things get killed automatically.


lazka:
@elieux pacman-contrib build fails now (in CI and locally, maybe because of the pacman meson change?)

Alexpux:
@lazka @elieux Im mention that on gitter. This is because of having
````
/usr/lib/gcc/x86_64-pc-msys/9.3.0/../../../../lib/libintl.a
````
in libalpm.pc

Issue #1976, opened 2020-05-25
Move from a base group to a base meta package


lazka:
See https://lists.archlinux.org/pipermail/arch-dev-public/2019-January/029435.html and https://wiki.archlinux.org/index.php/Meta_package_and_package_group

By having a package we can easily add/remove packages from the base system.
For example we could split pacman and pacman-contrib (which atm get built together, but contrib builds against the previously installed one) without users losing pacman-contrib.

Open questions:

* What will win in case of `pacman -S base` if we have both a group and a package named "base"?

elieux:
> What will win in case of `pacman -S base` if we have both a group and a package named "base"?

I just checked with a dummy package and a dummy repo:

```
$ cat > PKGBUILD
pkgname=("base")
pkgver=1
pkgrel=1
pkgdesc="All your base are belong to us"
url="https://msys2.org/"
arch=('any')
license=('custom')
source=()
sha512sums=()

package() {
    touch "${pkgdir}/BASE"
}
^D

$ makepkg
[...]

$ repo-add test.db.tar.xz base-1-1-any.pkg.tar.xz
[...]

# echo "[test]" >> /etc/pacman.conf
# echo "Server = file://$(pwd)" >> /etc/pacman.conf
# pacman -Sy
[...]

# pacman -S base
resolving dependencies...
looking for conflicting packages...

Package (1)  New Version  Net Change  Download Size

test/base    1-1            0.00 MiB       0.01 MiB

Total Download Size:  0.01 MiB

:: Proceed with installation? [Y/n] n

$ pacman -Si base
Repository      : test
Name            : base
Version         : 1-1
Description     : All your base are belong to us
Architecture    : any
URL             : https://msys2.org/
Licenses        : custom
Groups          : None
Provides        : None
Depends On      : None
Optional Deps   : None
Conflicts With  : None
Replaces        : None
Download Size   : 7.67 KiB
Installed Size  : 0.00 KiB
Packager        : David Macek <david.macek.0@gmail.com>
Build Date      : Mon May 25 22:47:41 2020
Validated By    : MD5 Sum  SHA-256 Sum  Signature

$ pacman -Sg base
base bash
base bash-completion
base bsdcpio
base bsdtar
base bzip2
base coreutils
base curl
base dash
base dtc
base file
base filesystem
base findutils
base flex
base gawk
base gcc-libs
base getent
base getopt
base grep
base gzip
base inetutils
base info
base less
base libargp
base lndir
base mintty
[...]

PR #1979, opened 2020-05-27
Add a base package

lazka:
This is meant as a replacement for the base group and contains the
same packages. If a user has the base package installed (the installer
will do this in the future) then we can add new packages to the base
and users with existing installations will get those packages installed
automatically.

This follows what Arch did last year:
https://www.archlinux.org/news/base-group-replaced-by-mandatory-base-package-manual-intervention-required/

See #1976

### DIFF
--- a/base/PKGBUILD
+++ b/base/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=base
 pkgver=1
-pkgrel=1
+pkgrel=2
 pkgdesc='Minimal package set to define a basic MSYS2 installation'
 url='https://www.msys2.org'
 arch=('any')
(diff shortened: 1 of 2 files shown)


lazka:
@Alexpux thoughts?

lazka:
* I've removed u-boot-tools from the base group -> c8dec678d3891e and from the PR
* I've added zstd and wget in a second commit

Alexpux:
@lazka try add
````source=()````

lazka:
> @lazka try add
> `source=()`

Didn't help

Alexpux:
@lazka add then maybe some README file to source and generate also checksums

lazka:
ok

PR #1983, opened 2020-05-29
base: Remove unneeded Linux-specific packages and dependencies

elieux:


lazka:
lgtm, the only thing I use regularly is pkgfile, but it can always be installed

elieux:
> the only thing I use regularly is pkgfile

Is there something wrong with `pacman -F`?

Alexpux:
pkgfile is small utility so I prefer to leave it in base

elieux:
> pkgfile is small utility so I prefer to leave it in base

I must ask again, what's wrong with `pacman -F`?

lazka:
> Is there something wrong with `pacman -F`?

TIL, never mind then XD

elieux:
> TIL, never mind then XD

Pacman deveopers added that some years ago. It might be a bit slower than pkgfile, but being built in beats that IMO.

PR #1984, opened 2020-05-29
[RFC] Cleanup

elieux:
After we publish an installer with the base *package*, we can remove the base *group*.  This PR also cleans some other stuff up.


lazka:
Isn't msys2-runtime in theory still a dependency, or is there a set of packages which we assume always exists?

elieux:
It's a direct dependency for binaries and indirect one for all other executables. Currently I don't think there is any possible situation where `msys2-runtime` is not present. Even if we don't assume `base` as a whole to be installed, counting on the runtime sounds valid. If a different situation arises, I'd be okay with specifying it as a dependency, although it would have to be done correctly, not just for 10 packages.

lazka:
ok, understand. For other things I'm slightly in favor of explicitly depending on them, otherwise you get scared removing things from a "always there anyway base" thing because of unknown deps.

elieux:
I agree, but in practice we ought to be scared anyway, because dependencies on base (or on base's deps) are hard to find. I'm pretty sure most packages don't declare a dependency on `bash` or `gcc-libs` even though they do depend on them. It's one of the reasons I wanted the new *base* to be as small as reasonably possible at this moment.

Alexpux:
@elieux rebase changes. We dont have msys2-launcher-git anymore

elieux:
Rebased.

PR #1985, opened 2020-05-30
msys2-launcher: Switch to release v1.0

elieux:
Is the `conflicts` and `replaces` correct?


lazka:
lgtm (I think conflicts is not needed because there already is one from the other package, but it shouldn't hurt)

Biswa96:
`PREFIX=/opt/bin/...` ??

msys2-launcher/PKGBUILD
```diff
@@ -0,0 +1,27 @@
...
+pkgdesc="Helper for launching MSYS2 shells"
+arch=('x86_64' 'i686')
+license=('MIT')
+groups=('base')
+backup=({mingw32,mingw64,msys2}.ini)
+conflicts=("${_realname}-git")
+replaces=("${_realname}-git")
+depends=("mintty>=1~2.2.1")
+makedepends=('mingw-w64-cross-gcc')
+source=("https://github.com/msys2/msys2-launcher/archive/v1.0.tar.gz")
```
Alexpux, line 16: Use pkgver instead hardcoding version in source URL

elieux: Thanks. Fixed.

elieux:
> `PREFIX=/opt/bin/...` ??

It's probably a bad name for the variable. The idea is to have a simple way to cross-compile.

Issue #1991, opened 2020-06-01
[BUG] Automatically killing MSYS2 processes when updating doesn't work


HolyBlackCat:
## Describe the bug

After updating some MSYS2 packages, instead of asking you to manually terminate all MSYS2 processes, it now asks for a confirmation do it automatically.

When I confirm, it then tells me that it can't terminate some processes, and then gives me the shell prompt.

If I try to update the packages again (without restarting the shell!), it continues the update as if nothing happened. This can't be good, because the shell was never restarted.

## Steps to Reproduce the Problem

Make a fresh install of MSYS2. Run `pacman -Syuu`.
Eventually this appears:
```
:: To complete this update all MSYS2 processes including this terminal will be closed.
    Confirm to proceed [Y/n]
```

After pressing <kbd>Enter</kbd>, I get following errors:
```
Ошибка: Не удается завершить процесс с идентификатором 6920.
Причина: Данный процесс может быть прерван только насильственно ( с  параметром /f ).
Ошибка: Не удается завершить процесс с идентификатором 6540.
Причина: Данный процесс может быть прерван только насильственно ( с  параметром /f ).
Ошибка: Не удается завершить процесс с идентификатором 5904.
Причина: Данный процесс может быть прерван только насильственно ( с  параметром /f ).
```
Then the shell prompt appears. If you run `pacman -Syuu`, it tries to continue the update (as if you restarted the terminal).

The errors mean something along the lines of: "Can't kill a process with pid X, use `/f` to force-kill it."
The pids correspond to `gpg-agent.exe`, `dirmngr.exe`, and `bash.exe`.

**Expected behavior:**
1. I assume `/f` should be passed to whatever is used to kill the processes.

and/or:

2. If the processes can't be killed, the user should probably be asked to kill the terminal session (like it was done before the update).

## Additional Context

(Not sure if any of this matters.)

- OS: Windows 7 x64
- I'm using ConEmu: `msys2_shell.cmd -conemu -mingw64 -full-path`
- ConEmu shows a little blue&yellow shield next to the tab name, so I suppose the shell is running with admin rights.

I haven't tested this without ConEmu, but can do it if needed.

elieux:
We already call `taskkill /f`.  The elevation is probably what's interfering.  Can you check what context are pacman and the other processes running in (all elevated, all non-elevated, or mixed)?

HolyBlackCat:
[...]

PR #2002, opened 2020-06-05
pacman: Fix core update killall behavior

cbrt64:
This addresses #1991 as I understand it. [The documentation](https://linux.die.net/man/3/execvp) is not very clear on this, but `execvp()`'s argv appears to start at 0, not 1, so "/F" was not being passed at all. Now it kills all the consoles without asking again. :)


lazka:
oops :D

Thank you!

PR #2022, opened 2020-06-23
filesystem: fix upgrade

lazka:
msys2-base was removed from dash and filesystem, but filesystem gets updated
first and then fails beause it would break dash which hasn't been updated yet.

So add the provides back.

This fixes:

:: installing filesystem (2020.02-3) breaks dependency 'msys2-base' required by dash

### DIFF
--- a/filesystem/PKGBUILD
+++ b/filesystem/PKGBUILD
@@ -6,11 +6,13 @@
 PKGEXT='.pkg.tar.xz'
 pkgname=filesystem
 pkgver=2020.02
-pkgrel=3
+pkgrel=4
 pkgdesc='Base filesystem'
 arch=('i686' 'x86_64')
 license=('BSD')
 url='https://sourceforge.net/projects/msys2/'
+provides=('msys2-base')
+replaces=('msys2-base')
 backup=('etc/fstab' 'etc/shells' 'etc/profile'
         etc/bash.bash{rc,_logout} etc/skel/.bash{rc,_profile,_logout}
         etc/nsswitch.conf)


marler8997:
So it sounds like `filesystem` will have to maintain this `provides=('msys2-base')` for the forseeable future if we want to avoid this error when upgrading `filesystem` from the current or previous versions to a newer one.

It seems like the better solution would be to fix `pacman` to handle when provides are removed and the upgraded versions of their dependents (in this case `dash`) no longer require them.

If `pacman` is unable to upgrade a package because it has removed a `provides` that is needed by another package, shouldn't it save that operation and come back to it later after attempting to upgrade the rest of the packages?  In this case, it would fail to upgrade `filesystem` because it no longer provides `msys2-base` and `dash` depends on it, but then it could continue on and eventually upgrade `dash` which will no longer depend on `msys2-base` then it can come back to any packages that previously failed and retry.  It seems like an mechanism like this would be much more robust.

marler8997:
For reference, looks like this is the PR that caused the issue: https://github.com/msys2/MSYS2-packages/pull/1984

lazka:
> If `pacman` is unable to upgrade a package because it has removed a `provides` that is needed by another package, shouldn't it save that operation and come back to it later after attempting to upgrade the rest of the packages? In this case, it would fail to upgrade `filesystem` because it no longer provides `msys2-base` and `dash` depends on it, but then it could continue on and eventually upgrade `dash` which will no longer depend on `msys2-base` then it can come back to any packages that previously failed and retry. It seems like an mechanism like this would be much more robust.

The problem is that we have a "core update group" thingy which gets always update first, which contains filesystem and not dash. pacman can handle this all just fine, we break it by splitting the update into two parts. We need to think about getting rid/reducing the "core update group"..

marler8997:
> The problem is that we have a "core update group" thingy which gets always update first, which contains filesystem and not dash. pacman can handle this all just fine, we break it by splitting the update into two parts. We need to think about getting rid/reducing the "core update group"..

Oh I see, thanks for that explanation.

pqnet:
For anyone ending up here and looking for a workaround: this is fixed if you upgrade manually `dash` by running `pacman -S dash`

lazka:
@Alexpux thanks!

PR #2028, opened 2020-06-28
[RFC] rebase: always kill all running msys2 processes

Artoria2e5:
By using powershell to find and kill all processes using the dll, we can ensure that the rebase will actually work.

A bit of testing is needed with someone on Windows Vista. We should also maybe come up with a switch to disable killing, since the search takes time.

### DIFF
--- a/rebase/autorebase.bat
+++ b/rebase/autorebase.bat
@@ -1,4 +1,8 @@
 @echo off
 
+set dir=%~dp0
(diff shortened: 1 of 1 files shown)


Alexpux:
Autorebase mostly need only for 32-bit MSYS2 that we not support anymore

Artoria2e5:
I just got an effed up runtime update caused by a lingering process. Maybe it should go in the thing that moves the dll?

PR #2041, opened 2020-07-12
allow uppercase MINGW_INSTALLS

### LINKED ISSUE #2040: [pacman] Make MINGW_INSTALLS case insensitive
In a makepkg CI matrix you have to define MINGW64 and mingw64 for MSYSTEM and MINGW_INSTALLS.

Allowing uppercase values for MINGW_INSTALLS would allow us to remove one.

https://github.com/msys2/MSYS2-packages/blob/915946a637e1f2b7e26e32782f3af322009293db/pacman/makepkg-mingw#L62

eine:
Close #2040

In this PR, `MINGW_INSTALLS` is converted to lowercase, before checking if empty (and setting a default value). This allows to use a single parameter in CI, as commented in #2040 and msys2/setup-msys2#40.

/cc @lazka


lazka:
needs a pkgrel bump

eine:
Done.

lazka:
Ah, you also need to update the checksums, sorry I missed that.

Run `updpkgsums` in the PKGBUILD driectory.

eine:
Thanks. I had to install `pacman-contrib`. But, apart from that, it seems to be ok now.

lazka:
> Thanks. I had to install `pacman-contrib`. But, apart from that, it seems to be ok now.

See https://www.msys2.org/news/#2020-06-15-new-base-metapackage-pacman-contrib-is-now-separate

PR #2055, opened 2020-07-24
Remove getopt

lazka:
It's included in util-linux which also has a getopt provides.


lazka:
If someone is missing it please say so. We can always bring it back.

PR #2071, opened 2020-07-30
Fix URL for libmetalink source.

jeremyd2019:
Was getting a 404.


jeremyd2019:
The CI check revealed this package also needed libxml2-devel as a makedepends, so added that too.

Biswa96:
Also increase `pgkrel` value.

106062316:
I suggest you use this github link instead of launchpad
https://github.com/metalink-dev/libmetalink/releases

PR #2076, opened 2020-07-31
reflex: Update to 20200715

jannick0:
* reflex/PKGBUILD:
  - update version to 20200715
  - change to github releases whose unzipped packages have a different
    folder name.
  - add variables _src_dir and _build_dir for better maintenance.

### DIFF
--- a/reflex/PKGBUILD
+++ b/reflex/PKGBUILD
@@ -1,7 +1,7 @@
 
 _realname=reflex
 pkgname=${_realname}
-pkgver=20191123
+pkgver=20200715
 pkgrel=1
 pkgdesc="A variant of the flex fast lexical scanner"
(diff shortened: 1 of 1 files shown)


Biswa96:
Why is the actual source URL removed? It seems that the GitHub repository itself imports the code from invisible-island.net.

jannick0:
Correct - plus the benefit to immediately see the changes between releases on github.  (I will apply the same to the next update of btyacc, too.)

Biswa96:
At least, make the old source link as a comment instead of deleting it. See other PKGBUILD files.

jannick0:
Your politeness is demanding ... done.

lazka:
Not a big fan of the builddir changes, but OK..

PR #2078, opened 2020-08-01
perl-Alien-Build: makedepend on libcrypt-devel

jeremyd2019:
This seems to more properly be a dependency of things that compile against perl, so maybe if a perl-devel package existed it would be a dependency of that


jeremyd2019:
I hit the same error in perl-Net-SSLeay, I'm thinking I should try all perl-* and make this pull request for any of them I find that fail due to this

PR #2079, opened 2020-08-01
Fix perl modules makedepends

jeremyd2019:
This seems to more properly be a dependency of things that compile against perl, so maybe if a perl-devel package existed it would be a dependency of that.

Replacement for PR #2078


jeremyd2019:
Draft because I'm still working my way through packages and expect to push more commits

jeremyd2019:
I think that's all the perl modules I need to build anyway, so I'm going to mark this ready for review.

lazka:
Thanks. A dummy perl-devel sounds like a good idea if want to work on that. Same for python-devel.

PR #2082, opened 2020-08-02
python: add libcrypt-devel to makedepends

jeremyd2019:
The 'crypt' module was no longer being built, which was a regression.

### DIFF
--- a/python/PKGBUILD
+++ b/python/PKGBUILD
@@ -4,14 +4,14 @@
 pkgbase=python
 pkgname=python
 pkgver=3.8.5
-pkgrel=2
+pkgrel=3
 _pybasever=${pkgver%.*}
 pkgdesc="Next generation of the python high-level scripting language"
 arch=('i686' 'x86_64')
 license=('custom')
 url="https://www.python.org/"
 depends=('libbz2' 'libexpat' 'libffi' 'liblzma' 'ncurses' 'libopenssl' 'libreadline' 'mpdecimal' 'libsqlite' 'zlib')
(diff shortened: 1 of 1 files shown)


jeremyd2019:
Annoyingly, python's build doesn't fail due to failed module builds, it only lists the modules that failed, which becomes buried in the build output.

From my pull request's check:
```
Python build finished successfully!
The necessary bits to build these optional modules were not found:
_dbm                  _gdbm                 _tkinter           
_uuid                 nis                   ossaudiodev        
spwd                                                           
To find the necessary bits, look in setup.py in detect_modules() for the module's name.


The following modules found by detect_modules() in setup.py, have been
built by the Makefile instead, as configured by the Setup files:
_abc                  atexit                parser             
pwd                   time          
```

From a check before adding libcrypt-devel as a makedepends:
```
Python build finished successfully!
The necessary bits to build these optional modules were not found:
_dbm                  _gdbm                 _tkinter           
_uuid                 nis                   ossaudiodev        
spwd                                                           
To find the necessary bits, look in setup.py in detect_modules() for the module's name.


The following modules found by detect_modules() in setup.py, have been
built by the Makefile instead, as configured by the Setup files:
_abc                  atexit                parser             
pwd                   time                                     


Failed to build these modules:
_crypt
```

lazka:
Thanks. We have a simple test file for the mingw version to check such build issues: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python/smoketests.py I guess we should do something similar for the msys one.

PR #2090, opened 2020-08-07
filesystem: Override %SHELL% for bash

### LINKED ISSUE #2089: [BUG] msys2_shell.cmd does not respect `$SHELL`
## Describe the bug

msys2_shell.cmd defaults to starting `bash`, but does not seem to care whether `%SHELL%` is set in the Windows environment variables, or override it. I assume it is bad practice to set `%SHELL%` in Windows, but it presents a problem when running code that relies on the accuracy of `$SHELL`, as it is just passed straight through even though it may differ from the shell that is actually started.

## Steps to Reproduce the Problem

1. Set `%SHELL%` to `/usr/bin/blah` (something nonexistent)
2. Open MSYS/MinGW terminal
3. Run `echo $SHELL` and `echo $0`

Expected behavior:

msys2_shell.cmd should either:
 - Attempt to use `%SHELL%` as the login shell if available and let it fail on nonexistent binaries/garbage values; or
 - Override it with `/usr/bin/bash` by default, provided `-shell` has not been passed.

## Additional Context: Operating System, Screenshots

https://github.com/qmk/qmk_firmware/issues/9950

fauxpark:
Fixes #2089.

In the event that `%SHELL%` is set in the Windows environment, the shell which msys2_shell.cmd starts (bash by default) and the value of `$SHELL` may differ. Thus when we detect bash in `filesystem/profile`, set `$SHELL` to bash, just to be sure.

### DIFF
--- a/filesystem/PKGBUILD
+++ b/filesystem/PKGBUILD
@@ -6,7 +6,7 @@
 PKGEXT='.pkg.tar.xz'
 pkgname=filesystem
 pkgver=2020.02
-pkgrel=6
+pkgrel=7
 pkgdesc='Base filesystem'
 arch=('i686' 'x86_64')
 license=('BSD')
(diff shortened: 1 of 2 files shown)


lazka:
you need to increase pkgrel in PKGBUILD and run updpkgsums to update the file checksums

Biswa96:
Interesting! CI is OK with the current change.

lazka:
Yes, it only looks for changes in PKGBUILD files as far as I know.

fauxpark:
Whoops, thought I'd done that. Thanks!

PR #2094, opened 2020-08-09
CI: verify signatures of source files

lazka:
With auto-key-retrieve enabled, gpg called by makepkg will fetch keys
from validpgpkeys in the PKGBUILD and verify the result.
Uses the ubuntu keyserver because that one isn't as flaky.

This requires that all our packages get an up-to-date validpgpkeys
array. But we no longer have to trust the users to have checked the
signature before creating a PR.

### DIFF
--- a/ci-build.sh
+++ b/ci-build.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 # AppVeyor and Drone Continuous Integration for MSYS2
 # Author: Renato Silva <br.renatosilva@gmail.com>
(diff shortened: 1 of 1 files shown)


lazka:
@elieux

lazka:
There is one downside to this. Some old packages have signatures which only contain a short key ID and gpg refuses to look them up because they could match multiple keys I guess. I assume they were created with older gpg versions.

So for those packages we'd have to drop the signature or update to a newer release with a better signature if possible. Since this only affects a handful of packages with old releases I think dropping the signature check there is find.

PR #2099, opened 2020-08-11
Add GN, google's metabuilder for Ninja.

Kwizatz:
GN is the metabuilder used to build chromium and v8.

This is part of an effort to also provide scripts to build native mingw v8.

The msys support patch is under review at google at:
https://gn-review.googlesource.com/c/gn/+/9660

if/once the patch gets merged, it can be removed, but since a specific tag is being checked out, there should be no conflicts with upstream even if the patch is merged.

### DIFF
--- a/gn/PKGBUILD
+++ b/gn/PKGBUILD
@@ -1,0 +1,42 @@
+
+pkgname=gn
+pkgver=1822
+pkgrel=1
+pkgdesc="GN is a meta-build system that generates build files for Ninja."
+arch=('i686' 'x86_64')
+url='https://gn.googlesource.com/gn/+/refs/heads/master/README.md'
+license=('BSD')
+depends=('ninja')
+makedepends=('python' 'ninja' 'git')
+source=('gn::git+https://gn.googlesource.com/gn')
+sha256sums=('SKIP')
+
+prepare()
+{
+  cd gn
+  git config advice.detachedHead false
+  git checkout 0649bd9
+  chmod 755 build/gen.py
+  python build/gen.py --platform=msys
+}
+
+build()
+{
+  cd gn
+  ninja -C out
+}
+
+check()
+{
+  cd gn
+  out/gn_unittests
+}
+
+package()
+{
+  cd gn
+  install -d ${pkgdir}/usr/bin
+  install -Dm755 -s out/gn.exe ${pkgdir}/usr/bin
+  install -d ${pkgdir}/usr/share/licenses/gn
+  install -m644 LICENSE ${pkgdir}/usr/share/licenses/gn
+}


lazka:
> This is part of an effort to also provide scripts to build native mingw v8.

So you would use cygwin python/ninja to build mingw v8?

Kwizatz:
@lazka Correct, I have a PR for adding V8 to vcpkg (visual c++) at https://github.com/microsoft/vcpkg/pull/12687

You'll notice that depot tools is not used on that PR, I have already successfully built v8 using mingw64, but I am missing gn for a full "native" build.

The patch has merged upstream by the way 😁

Biswa96:
Is it possible to convert this package to mingw one?

Kwizatz:
@Biswa96 It is possible, but why would you do that? ninja is msys, unless you want to redistribute gn you only need it at build time.

I have not tested whether these changes are sufficient for native mingw builds, my guess would be they are not, since there is a distinct path for Posix and Windows in the code and the code guards targeting MinGW may be stale and require more than adding some ifdefs here and there.

Biswa96:
MinGW packages are native to Windows OS. MSYS2 and Cygwin are compatible POSIX layer. So, MSYS2 programs are somewhat slower than native one. If a package can be compiled as native one then it will be preferred. But it would be better if Windows native build is supported by upstream project.

Kwizatz:
How much slower is slower? GN is like cmake, you run it once, takes a couple of seconds and then the heavy lifting is done by ninja.

Anyway, I see your point, there is msys ninja and mingw ninja, so how about I do both?

Kwizatz:
So, mingw builds with just this patch, there's some warnings but it builds, however there seems to be a bug in  mingw-w64-x86_64-ninja, for some reason the following rule deletes all input files and the static library:

rule alink_thin
  command = rm -f $out && $ar rcsT $out $in
  description = AR $out

seems to be a problem with the "rm -f $out" command since removing that command fixes the build. MSYS ninja does what it is supposed to do.

Biswa96:
There is no downside about building a package both for mingw and msys. But main goal of msys2 project is to build native programs. msys2 package is required when the program relies on POSIX functions and there is no way to convert it to Windows native.

I tried to build the gn program in mingw mode but need some code changes. Can you join in gitter to discuss more?

Biswa96:
There is no downside about building a package both for mingw and msys. But main goal of msys2 project is to build native programs. msys2 package is required when the program relies on POSIX functions and there is no way to convert it to Windows native.

I tried to build the gn program in mingw mode but need some code changes. gn already supports Windows platform, so no significant change is required. Can you join in gitter to discuss more?

Kwizatz:
Sure, I've never use gitter before, I just joined with github so ping me there or add me to a discussion.
I am busy with other stuff right now, so I don't know how responsive I'll be for the next hours.
I created a PKGBUILD for mingw, but it failed recognising the gn.exe generated, may have to do with the install calls.
The one I created directly from the upstream repo did run, so it is not like it is a defective binary.

Kwizatz:
This is how I got it to build assuming you've cloned upstream and have a MINGW64 console already at the root of the repo:

./build/gen.py --platform=mingw
/usr/bin/ninja -C out

Biswa96:
If possible I shall contribute some code in gn repository, then create a pkgbuild and ping you when it is ready. 1-2 days.

Kwizatz:
That would be great, most of the warnings are due to GN redefining some macros in windows_types.h so they don't have to include windows.h, other than that I saw some pointer casting warnings, these can be fixed or silenced, however the bug in native mingw ninja is what worries me most.

Biswa96:
How can I make a pull request there?

Biswa96:
How can I make a pull request or create an issue there? The page has only the repository.

Biswa96:
I don't know how to send pull request there. I follow all the steps but it does not appear in review webpage. If I send the changes can you add it there?

Kwizatz:
Sure, send me patch.

Kwizatz:
So, what else is needed on this one to get merged?

Biswa96:
If gn works as MinGW package then this can be removed.

lazka:
Since this is now part of the v8 built itself I'm closing this.

As @Biswa96 we try to keep msys packages at a minimum and prefer native builds.

If the need for this arises in the future feel free to re-open though and we can have another look.

PR #2101, opened 2020-08-12
Set validpgpkeys for various packages

lazka:


Biswa96:
Though I am not against adding pgp keys, I think those should be added when the individual packages are updated, instead of adding in one commit. Just for make things easy and clean.

lazka:
hm, easier for whom?

I was trying to prevent package updates from failing in CI once I enforce signature checks in #2094.

PR #2102, opened 2020-08-13
ci-build: don't overwrite an existing gpg.conf

jeremyd2019:
Assume that if a gpg.conf exists, that it has proper contents already.

Just a little safety check.


lazka:
But I kinda wanted to override it

jeremyd2019:
It seemed like other parts (like git_config:
https://github.com/msys2/MSYS2-packages/blob/033f6ea17f164c15cb59851c2d8344778069f4d3/ci-library.sh#L113-L114
) avoided overwriting existing configs.  But if there's a reason to overwrite it, that's fine too.

PR #2108, opened 2020-08-16
[RFC] rebase: add a pacman hook on i686.

jeremyd2019:
It will rebase any *.{dll,so,oct} except usr/bin/msys-2.0.dll.  Note it
requires the rebase db to be present, so essentially it requires
rebaseall to have been run (probably at install).

### DIFF
--- a/rebase/PKGBUILD
+++ b/rebase/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=rebase
 pkgver=4.4.4
-pkgrel=3
+pkgrel=4
 pkgdesc="The Cygwin rebase distribution contains four utilities, rebase, rebaseall, peflags, and peflagsall"
 arch=('i686' 'x86_64')
 license=('custom')
@@ -18,7 +18,8 @@ source=(${pkgname}-${pkgver}::git://sourceware.org/git/cygwin-apps/rebase.git#ta
         '005-make-verbose-give-a-reason.patch'
         '006-fix-some-errors.patch'
         '007-peflags-high-entropy-va.patch'
-        'autorebase.bat')
+        'autorebase.bat'
+        'rebase.hook')
 sha256sums=('SKIP'
             '6e25bfaae229062b20bed52b79824cfd1145de10c2ce50cd72b5a82fff9fbacf'
             'b06d561d82e7c32573082cf0207ca80275bab241c61debf2d3d8cc10cf79e31b'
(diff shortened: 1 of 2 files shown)


lazka:
I just have the issue that rebuilding python as is right now results in fork errors until rebaseall is called, with 64bit.. :/

So I'm wondering if we should special case python to locally rebase its modules or try this instead.

jeremyd2019:
> I just have the issue that rebuilding python as is right now results in fork errors until rebaseall is called, with 64bit.. :/
> 
> So I'm wondering if we should special case python to locally rebase its modules or try this instead.

This *should not* happen on 64-bit.  `--enable-auto-image-base` should choose base addresses based on a hash of the file name iirc, and there should be enough bits for that to not conflict.  The fact that the base address was 0x4000... leads me to believe that somehow `--enable-auto-image-base` is not being used, or is not working.

See also #2202.  On 32-bit, there are not enough bits for `--enable-auto-image-base` to not happen to conflict on python modules.

> > Note it
> > requires the rebase db to be present, so essentially it requires
> > rebaseall to have been run (probably at install).
> 
> What happens if there is no db? Rebase to the potentially wrong addr I guess?

It fails (does nothing and exits with a non-zero exit code).  I don't know what pacman does in that case, if a hook 'fails'.  Probably nothing much.

lazka:
The address in error message doesn't match the base address:

```
      0 [main] python 12280 child_info_fork::abort: address space needed by '_ssl.cpython-38-x86_64-msys.dll' (0x1D0000) is already occupied
```

```
$ objdump.exe -p /usr/lib/python3.8/lib-dynload/_ssl.cpython-38-x86_64-msys.dll | grep ImageBase
ImageBase               00000005d76d0000
```

*shrug*

lazka:
Same in CI, I've filed https://github.com/msys2/MSYS2-packages/issues/2487

jeremyd2019:
This has been kicking around for quite a while, and seems to work quite well for me, with one caveat:
if ANY non-static msys2 process is running (such as say, randomly spawned gnupg daemon, or even the /usr/bin/env I use to set `MSYSTEM` when launching through Windows Terminal), it will complain that dlls are in use and could not be rebased.  Even if said dlls were not included in the packages that were updated, which surprised me.  I guess pacman passes ALL matching installed files to the hook, not just those that were in the packages modified in that transaction.  🤷

PR #2109, opened 2020-08-17
python: enable makeflags option

jeremyd2019:
It mostly worked out of the box, except for some ln failures in make install due to not
being able to create a broken symlink to a file to be created later.
Fixed by making the make rule that ln's depend on the rule that creates
the files.


jeremyd2019:
You may notice I didn't bump pkgrel, and that the patch number has a gap from the last one.  This is because this was extracted from i686 experimentation.  For some reason I have not yet determined, python build has issues in the github runner that I do not get on my local machine.  This is one of the things I noticed trying to track that down, and seemed to be generally beneficial.

PR #2110, opened 2020-08-17
CI: don't attempt to install known-troublesome packages.

jeremyd2019:
Namely, msys2-runtime and bash.  Currently, msys2-runtime always fails CI.

### DIFF
--- a/ci-build.sh
+++ b/ci-build.sh
@@ -31,7 +31,7 @@ execute 'Approving recipe quality' check_recipe_quality
 for package in "${packages[@]}"; do
     execute 'Building binary' makepkg --noconfirm --noprogressbar --nocheck --syncdeps --rmdeps --cleanbuild
     execute 'Building source' makepkg --noconfirm --noprogressbar --allsource
-    execute 'Installing' yes:pacman --noprogressbar --upgrade *.pkg.tar.*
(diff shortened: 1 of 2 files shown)


ci-build.sh
```diff
@@ -31,7 +31,7 @@ execute 'Approving recipe quality' check_recipe_quality
 for package in "${packages[@]}"; do
     execute 'Building binary' makepkg --noconfirm --noprogressbar --nocheck --syncdeps --rmdeps --cleanbuild
     execute 'Building source' makepkg --noconfirm --noprogressbar --allsource
-    execute 'Installing' yes:pacman --noprogressbar --upgrade *.pkg.tar.*
+    grep -qFx "$(<../ci-dont-install-list.txt)" <<< "${package}" || execute 'Installing' yes:pacman --noprogressbar --upgrade *.pkg.tar.*
```
1480c1, line 34: interesting way of using the subshell+redirect to use a file as a list of exceptions. Haven't really seen this thing done at all.

jeremyd2019: I'm not quite sure if that was a compliment or a criticism :smiley:

jeremyd2019: had a 🤦 epiphany trying to fall asleep
```bash
grep -qFx "${package}" ../ci-dont-install-list.txt || ...
```
What was I thinking?

Biswa96: [KISS](https://wiki.archlinux.org/index.php/Arch_terminology#KISS)

PR #2114, opened 2020-08-22
Add a python-devel package

lazka:
This doesn't contain anything and just depends on python itself and
libcrypt-devel which it uses in Python.h.

This allows packages building against Python to makedepend on this without
having to specify libcrypt-devel etc.

### DIFF
--- a/python/PKGBUILD
+++ b/python/PKGBUILD
@@ -2,19 +2,15 @@
 # Contributor: Ray Donnelly <mingw.android@gmail.com>
 
 pkgbase=python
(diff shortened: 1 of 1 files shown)


jeremyd2019:
> (not sure if this covers all of them)

It does not.  Also included are things which link to msys-python*.dll.  I know of one I hit personally while building updated i686 packages: gdb.

lazka:
Thanks. I think we can always move the files later, I've opened #2117 for this.

PR #2121, opened 2020-08-24
Add a perl-devel package.

jeremyd2019:
Like the recent python-devel package (#2114), it doesn't contain anything and
just depends on perl itself and libcrypt-devel.


perl/PKGBUILD
```diff
@@ -1,15 +1,13 @@
 # Maintainer: Alexey Pavlov <Alexpux@gmail.com>
 
-pkgname=perl
+pkgname=('perl' 'perl-devel')
```
lazka, line 3: Add `pkgbase=perl`

lazka:
lgtm thanks!

perl/PKGBUILD
jeremyd2019: Must have missed that on the python changes, tried to replicate them.

PR #2125, opened 2020-08-26
Use perl- and python-devel makedepends

jeremyd2019:
A quick grep for libcrypt-devel in PKGBUILDs and a review for using perl or python (or both).  Will take out of draft status when the new packages are published and CI is able to succeed.


cython/PKGBUILD
```diff
@@ -4,13 +4,13 @@ _realname=cython
 pkgbase=${_realname}
 pkgname=('cython')
 pkgver=0.29.21
-pkgrel=1
+pkgrel=2
 pkgdesc='C-Extensions for Python'
 arch=('i686' 'x86_64')
 url='https://cython.org/'
 license=('APACHE')
 depends=('python-setuptools')
```
jeremyd2019, line 12: An argument could be made that cython depends rather than makedepends on python-devel, as its purpose is to generate C/C++ extension code.

perl-Digest-MD4/PKGBUILD
```diff
@@ -3,13 +3,13 @@
...
 pkgver=1.9
-pkgrel=5
+pkgrel=6
 pkgdesc="Perl Module: Perl interface to the MD4 Algorithm."
 arch=('any')
 license=('PerlArtistic')
 url="https://metacpan.org/${_realname}/"
 groups=('perl-modules')
-makedepends=('libcrypt-devel')
+makedepends=('perl-devel' 'libcrypt-devel')
```
jeremyd2019, line 12: I left libcrypt-devel in this one, because Digest-MD4 is old enough that it could actually be using libcrypt.  I should look and see if this is the case or not.

perl-Sys-CPU/PKGBUILD
```diff
@@ -3,12 +3,13 @@
 _realname=Sys-CPU
 pkgname=perl-${_realname}
 pkgver=0.61
-pkgrel=8
+pkgrel=9
 pkgdesc="Provide commonly requested regular expressions"
 arch=('i686' 'x86_64')
 url="https://metacpan.org/author/MZSANFORD${_realname}-${pkgver}"
 groups=('perl-modules')
-depends=('perl' 'libcrypt-devel')
```
jeremyd2019, line 11: Whoops.  Confirmed it doesn't really depend on libcrypt-devel at runtime.

perl-Digest-MD4/PKGBUILD
jeremyd2019: Nope.  Will clean that up.

jeremyd2019:
I noticed that perl-Digest-MD4 had arch=('all') even though it built/included a .dll.  I went through and changed all of the dll-producing perl-* in my pull request to arch=('i686' 'x86_64') as well.  Interestingly, the python modules didn't seem to have that problem.

cython/PKGBUILD
lazka: On the other hand it just generates the C code and doesn't build it, right?

lazka:
> I noticed that perl-Digest-MD4 had arch=('all') even though it built/included a .dll. I went through and changed all of the dll-producing perl-* in my pull request to arch=('i686' 'x86_64') as well. Interestingly, the python modules didn't seem to have that problem.

I guess this wasn't a problem because we used to have two different repos for these packages. But nice to fix nevertheless

lazka:
lgtm!

irssi/PKGBUILD
```diff
@@ -21,6 +21,7 @@ options=('!strip')
 sha256sums=('6727060c918568ba2ff4295ad736128dba0b995d7b20491bca11f593bd857578'
             'SKIP'
             'ec2dc647191717c4abd9b86a51c085e9bb8ca13998726ef0be97f4d213577eb3')
+validpgpkeys=('7EE65E3082A5FB06AC7C368D00CCB587DDBEF0E1') # The Irssi project <staff@irssi.org>
```
jeremyd2019, line 24: From https://github.com/archlinux/svntogit-packages/blob/packages/irssi/trunk/PKGBUILD#L19

mercurial/PKGBUILD
```diff
@@ -19,6 +19,9 @@ sha256sums=('5c8b93da701ee39e312da9e35a7f3163e17ed173a4707857bc467c3b3ab74853'
             'SKIP'
             '87427151713e689cd87dc50d50c048e0e58285815e4eb61962b50583532cbde5'
             '1a35962837bb1eb8893e4ab8aa3a49bb18023b632d4c3a2fa903ebfdd7e598ce')
+validpgpkeys=(2BCCE14F5C6725AA2EA8AEB7B9C9DC824AA5BDD5
+              3A8155163D0E20A530FCB78647A67FFAA346AACE
+              EB851395B4223EE2F7BA0B28DA54740BF08732BA) # Pulkit Goyal <7895pulkit@gmail.com>
```
jeremyd2019, line 24: From https://github.com/archlinux/svntogit-packages/blob/packages/mercurial/trunk/PKGBUILD#L24-L26

PR #2127, opened 2020-08-27
makepkg-mingw: Enable hardening linker flags by default

lazka:
This enables ASLR, DEP by default.

The goal here is to give these a try for our own package builds and
if there are problems disable some flags again in the respective PKGBUILDs.

Once most of our packages work fine we can think about enabling it
in binutils directly for all users.

Related ASLR change for binutils: https://github.com/msys2/MINGW-packages/pull/6880
Related discussion: https://github.com/msys2/MINGW-packages/issues/6674

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -5,7 +5,7 @@
 PKGEXT='.pkg.tar.xz'
 pkgname=pacman
 pkgver=5.2.2
-pkgrel=1
+pkgrel=2
 pkgdesc="A library-based package manager with dependency support (MSYS2 port)"
 arch=('i686' 'x86_64')
 url="https://www.archlinux.org/pacman/"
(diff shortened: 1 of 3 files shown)


lazka:
@jeremyd2019 @mati865

lazka:
To summarize my minimal testing:

Works: gtk3, ninja, harfbuzz, binutils
Crashes at buildtime: python (removing `--high-entropy-va` makes it work)

jeremyd2019:
Something that will probably also need to be added to the list of things that don't work with ASLR is GCC, since it is known that PCH is incompatible with it (on any platform)

lazka:
For future reference, https://github.com/msys2/MINGW-packages/issues/6674#issuecomment-682078943 binutils upstream has these enabled by default now for the next stable release.

jeremyd2019:
I tried building gcc with these (plus --no-seh), but was surprised when the resulting binaries did not have the flags set in their PE headers.  It appears that the LDFLAGS get lost when building later stages of gcc.  But when they are enabled by default in binutils gcc WILL be affected.  I have some ideas for a 'surgical' solution so that gcc's libs don't have to have ASLR disabled.

PR #2128, opened 2020-08-27
mc: Recognize Win32 absolute paths

elieux:
Running `mcedit C:/foo/bar` would result in a message like _Cannot open /usr/bin/C:/foo/bar_.  This patch fixes this use case.


elieux:
Shoot, it seems to crash in some cases.

elieux:
Okay, that was a fun typo. Doesn't crash anymore.

lazka:
lgtm.

I'm wondering, should we have some path auto conversion like for non-cygwin binaries?

mingwandroid:
Maybe, could we override it somehow too?

lazka:
I've created https://github.com/msys2/msys2-runtime/issues/14 so we don't forget. Let's not block this..

PR #2129, opened 2020-08-27
xxhash: Update to 0.8.0

jannick0:


lazka:
The DLL name changed, so the dependent packages need their pkgrel increased. In this case it's only rsync.

jannick0:
@lazka Thanks for this!

BTW: I am intrigued by the problem to identify that the dll interface version changed (which is the case here at hand).  Do you have any tool on the shelves to be alerted when a dll file name is changed?

lazka:
There are no tools atm. I just looked at https://packages.msys2.org/package/libxxhash?repo=msys&variant=x86_64 to see if the version is encoded int he DLL name.

jannick0:
@lazka Best if CI could run this look up job I think. I'll see if I can come up with something soon.

PR #2132, opened 2020-08-28
Update to binutils 2.35

jeremyd2019:
Synced mingw-w64-cross-binutils with msys2/MINGW-packages mingw-w64-binutils patches.

Also removed install script that only dealt with info files, which is now handled by a pacman hook.


Biswa96:
Don't set pkgrel to 2.

Biswa96:
When a new version of the software is released, the `pkgrel` value must be reset to 1.

jeremyd2019:
OK, thought it made sense to match it up with the corresponding 'native' mingw package.

PR #2133, opened 2020-08-29
pacman: add --no-seh to ming32 LDFLAGS.

jeremyd2019:
As mentioned in https://github.com/msys2/MINGW-packages/issues/6674#issuecomment-683138852 a different exception mechanism is used by gcc on 32-bit Windows, and this is already the default in lld.

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -5,7 +5,7 @@
 PKGEXT='.pkg.tar.xz'
 pkgname=pacman
 pkgver=5.2.2
-pkgrel=2
+pkgrel=3
 pkgdesc="A library-based package manager with dependency support (MSYS2 port)"
 arch=('i686' 'x86_64')
 url="https://www.archlinux.org/pacman/"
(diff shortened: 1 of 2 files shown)


jeremyd2019:
I tested building mingw-w64-python with this, because I knew the build process of python exercises the built python quite heavily.

lazka:
So, as far as I understand the other thread this is the default, so what does this change?

jeremyd2019:
It is the default in lld (as in llvm's linker).

lazka:
ah, ok. So the question remains, what does this change?

jeremyd2019:
Perhaps @mati865 can explain it better, but I'll give it a try.
At the lowest level, it sets another bit in the PE header.  What that bit means to Windows is that no SEH (Structured Exception Handling) handler is allowed to be located in the image, preventing memory corruption/buffer overflows from being able to forge an exception handler record and then trigger an exception to hijack execution to code in that image.  GCC does not support SEH, it instead uses either Dwarf2 or SJLJ for exception handling.

mati865:
SEH used by x86 MSVC is vulnerable to attacks so SafeSEH was introduced as opt-out (because of compatibility issues).

Various binary binary analysers have bug where they assume SEH is used for 32-bit binaries. If SafeSEH is not present the binary will be reported as insecure.

To fight that MinGW linkers allow setting `IMAGE_DLL_CHARACTERISTICS_NO_SEH` characteristic. LLD does it by default but LD (BFD) does not.

For 64-bit binaries SEH used by MinGW and MSVC works differently and does not have SafeSEH characteristic at all.

lazka:
Thanks! So this sounds like we should enable this in binutils directly in the long run, right?

Biswa96:
Before compiling any other programs, are binutils > gcc > crt > winpthreads need to be recompiled for this change?

mati865:
> sounds like we should enable this in binutils directly in the long run, right? 

This is just one of the many things binutils doesn't so but should...

> Before compiling any other programs, are binutils > gcc > crt > winpthreads need to be recompiled for this change?

This does not affect how programs work but all the DLLs and executables have to be recompiled to have it enabled.

jeremyd2019:
> > sounds like we should enable this in binutils directly in the long run, right?
> 
> This is just one of the many things binutils doesn't so but should...

I'm not confident in just enabling the flag in binutils, because I feel there's no guarantee that binutils is linking code from a tool that can't generate SEH.  The correct solution would be more like https://reviews.llvm.org/rLLD320860.

PR #2139, opened 2020-09-03
sqlite: don't replace lemon with itself

lazka:
also no reason it should be in base-devel


lazka:
@jannick0 ok?

jeremyd2019:
Fixes #2141

jannick0:
Sorry for being a bit late to the party. ;)

@lazka Yes for removing `replace`, while removing `groups` might impact packages relying on `lemon` in an unexplicit manner.

The former issue begs for a CI test I believe.

lazka:
> @lazka Yes for removing `replace`, while removing `groups` might impact packages relying on `lemon` in an unexplicit manner.

thanks. other packages should rely on the package directly instead.

PR #2145, opened 2020-09-05
rebase: teach peflags about high-entropy-va flag.

jeremyd2019:
This has been my go-to tool to display or modify dll characteristics
flags, but it didn't know about the high-entropy-va flag.  Now it does.

### DIFF
--- a/rebase/007-peflags-high-entropy-va.patch
+++ b/rebase/007-peflags-high-entropy-va.patch
@@ -1,0 +1,165 @@
+diff --git a/peflags.c b/peflags.c
+index 4d22e4a..358199a 100644
+--- a/peflags.c
++++ b/peflags.c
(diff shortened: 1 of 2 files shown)


jeremyd2019:
Not implemented in this patch are warnings for meaningless uses of high-entropy-va: without dynamicbase or on i686 images.  There is an existing section where warnings are output so it would probably make sense to add these.

jeremyd2019:
I sent this upstream, and got it merged, and resulted in a new release of rebase: https://cygwin.com/pipermail/cygwin/2021-May/248534.html.

If nobody beats me to it, I'll do a PR to update rebase here (and probably include my i686 rebase pacman hook cause that has done pretty well for me)

lazka:
Sounds good, thanks.

PR #2156, opened 2020-09-21
New Package: GNU Stow v2.3.1

tavrez:


lazka:
What's is this needed for?

tavrez:
It's a great application for handling symlink files, used for managing dotfiles https://dotfiles.github.io/

1480c1:
I am sort of curious is gnu stow would be good at all since we can't make symlinks (as in actual symlinks) without either admin power or dev mode

siprbaum:
I agree that this needs admin or developer mode, but considering that cygwin is usually used by people I consider power users,
the chances are very high that these permissions are available.

So I would appreciate having this in and would definitely use it for my dotfile handling like I used to under linux

tavrez:
I should also say that stow(and ln) use hard links by default. And hard links does not need admin permission or developer mode.

soredake:
Any progress on this?

fralalonde:
GNU Stow is an critical component of many `.dotfiles` setup. It allows things like storing and recreating your environment from a personal git repo. It is a simple, safe and important component of a modern distribution.

Please merge this.

fralalonde:
I see this has been merged, thanks! 

Just to confirm, I've meanwhile installed `stow` manually from the submitter's branch (using `makepkg`). I then used it to deploy my own dotfiles and it ran without a hitch. I'll reinstall from the now-official pkg on my next update.

Issue #1974, opened 2020-05-23
[BUG] Wrong charset after upgrade to msys2-runtime-3.1.4


nyfair:
## Steps to Reproduce the Problem
`python -c 'print("世界")'`
荳也阜

## Additional Context
`locale`
LANG=en_US.UTF-8
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_ALL=

## The Following Methods Can All Get Correct Result
1. Downgrade msys2-runtime to 3.0.7
2. `winpty python -c 'print("世界")'`
3. `python -c 'print("世界")' | iconv -f utf-8 -t utf-8`
4. `chcp.com 65001 && python -c 'print("世界")'`

mati865:
Cc @dscho

mati865:
@dscho I though it could be potential issue for Git for Windows users. I suppose there is no correlation between this and your latest patches for the runtime?

dscho:
It's possible, but `python` is not shipped with Git for Windows. It's possible that other software is potentially affected, too, of course.

lazka:
I can't reproduce here

dscho:
> I can't reproduce here

Neither can I...

```sh
$ python -c 'print("世界")'
世界
```

nyfair:
I suppose it is a cygwin issue since it changed encoding handing from 3.1
This also affect other programs as well

```
cargo new a && cd a && sed -i 's/world/世界/' src/main.rs && cargo run
     Created binary (application) `a` package
   Compiling a v0.1.0 (D:\msys64\opt\crawl\a)
    Finished dev [unoptimized + debuginfo] target(s) in 0.44s
     Running `target\debug\a.exe`
Hello, 涓栫晫!
```

dscho:
@nyfair it would be helpful if you could figure out how others can reproduce the problem you are seeing. So far the score is that two others _cannot_, while nobody else reproduced your issue yet.

elieux:
@nyfair, your original post mentions various methods to get the correct result, but doesn't say if you're running mintty in the failing case. Other useful information would be the result of `command -v python` and output of `chcp.com`.

fasterthanlime:
Hi! I'm able to reproduce the issue (in mintty).

[...]

PR #2158, opened 2020-09-22
msys2-runtime: Update patches

lazka:
See https://github.com/msys2/msys2-runtime/pull/15

### DIFF
--- a/msys2-runtime/0029-Cygwin-pty-Prevent-garbled-output-when-pseudo-consol.patch
+++ b/msys2-runtime/0029-Cygwin-pty-Prevent-garbled-output-when-pseudo-consol.patch
@@ -1,0 +1,32 @@
+From cfe1db3994b9c466b4bc1fc83d21d98e81a835a0 Mon Sep 17 00:00:00 2001
+From: Takashi Yano <takashi.yano@nifty.ne.jp>
+Date: Fri, 4 Sep 2020 18:31:35 +0900
+Subject: [PATCH 29/N] Cygwin: pty: Prevent garbled output when pseudo console
+ is disabled.
+
+- If pseudo console is disabled, non-cygwin apps do not detect
+  console device. In this case, the some apps may output messages
+  based on the locale. In addition, some apps output UTF-8 string
+  regardless of the locale setting. At least git-for-windows and
+  node.js do that. Even in this case, garbled output is prevented
+  with this patch in most cases because mintty uses UTF-8 by default.
+---
+ winsup/cygwin/fhandler_tty.cc | 2 ++
+ 1 file changed, 2 insertions(+)
+
+diff --git a/winsup/cygwin/fhandler_tty.cc b/winsup/cygwin/fhandler_tty.cc
+index 43eebc1..b9cb45b 100644
+--- a/winsup/cygwin/fhandler_tty.cc
++++ b/winsup/cygwin/fhandler_tty.cc
+@@ -2905,6 +2905,8 @@ fhandler_pty_slave::setup_locale (void)
+ 	  get_ttyp ()->term_code_page = cs_names[i].cp;
+ 	  break;
+ 	}
++  SetConsoleCP (get_ttyp ()->term_code_page);
++  SetConsoleOutputCP (get_ttyp ()->term_code_page);
+ }
+ 
+ void
+-- 
+2.27.0
+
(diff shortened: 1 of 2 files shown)


dscho (approved):

nga888:
I recently ran into an issue with broken output with pseudo console enabled, i.e. `MSYS=enable_pcon`, and I traced it back to this change.

Shouldn't the setting of the console codepage at the end of `setup_local` be conditional on pseudo console being disabled as suggested by @tyan0 in https://github.com/msys2/MSYS2-packages/issues/1974#issuecomment-686837959?

dscho:
> I recently ran into an issue with broken output with pseudo console enabled, i.e. `MSYS=enable_pcon`, and I traced it back to this change.

Would you be able to provide *some* sort of detail?

> Shouldn't the setting of the console codepage at the end of `setup_local` be conditional on pseudo console being disabled as suggested by @tyan0 in https://github.com/msys2/MSYS2-packages/issues/1974#issuecomment-686837959?

We specifically did *not* do that, in order to fix the reported issue both with enabled *and* with disabled support for Pseudo Consoles.

nga888:
> Would you be able to provide _some_ sort of detail?

Sorry, I do not have a simple reproducible, all I've managed to determine so far is that without this change my problem goes away. It's a work related testing system that I run from bash in mintty that runs native python, NAnt and various native applications that also involves redirection of stdin/stdout.

> We specifically did _not_ do that, in order to fix the reported issue both with enabled _and_ with disabled support for Pseudo Consoles.

I did see mention of this in that thread, but the reproducible for that issue works fine for me with pseudo console enabled. I think @tyan0 also said it worked for him too, IIRC.

Sorry that I can't provide any more useful info. If I get time to investigate further, I'll follow up. For now, I'm fine with my patched version of the runtime and I'm probably in the minority that run with pseudo console enabled.

dscho:
> Sorry, I do not have a simple reproducible, all I've managed to determine so far is that without this change my problem goes away.

Unfortunately, this is saying something like "you have a bug, but I won't tell you anything about it, so you cannot fix it either"

Could you _at least_ talk about the symptoms you are seeing?

As far as anybody on this side of the report can tell, the bug might be in your testing system. Or it might be in the MSYS2 runtime. Or in the Cygwin portion of the code. Nobody except for you has enough information to tell...

nga888:
> Unfortunately, this is saying something like "you have a bug, but I won't tell you anything about it, so you cannot fix it either"
> 
> Could you _at least_ talk about the symptoms you are seeing?

I would normally do more investigation before commenting, but in this case I saw that the author of the original code had already suggested the change that would address my issue, which is why I asked the question. But as it was a conscience decision to go with the current patch, then I have no complaints.

As far as symptoms are concerned, it appears that output/input is getting "corrupted" somewhere between the native applications, but I haven't had time to pinpoint where and how.

> As far as anybody on this side of the report can tell, the bug might be in your testing system. Or it might be in the MSYS2 runtime. Or in the Cygwin portion of the code. Nobody except for you has enough information to tell...

Yes, without further information and investigation, it is hard to know what might be at fault. However, this is a "legacy" testing system that has not changed for a very long time and was running fine until this patch. Unfortunately, as I mentioned before, I don't really have the time now to investigate this in more detail. If I do get time to look into this in more detail, I'll follow up if appropriate.

Sorry for any unnecessary noise caused.

dscho:
> > We specifically did _not_ do that, in order to fix the reported issue both with enabled _and_ with disabled support for Pseudo Consoles.
> 
> I did see mention of this in that thread, but the reproducible for that issue works fine for me with pseudo console enabled. I think @tyan0 also said it worked for him too, IIRC.

I could _not_ replicate the findings. In my hands, the patch was required to fix both `disable_pcon`/`enable_pcon`. That is why I contributed the patch in the current form, and unless I get some tangible evidence where the patch does anything unintended, I'll remain unconvinced.

nga888:
FYI I managed to find some time to look into this and my problem arises due to the way NAnt feeds input to stdin when executing a program. When the code page is set to 65001, it prefixes a BOM to the input stream and this causes a problem if the receiving program does not expect this. When the code page is 850 (my default) it works fine.

So my problem does arise as a side-effect of this patch. However, what's the "real" problem is not so clear, as many of the older native programs do not support UTF-8 and it's also quite unusual to have a BOM for UTF-8.

PR #2161, opened 2020-09-25
makepkg: Add --default-image-base-high to LDFLAGS

lazka:
This will allow us to change the default to low in binutils, so that our
built packages get proper ASLR support, while users get the same behaviour as
with an older binutils with the old defaults.

See https://github.com/msys2/MINGW-packages/issues/7023#issuecomment-698246859

### DIFF
--- a/pacman/makepkg_mingw64.conf
+++ b/pacman/makepkg_mingw64.conf
@@ -59,7 +59,7 @@ DXSDK_DIR=${MINGW_PREFIX}/${MINGW_CHOST}
 CPPFLAGS="-D__USE_MINGW_ANSI_STDIO=1"
 CFLAGS="-march=x86-64 -mtune=generic -O2 -pipe"
 CXXFLAGS="-march=x86-64 -mtune=generic -O2 -pipe"
-LDFLAGS="-pipe -Wl,--dynamicbase,--high-entropy-va,--nxcompat"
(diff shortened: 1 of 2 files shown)


lazka:
@mati865 @jeremyd2019

mati865 (approved):

pacman/makepkg_mingw64.conf
```diff
@@ -59,7 +59,7 @@ DXSDK_DIR=${MINGW_PREFIX}/${MINGW_CHOST}
 CPPFLAGS="-D__USE_MINGW_ANSI_STDIO=1"
 CFLAGS="-march=x86-64 -mtune=generic -O2 -pipe"
 CXXFLAGS="-march=x86-64 -mtune=generic -O2 -pipe"
-LDFLAGS="-pipe -Wl,--dynamicbase,--high-entropy-va,--nxcompat"
+LDFLAGS="-pipe -Wl,--dynamicbase,--high-entropy-va,--nxcompa,--default-image-base-high"
```
jeremyd2019, line 62: Looks like you lost a 't' there on --nxcompat

jeremyd2019 (changes requested):

pacman/makepkg_mingw64.conf
lazka: oops, thanks!

Issue #1353, opened 2018-07-21
grep 3.1 regression


Furniel:
After msys2 update i686 mingw gcc build tests began to fall with error:
```
terminate called after throwing an instance of 'int'
terminate called recursively

This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.
```
I've managed to track down that grep update from 3.0 to 3.1 causes this problem.

On the other hand, after comparison of build logs, i found out than for some gcc libs libtool generates wrong configs:
grep 3.0:
```
# Dependencies to place before and after the objects being linked to
# create a shared library.
predep_objects="C:/msys64/mingw32/i686-w64-mingw32/lib/dllcrt2.o C:/temp/i686-8b-posix-dwarf-rt_v5/build/gcc-8-branch/gcc/crtbegin.o"
postdep_objects="C:/temp/i686-8b-posix-dwarf-rt_v5/build/gcc-8-branch/gcc/crtend.o"
predeps=""
postdeps="-lmingw32 -lgcc_s -lgcc -lmoldname -lmingwex -lmsvcrt -lpthread -ladvapi32 -lshell32 -luser32 -lkernel32 -liconv -lmingw32 -lgcc_s -lgcc -lmoldname -lmingwex -lmsvcrt"

```
grep 3.1
```
# Dependencies to place before and after the objects being linked to
# create a shared library.
predep_objects="C:/msys64/mingw32/i686-w64-mingw32/lib/dllcrt2.o C:/temp/i686-8b-posix-dwarf-rt_v5/build/gcc-8-branch/gcc/crtbegin.o"
postdep_objects=""
predeps=""
postdeps="-lmingw32 -lgcc_s -lgcc -lmoldname -lmingwex -lmsvcrt -lpthread -ladvapi32 -lshell32 -luser32 -lkernel32 -liconv -lmingw32 -lgcc_s -lgcc -lmoldname -lmingwex -lmsvcrt"

```
Looks like real cause of the problem is one of this changes in grep 3.1:
```
The following changes affect only MS-Windows platforms.  First, the
  --binary (-U) option now governs whether binary I/O is used, instead
  of a heuristic that was sometimes incorrect.  Second, the
  --unix-byte-offsets (-u) option now has no effect on MS-Windows too.
```

Alexpux:
This is upstream grep changes:
[...]

PR #2163, opened 2020-09-28
grep: Update to 3.5

jannick0:
* grep/PKGBUILD:
  - update version and hash
  - remove references to outdated and obsolete patches
  - build binary in separate directory
* grep/{grep-2.23-msys2.patch,grep-3.0-2.patch}: remove files


Biswa96:
Are those patches in upstream repository? If yes then a commit message with that upstream commit link would be helpful.

Alexpux:
It not possible to update grep for MSYS2 as upstream drop support for crlf. So I hope it will not be merged

Alexpux:
here is commit:
https://git.savannah.gnu.org/cgit/grep.git/commit/?id=5ac3d87a69122d45a837840914c2f9e66b5bb8d4

jannick0:
@Alexpux  I fully agree that grep is a central piece to MSYS2. Yesterday I was not aware of this severe issue. 

* The documentation (changed in the commit you are pointing to - thanks) does not read very clear to me (highlilghts from me):

Under MS-Windows, use binary I/O when reading and writing files **other
than the user's terminal**, so that all input bytes are read and written
as-is.  This overrides the default behavior where @command{grep}
follows the operating system's advice whether to use binary or text
I/O.

* grep-3.0-2.src.patch cannot be reverted in the source of grep-3.5, since the target source file is removed meanwhile.  So the much feared change does not show up in the source anymore.   
* I am confused about grep's expected vs actual CRLF/LF handling now: What should grep do and what shouldn't it do (test cases?) with respect to  `-U`, CRLF/LF in shell MSYS2/MINGW64/MINGW32,..., which input to grep?

Or am I missing something here?

Alexpux:
@jannick0 look at https://github.com/msys2/MSYS2-packages/issues/1353

PR #2165, opened 2020-09-30
Add lcov package

lazka:
Useful for creating coverage reports of mingw builds


marknsikora:
Glad to see a 1.15 version was finally released and removed the need for the intermediate format patches

👍

PR #2170, opened 2020-10-02
msys2-keyring: Update to 20201002

elieux:
I extended my key's validity to 2021-10-02.

### DIFF
--- a/msys2-keyring/PKGBUILD
+++ b/msys2-keyring/PKGBUILD
@@ -2,23 +2,18 @@
 
 PKGEXT='.pkg.tar.xz'
 pkgname=msys2-keyring
-pkgver=r21.b39fb11
+epoch=1
+pkgver=20201002
 pkgrel=1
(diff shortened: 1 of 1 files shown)


msys2-keyring/PKGBUILD
```diff
@@ -2,23 +2,18 @@
 
 PKGEXT='.pkg.tar.xz'
 pkgname=msys2-keyring
-pkgver=r21.b39fb11
+epoch=1
+pkgver=20201002
```
lazka, line 6: `vercmp.exe 10201002 r21.b39fb11` gives me 1, so an epoch is strictly not needed afaics.

lazka:
lgtm

jeremyd2019:
When this package is uploaded, would you consider making a special dispensation for msys2-keyring package to continue to be published to i686 also, so that mingw repos/packages can continue to be validated?  It is an 'any' arch package, so shouldn't require any additional building.

PR #2177, opened 2020-10-07
pacman-mirrors: Remove repo.msys2.org

elieux:
We moved the primary to a smaller hosting and there's probably enough
mirrors now.

### DIFF
--- a/pacman-mirrors/PKGBUILD
+++ b/pacman-mirrors/PKGBUILD
@@ -2,18 +2,18 @@
 
 PKGEXT='.pkg.tar.xz'
 pkgname=pacman-mirrors
-pkgver=20201003
+pkgver=20201007
 pkgrel=1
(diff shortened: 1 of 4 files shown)


lazka:
lgtm

lygstate:
This is not a good idea, I have script that using sed to replace repo.msys2.org with own mirror
After this changed, didn't know how to replace it properly

lazka:
You can just add your server at the top

Biswa96:
Goodbye, repo.msys2.org 😭 We've been together many days!

lygstate:
> You can just add your server at the top

I means lack a template to replace in a single short, maybe it's better leave a template to sed and replace?

mati865:
@lygstate you can just override the mirror list with your single url.

lygstate:
@mati865 Yes, three command.

PR #2181, opened 2020-10-09
filesystem: use matching icon for console

kmilos:
Helps identifying window thumbnails when switching between MSYS and MINGW64 consoles


lazka:
you have to update pkgver and set pkgrel to 1 in the PKGBUILD file. and update the checksums using `updpkgsums`

lazka:
Thanks!

PR #2192, opened 2020-10-16
pkgconf/pkg-config: some cleanups/updates, make it possible to install pkgconf as an alternative

lazka:
Maybe we can default to pkgconf in the future, like Arch.


lazka:
I've built pacman and glib (both autotools and meson) with it and seems to work fine.

PR #2195, opened 2020-10-17
add ragel package

ksherlock:
There is already a mingw64 ragel package already but it would be better to have it in msys2 so it's available for mingw32, 64, and msys2.

ragel generates code and the generated code is identical everywhere so one binary can service them all.


Biswa96:
> it would be better to have it in msys2 

Is it required by any other package? Or what is your use case in  which the mingw one can't satisfy the need?

ksherlock:
It's not needed by existing package as far as I know but it's a development utility.  Use case is being able to run it without worrying about if i'm in an msys2 shell, mingw32 shell, or mingw64 shell (where a package is available).

lazka:
sorry, but we try to limit the amount of cygwin packages. if the mingw shell works, use that.

PR #2197, opened 2020-10-18
Switch from pkg-config to pkgconf

lazka:
pkgconf is maintained, doesn't include glib and supports more pkg-config
features and builds faster.

Arch and Fedora have switched long ago, so we should too.

### DIFF
--- a/pkg-config/PKGBUILD
+++ b/pkg-config/PKGBUILD
@@ -2,12 +2,11 @@
 
 pkgname=pkg-config
 pkgver=0.29.2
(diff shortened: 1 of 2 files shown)


lazka:
@elieux any thoughts on this?

lazka:
I guess no one has any strong opinions on this, so I'm going to merge this tomorrow.

lygstate:
So pkg-config is going to be removed in future?

lazka:
there are no plans to remove it right now at least.

PR #2200, opened 2020-10-22
msys2-runtime: expose the "deep-copy" symlink behavior via a new mode

dscho:
This patch synchronizes msys2/MSYS2-packages/msys2-runtime with
msys2/msys2-runtime after merging
https://github.com/msys2/msys2-runtime/pull/16.

Unfortunately, most of the patch is somewhat useless churn: the signature of the patches records the Git version used to generate the patches, and that version changed from v2.27.0 to v2.28.0.


dscho:
Oh, I should also mention that all I did was running `./update-patches.sh`, then incrementing the `pkgrel` and then transmogrifying the PR description into the commit message.

PR #2204, opened 2020-10-24
fixes for packages that failed to build in msys2-autobuild

jeremyd2019:
Assorted fixes for packages that failed to build in https://github.com/msys2/msys2-autobuild/runs/1300765313?check_suite_focus=true

`less` is a pgp nightmare. It was missing validpgpkeys, and the .sig file was named without the .tar.gz extension, and even with both of those fixed it still failed to validate, I assume it failed to retrieve the key.  I commented out the sig file so it could build.

### DIFF
--- a/less/PKGBUILD
+++ b/less/PKGBUILD
@@ -2,16 +2,19 @@
 
 pkgname=less
 pkgver=563
-pkgrel=1
+pkgrel=2
 pkgdesc="A terminal based program for viewing text files"
 license=('GPL3')
 arch=('i686' 'x86_64')
 url="http://www.greenwoodsoftware.com/less"
 depends=('ncurses' 'libpcre')
 makedepends=('ncurses-devel' 'pcre-devel')
-source=(http://www.greenwoodsoftware.com/${pkgname}/${pkgname}-${pkgver}{.tar.gz,.sig})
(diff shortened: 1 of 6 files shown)


less/PKGBUILD
```diff
@@ -2,16 +2,19 @@
...
 license=('GPL3')
 arch=('i686' 'x86_64')
 url="http://www.greenwoodsoftware.com/less"
 depends=('ncurses' 'libpcre')
 makedepends=('ncurses-devel' 'pcre-devel')
-source=(http://www.greenwoodsoftware.com/${pkgname}/${pkgname}-${pkgver}{.tar.gz,.sig})
-sha256sums=('ce5b6d2b9fc4442d7a07c93ab128d2dff2ce09a1d4f2d055b95cf28dd0dc9a9a'
-            'SKIP')
+source=("http://www.greenwoodsoftware.com/${pkgname}/${pkgname}-${pkgver}.tar.gz")
+# pgp check still fails with FAILED (unknown public key F153A7C833235259)
```
lazka, line 13: The error with a short key means the signature was created with an ancient gpg that doesn't include the full key ID in the signature so gpg can't auto fetch the key (because it can't guarantee it is unique I think).

lazka:
Thanks!

PR #2205, opened 2020-10-25
gnutls-devel: add depends on other -devel packages

jeremyd2019:
`pkg-config --exists --print-errors gnutls` fails if those .pc files are missing.

Rebuild lftp and emacs after this update, as they were failing due to this.  Also added libunistring-devel as a makedepends on emacs to fix its build.


lazka:
Thanks!

PR #2211, opened 2020-10-27
Add support for `tmate`

dscho:
https://tmate.io/ allows "instant terminal sharing". This PR brings the client to MSYS2.


dscho:
The build will probably fail because `tmate` can only be built once the `msgpack-c`/`libssh` packages are available.

PR #2215, opened 2020-10-28
mingw-w64-cross-*: hardcode the commit hash in the version

lazka:
Otherwise "git am" used in prepare() wil change HEAD and as a result pkgver().
Also when "git am" is used the hash in the version isn't in the upstream repo
which makes it hard to figure out where it was built from.

Just hardcode to the commit that was checkedd out initially.


Biswa96:
Is it possible to put the `_commit` variable somewhere so that it can be accessed by allllll the mingw-w64 related (normal, cross, msys2) pkgbuild files?

PR #2217, opened 2020-10-29
libevent: fix DNS lookups in MSYS2

dscho:
When I tried to verify that the `tmate` package at https://github.com/msys2/msys2-autobuild/releases/tag/staging-msys works correctly, it sadly failed to start, with a terse "lost server" message.

The reason why I did not see this breakage locally was that I still ran with libevent v2.1.11, which for some reason does not seem to have any problem looking up the name server IPs, even if the code in [`evdns.c`](https://github.com/libevent/libevent/blob/HEAD/evdns.c) does not seem to have changed all that dramatically.

Be that as it may, this here PR fixes `libevent` so that the name server IPs are obtained "the Win32 way". And with this, I can finally start `tmate` in an Azure Pipeline!

### DIFF
--- a/libevent/0001-Fix-DNS-lookups-in-MSYS2.patch
+++ b/libevent/0001-Fix-DNS-lookups-in-MSYS2.patch
@@ -1,0 +1,148 @@
+From 4cdf3fd2289d9658ecce21abf613cac389c4e732 Mon Sep 17 00:00:00 2001
+From: Johannes Schindelin <johannes.schindelin@gmx.de>
+Date: Thu, 29 Oct 2020 22:18:33 +0100
+Subject: [PATCH] Fix DNS lookups in MSYS2
+
+We need to use the Win32 way to obtain the name server IPs.
+
+Signed-off-by: Johannes Schindelin <johannes.schindelin@gmx.de>
+---
+ evdns.c         | 21 ++++++++++++++-------
+ evutil.c        | 23 ++++++++++++++---------
+ util-internal.h |  4 ++--
+ 3 files changed, 30 insertions(+), 18 deletions(-)
+
+diff --git a/evdns.c b/evdns.c
+index a5b31a3c..cc7cad84 100644
+--- a/evdns.c
++++ b/evdns.c
+@@ -85,6 +85,11 @@
+ #include <shlobj.h>
+ #endif
+ 
++#ifdef __MSYS__
++#include <w32api/shlobj.h>
++#include <w32api/iptypes.h>
++#endif
++
+ #include "event2/dns.h"
+ #include "event2/dns_struct.h"
+ #include "event2/dns_compat.h"
+@@ -3637,7 +3642,7 @@ evdns_base_resolv_conf_parse(struct evdns_base *base, int flags, const char *con
+ static char *
+ evdns_get_default_hosts_filename(void)
+ {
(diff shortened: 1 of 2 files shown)


dscho:
@lazka sorry for not testing `tmate` better before opening that PR... :-(

Alexpux:
@dscho no worry, everyone here work in the same way. There are no possibility to test everything

lazka:
There were some autotools changes compared to the old version. Do you think it's worth comparing config.h to figure out why it worked before? Or is your patch an improvement never the less?

dscho:
> There were some autotools changes compared to the old version.

I did not see anything in particular that stood out that could possibly explain the regression.

> Do you think it's worth comparing config.h to figure out why it worked before?

Typically I would say: yes. However, in this case, I think the breakage just intends to punish me for coming up with https://github.com/msys2/MSYS2-packages/pull/2211/commits/09b14d43629a44249eb5b2ada8f987db47e5b88b#diff-e8bef9f719d924bc36ef9a4c1ea2fe092e6c68bc948aa4edb3b28f34482a3567R32-R37 instead of fixing the DNS lookup correctly.

I could imagine that we may be able to drop https://github.com/dscho/MSYS2-packages/blob/09b14d43629a44249eb5b2ada8f987db47e5b88b/tmate/0005-Try-to-side-step-libevent-for-DNS-resolution.patch altogether once this PR is in.

> Or is your patch an improvement never the less?

I would contend that it is a strict improvement: without this PR, `libevent` will fail to perform any DNS lookups. With this PR, it will perform them as intended.

There is probably _some_ lesson to be learned from investigating why `tmate` worked with libevent v2.1.11 (and with my work-around), even if the bug was already there where no name server IPs could be determined at all, but stopped working with libevent v2.1.12. My gut feeling is that the answer will be something like "oh, there was _another_ bug that hid that bug" or at least a very similar answer.

In short: this PR is the right thing to do in any case, I believe.

dscho:
> no worry, everyone here work in the same way. There are no possibility to test everything

@Alexpux thank you. It helps me feeling less embarrassed.

lazka:
> > There were some autotools changes compared to the old version.
> 
> I did not see anything in particular that stood out that could possibly explain the regression.

I saw this: https://github.com/libevent/libevent/commit/6d54be2cc078351b4ce1e469bedc4a976f4e3636 but no idea if this is relevant.

> In short: this PR is the right thing to do in any case, I believe.

ok, thanks :)

dscho:
> > > There were some autotools changes compared to the old version.
> > 
> > 
> > I did not see anything in particular that stood out that could possibly explain the regression.
> 
> I saw this: [libevent/libevent@6d54be2](https://github.com/libevent/libevent/commit/6d54be2cc078351b4ce1e469bedc4a976f4e3636) but no idea if this is relevant.

This looks very close to the issue at hand, but it is safely in the MINGW realm, not in the MSYS one. In particular, `_WIN32` is _not_ defined in MSYS builds, while `__MSYS__` is.

And yes, this patch is about `getaddrinfo()`, but we need to determine the IPs of the name servers, and the classical Unix method (parse `/etc/resolv.conf`) simply won't work in MSYS2.

> > In short: this PR is the right thing to do in any case, I believe.
> 
> ok, thanks :)

Thank you for merging it!

PR #2230, opened 2020-11-04
Add packages for bootstrapping Clang subsystem

mati865:
Surely some of the patches are unnecessary for MSYS2 package but I included them from MINGW package so the are more "in-line".

Clang's msys target doesn't support using GCC's libstdc++ but it's not necessary for bootstrapping.
Headers and CRT use `_variant=gcc` so they have to built locally and uploaded (their pkgname differs if `_variant=clang` is chosen).

### DIFF
--- a/mingw-w64-cross-compiler-rt/PKGBUILD
+++ b/mingw-w64-cross-compiler-rt/PKGBUILD
@@ -1,0 +1,68 @@
+# Maintainer: Mateusz Mikuła <mati865@gmail.com>
+
+_realname=compiler-rt
+_mingw_suff=mingw-w64-cross
+pkgname=("${_mingw_suff}-${_realname}")
+pkgver=11.0.0
+pkgrel=1
+pkgdesc="Cross Clang for the MinGW-w64"
+arch=('i686' 'x86_64')
+url="https://llvm.org"
+license=('custom:Apache 2.0 with LLVM Exception')
+depends=('clang')
+makedepends=("cmake" "ninja" "${_mingw_suff}-clang-headers")
+options=('!strip' 'staticlibs' '!emptydirs' '!buildflags')
+_url=https://github.com/llvm/llvm-project/releases/download/llvmorg-${pkgver}
+source=(${_url}/compiler-rt-${pkgver}.src.tar.xz{,.sig})
+sha256sums=('374aff82ff573a449f9aabbd330a5d0a441181c535a3599996127378112db234'
+            'SKIP')
+validpgpkeys=('B6C8F98282B944E3B0D5C2530FC3042E345AD05D'  # Hans Wennborg, Google.
+              '474E22316ABF4785A88C6E8EA2C794A986419D8A') # Tom Stellard
+
+_target="x86_64-w64-mingw32"
+
+prepare() {
+  cd ${srcdir}
+}
+
+build() {
+  cd "${srcdir}"
+
+  [[ -d build-${CARCH} ]] && rm -rf build-${CARCH}
+  mkdir build-${CARCH} && cd build-${CARCH}
+
+  export CC="clang" CXX="clang++" AS="clang" AR="llvm-ar" RANLIB="llvm-ranlib" DLLTOOL="llvm-dlltool" LD="clang"
+
+  CFLAGS+=" -target ${_target} --sysroot=/opt/${_target}"
+  CPPFLAGS+=" -target ${_target} --sysroot=/opt/${_target}"
+  CXXFLAGS+=" -target ${_target} --sysroot=/opt/${_target}"
+  LDFLAGS+=" -target ${_target} --sysroot=/opt/${_target} -fuse-ld=lld"
+
+  cmake \
+    -G'Ninja' \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_C_COMPILER_TARGET=${CARCH}-windows-gnu \
+    -DCMAKE_C_COMPILER_WORKS=ON \
+    -DCMAKE_CXX_COMPILER_WORKS=ON \
+    -DCMAKE_INSTALL_PREFIX=/usr \
+    -DCMAKE_SIZEOF_VOID_P=8 \
+    -DCMAKE_SYSTEM_IGNORE_PATH=/usr/lib \
+    -DCMAKE_SYSTEM_NAME=Windows \
+    -DCOMPILER_RT_BUILD_LIBFUZZER=OFF \
+    -DCOMPILER_RT_BUILD_PROFILE=OFF \
+    -DCOMPILER_RT_BUILD_SANITIZERS=OFF \
+    -DCOMPILER_RT_BUILD_XRAY=OFF \
+    -DCOMPILER_RT_DEFAULT_TARGET_ONLY=ON \
+    -DCOMPILER_RT_USE_BUILTINS_LIBRARY=ON \
+    ../${_realname}-${pkgver}.src
+
+  ninja
+}
+
+package() {
+  cd $srcdir/build-${CARCH}
+
+  # ninja install needs working compiler detection to find correct directory
+  mkdir -p "${pkgdir}/usr/lib/clang/${pkgver}/lib/windows"
+  cp "lib/windows/libclang_rt.builtins-${CARCH}.a" "${pkgdir}/usr/lib/clang/${pkgver}/lib/windows"
+}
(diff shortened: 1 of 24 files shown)


lazka:
(rebased on top of master so we get more disk space in CI)

mati865:
`mingw-w64-cross-headers-git` and `mingw-w64-cross-crt-git` have to built manually (with `_variant=clang`).

mingw-w64-cross-compiler-rt/PKGBUILD
```diff
@@ -0,0 +1,68 @@
...
+_mingw_suff=mingw-w64-cross
+pkgname=("${_mingw_suff}-${_realname}")
+pkgver=11.0.0
+pkgrel=1
+pkgdesc="Cross Clang for the MinGW-w64"
+arch=('i686' 'x86_64')
+url="https://llvm.org"
+license=('custom:Apache 2.0 with LLVM Exception')
+depends=('clang')
+makedepends=('ninja')
```
jeremyd2019, line 13:
```suggestion
makedepends=('cmake' 'ninja')
```

mati865: I have it committed locally, just waiting on `-DLLVM_LINK_LLVM_DYLIB=ON` build to finish to see if LLD needs the recent patch. That will reduce the size.

jeremyd2019: I would be interested to know that too.  I was thinking it would be possible for something 'invasive' like cygwin/msys to implement something different for global destructors in DLLs, but I don't know if they needed to in order to get sufficiently posix-like behavior for their purposes.

mati865:
Had to fix patches a bit for dynamic linking to even succeed.

Static vs dynamic size:
```
Total Installed Size:    667.66 MiB
Net Upgrade Size:      -2369.18 MiB
```

The saving is neat but unfortunately Clang crashes when building libc++. There must be something else missing in the patches but it's a needle in a haystack.

Clang has extracted 13 MiB C++ source file that reproduces the crash so I can reliably reproduce the crash. However I have notices something weird, it doesn't crash when calling `/usr/bin/clang++` through `/usr/bin/gdb`.
The best I could get is WinDbg dump: 
[clang.zip](https://github.com/msys2/MSYS2-packages/files/5498087/clang.zip)

It might be not worth the trouble though so I kept static version for now.

mati865:
According to the stacktrace it seems to crash somewhere here: https://github.com/llvm/llvm-project/blob/176249bd6732a8044d457092ed932768724a6f06/llvm/lib/CodeGen/AsmPrinter/DIE.cpp#L214-L219 called from https://github.com/llvm/llvm-project/blob/176249bd6732a8044d457092ed932768724a6f06/llvm/lib/CodeGen/AsmPrinter/DwarfDebug.cpp#L1238

Getting debug build will be hard because of the MSYS overhead (this "minimal" release build already takes much longer than all LLVM related tools when building with MinGW).

jeremyd2019:
> According to the stacktrace it seems to crash somewhere here: https://github.com/llvm/llvm-project/blob/176249bd6732a8044d457092ed932768724a6f06/llvm/lib/CodeGen/AsmPrinter/DIE.cpp#L214-L219

Actually disassembly looks suspiciously like https://github.com/llvm/llvm-project/blob/176249bd6732a8044d457092ed932768724a6f06/llvm/lib/CodeGen/AsmPrinter/DIE.cpp#L203-L212, except for some reason the initial NULL check of the `while (p)` loop was elided, so there was a NULL-pointer dereference.  Maybe there's something in the language spec that allows the compiler to assume `this` cannot be NULL.  I see the caller doesn't check for a NULL `Die` pointer other than by an `assert`

mati865:
> Actually disassembly looks suspiciously like llvm/llvm-project@176249b/llvm/lib/CodeGen/AsmPrinter/DIE.cpp#L203-L212

Yeah, you are right. I looked at this code locally but posted wrong snippet.
AFAIK `this` cannot be null unless there is UB somewhere.

jeremyd2019:
> AFAIK `this` cannot be null unless there is UB somewhere.

Yup, I guess the proper question would be how an `Entity` with a NULL `DIE *` got in there in the first place - it seems this code assumes that can't happen...  But like you said, probably not worth spending time investigating, as this package is intended only to bootstrap a clang subsystem not many will be bothered by the size

mati865:
WOW, this explains a lot:
```
mateusz@DESKTOP-mati MSYS /d/tmp
$ clang -target x86_64-w64-mingw32 -v hello.c
clang version 11.0.0 (https://github.com/mati865/msys2-packages 65a38d271ebae7d8122723b9ce8945c678aa1a1d)
Target: x86_64-w64-windows-gnu
Thread model: posix
InstalledDir: /usr/bin
 "/usr/bin/clang" -cc1 -triple x86_64-w64-windows-gnu -emit-obj -mrelax-all -disable-free -disable-llvm-verifier -discard-value-names -main-file-name hello.c -mrelocation-model pic -pic-level 2 -mframe-pointer=none -fmath-errno -fno-rounding-math -mconstructor-aliases -mms-bitfields -munwind-tables -target-cpu x86-64 -debugger-tuning=gdb -v -resource-dir /usr/lib/clang/11.0.0 -internal-isystem /usr/lib/clang/11.0.0/include -internal-isystem /mingw64/x86_64-w64-mingw32/sys-root/mingw/include -internal-isystem /mingw64/x86_64-w64-mingw32/include -internal-isystem /mingw64/include -fdebug-compilation-dir /d/tmp -ferror-limit 19 -fno-use-cxa-atexit -fgnuc-version=4.2.1 -fseh-exceptions -fcolor-diagnostics -faddrsig -o /tmp/hello-cd9a81.o -x c hello.c
clang -cc1 version 11.0.0 based upon LLVM 11.0.0 default target x86_64-pc-msys
ignoring nonexistent directory "/mingw64/x86_64-w64-mingw32/sys-root/mingw/include"
#include "..." search starts here:
#include <...> search starts here:
 /usr/lib/clang/11.0.0/include
 /mingw64/x86_64-w64-mingw32/include
 /mingw64/include
End of search list.
 "/usr/bin/ld" -m i386pep -Bdynamic -o a.exe /mingw64/x86_64-w64-mingw32/lib/crt2.o /mingw64/lib/gcc/x86_64-w64-mingw32/10.2.0/crtbegin.o -L/mingw64/lib/gcc/x86_64-w64-mingw32/10.2.0 -L/mingw64/x86_64-w64-mingw32/lib -L/mingw64/lib -L/mingw64/x86_64-w64-mingw32/sys-root/mingw/lib /tmp/hello-cd9a81.o -lmingw32 -lgcc -lgcc_eh -lmoldname -lmingwex -lmsvcrt -ladvapi32 -lshell32 -luser32 -lkernel32 -lmingw32 -lgcc -lgcc_eh -lmoldname -lmingwex -lmsvcrt -lkernel32 /mingw64/lib/gcc/x86_64-w64-mingw32/10.2.0/crtend.o
```

mati865:
Finally fixed the build.

@lazka how do we want to build it?
https://github.com/msys2/MSYS2-packages/pull/2230/commits/a05b7627b3350ee028cf9ba2f408bca9b1a41a56 switches the variant so if we want to build it entirely on CI we will have to revert it afterward. Alternatively we can duplicate Clang version of `mingw-w64-cross-headers-git` and `mingw-w64-cross-crt-git` to separate directories to avoid shuffling `_variant`.

mingw-w64-cross-compiler-rt/PKGBUILD
mati865: FWIW I've used build with `-DLLVM_LINK_LLVM_DYLIB=ON` for some tests and LLD haven't deadlocked even once.

lazka:
Not sure if we can do bootstrapping in CI anyway.. so having that _variant thing seems fine (although defaulting to gcc).

lazka:
hm, on a second thought duplicating with different package names sounds like a good idea too.. if that works for you.

mati865:
I've split out changes for non Clang mingw-w64 packages, hopefully CI will succeed now.

jeremyd2019:
It appears that mingw-w64-cross-clang package failed to build in msys2-autobuild: https://github.com/msys2/msys2-autobuild/runs/1540405115?check_suite_focus=true#step:14:9264

```
  CMake Error at /c/_/S/mingw-w64-cross-clang/src/libunwind/src/CMakeLists.txt:92 (message):
    Compiler doesn't support generation of unwind tables if exception support
    is disabled.  Building libunwind DSO with runtime dependency on C++ ABI
    library is not supported.
```

I am more than a little suspicious that there was a missing (make)depends that resulted in the compiler just being broken, because it seemed like pretty much every test CMake did failed.

mati865:
>   -- Detecting C compiler ABI info - failed

Yeah the compiler is broken but it worked in this PR.

Can I somehow download packages that were build by autobuild but are not yet in the repo?

jeremyd2019:
You can grab them manually from https://github.com/msys2/msys2-autobuild/releases/tag/staging-msys (and probably rename them), or use `autobuild.py fetch-assets` from that repo to grab all staged assets properly named.

jeremyd2019:
It looks like the packages that built are now in the repo

jeremyd2019:
Makedepends tweak in #2272

mati865:
Thank you for looking into this, I was busy last week.

Issue #1697, opened 2019-07-16
openssh: sftp prompt broken


cw789:
I open this issue to have it documented here.
Initial reported at: https://github.com/git-for-windows/git/issues/2257

Command:
```
sftp hostname
```
Expected:
```
sftp>_
```
Got:
```
s     _
```

MNeuling:
The behaviour of the prompt is very disturbing. If you use auto completion or backspace the whole typed command is invisible. The bug makes the sftp command nearly unusable.

Is there a plan to fix it?

edwardbeckett:
I'm having the same issue. After logging in via sftp the prompt is broken and path navigation is very difficult. 

![sftp](https://user-images.githubusercontent.com/949373/67001278-f6586d00-f09e-11e9-80c7-313eaffba35c.png)

JulienCochuyt:
Same issue here using sftp invoked from MinTTY as shipped with Git for Windows
Any hint on how we could help diagnose/fix this?
Older versions did not exhibit this faulty behavior, but I cannot recall what was the last working one.

lhmouse:
ping?

Tantalus13A98B5F:
It's an issue caused by `libedit`, who mixed the use of `wint_t` and `wchar_t`. On real Linux (here Ubuntu 20.04 amd64) these two types are identical, but on MSYS2 `wchar_t` is only half of `wint_t`. Thus when you are passing `wint_t[]` to an argument of type `wchar_t*`, the program will take as if the wide-string has only one character, making it an MSYS2-specific bug.

A true fix of this problem should wait until `libedit` have their code refactored and come to a portable type definition. For a quickfix (caveat: don't know if this would be causing new problems), try the following patch with `libedit/src/el.h`, then recompile `libedit` and `openssh`:

```diff
120,121c120,121
<       wint_t          **el_display;   /* Real screen image = what is there */
<       wint_t          **el_vdisplay;  /* Virtual screen image = what we see */
---
>       wchar_t         **el_display;   /* Real screen image = what is there */
>       wchar_t         **el_vdisplay;  /* Virtual screen image = what we see */
```

cw789:
Close now as https://github.com/msys2/MSYS2-packages/pull/2241 is merged.
Thanks a lot @Tantalus13A98B5F .

PR #2241, opened 2020-11-24
fixing broken prompt of openssh

Tantalus13A98B5F:
Please refer to #1697.

Tried `-fno-short-wchar` but it didn't take effect with gcc 9.3.0 shipped with MSYS2, i.e. `sizeof(wchar_t)` is still 2 after that. Still, enabling the compiler to use 4-byte `wchar_t` would be a far better choice than this patch, but so far I haven't found a way.

Generally narrowing the buffer type to `wchar`s should be of no harm, for `wint`s are largely used to contain `wEOF`s, which is not a requirement for `el_v?display`. In practice this patch doesn't incur warnings of type narrowing, and the overall function works out fine, even under my CJK environment.


lazka:
increase pkgrel

lazka:
Looks good I guess. wchar_t is 2 bytes in cygwin and I don't think it can diverge from the Windows API sizes really.

"upstream" is at least aware that they don't support this configuration as can be seen by this patch: https://github.com/msys2/MSYS2-packages/blob/0705b15ae118c8b02a8309dcd62ee76f6e35a7fc/libedit/libedit-20130712-1.src.patch#L7

lazka:
hm, looking at the build logs, terminal_alloc_display/terminal_free_display might need adjustments too, though they likely will now just allocate too much.. so not a big issue.

PR #2243, opened 2020-11-29
Remove Gaphor DigitalOcean mirror

danyeaw:
Now that we have expanded the number of mirrors, I am shutting down my mirror on DigitalOcean because I don't think it is needed any longer.


lazka:
Thanks for your work!

PR #2245, opened 2020-12-01
sqlite: Update to 3.34.0

jannick0:
* PKGBUILD:
  - update version variables
  - replace 0010-lemon-share.patch by 0010-sqlite-3.34.0-lemon.patch
* 0010-sqlite-3.34.0-lemon.patch:
  - renamed from 0010-lemon-share.patch
  - adjust second patch hunk to sqlite 3.34.0


lazka:
thanks!

PR #2247, opened 2020-12-02
Changed Sourceforge mirror to be faster

LumitoLuma:
Changed sourceforge mirror URL so it has less redirections:
-   Old Sourceforge mirror: sourceforge.net
-   New Sourceforge mirror: downloads.sourceforge.net

Tested and it works fine for me.


Biswa96:
* Increase `pkgrel` by 1.
* Increase `pkgver` with the today's date in same format.

Biswa96:
Increase `pkgver` with the today's date in same format.

lazka:
That's a large improvement, thanks!

PR #2251, opened 2020-12-09
autoconf: Update to 2.70

jannick0:
Autoconf 2.70 is a years long awaited package release. 

I would like to encourage you to test it before merging into the repository, especially for the enhanced feature for AUTOTEST testsuites to ignore EOL differences by default.  The compiled package is available in this PR's artifact.

Thanks.

---------

Highlights:

1. Any testsuite generated by AUTOTEST 2.70 ignores by default differences
   between line endings (LF, CRLF) of files or program outputs.  This is
   effectively achieved by adding AT TEST TIME the DIFF flag
   '--strip-trailing-cr' if available.

   More details below, especially as to how to prevent using this feature.

   The approach is similar to the previous MSYS2-package AUTOCONF 2.69,
   which tacitly unified the EOLs.  The new approach applies more broadly
   and with higher speed.

2. config.guess patched for MSYS2 recognition in a separate patch file
   for better maintenance.

Changes:
* PKGBUILD:
  - admin stuff (update version number, change patch file names, hashes
    etc.)
  - apply VPATH build (i.e. build in a directory other than src directory)
  - add TESTSUITEFLAGS for better performance of 'make check'  (However,
    running the test suite still last very very long!)
  - remove .sig input file

* 0001-autoconf-2.70-MSYS2-specifics-AC_CYGWIN.patch:
  - left-over amendments of 0002-msys2.patch (for AC-macro AC_CYGWIN
    marked as obsolete anyways).

* 0002-autoconf-2.70-MSYS2-Autotest-Unify-EOLs-By-Default.patch:
  This patch tackles the end-of-line (EOL) issue on Windows when running
  an AUTOTEST testsuite which might compare files and/or program outputs.

  This patch sets the default behaviour of any AUTOTEST testsuite generated
  with AUTOCONF 2.70: When running an AUTOTEST test suite and comparing
  files or program outputs, EOL differences (like LF vs CRLF) are ignored
  **BY DEFAULT**, iff the program DIFF allows the flag
  '--strip-trailing-cr'.

  In order to switch off this feature and run the testsuite mocking at
  differences between different EOL patterns, run the testsuite with the
  environment variable 'MSYS2_AUTOTEST_NO_UNIFY_EOL' set to '1', e.g.:

         tests/testsuite -C tests MSYS2_AUTOTEST_NO_UNIFY_EOL=1

  MSYS2_AUTOTEST_NO_UNIFY_EOL impacts the behaviour of a testsuite on a
  global basis, but the EOL handling can be set differently for each
  testsuite.

* 0003-autoconf-2.70-MSYS2-specifics-config.guess.patch:
  - create a separate patch for config.guess (from 0002-msys2.patch) for
    better maintenance (i.e. to be able to update config.guess, config.sub
    separately)

Removed patches:
* 0001-fix-texinfo.patch: issue fixed.
* 0002-msys2.patch: non-obsolete parts split to
  - 0001-autoconf-2.70-MSYS2-specifics-AC_CYGWIN.patch
  - 0003-autoconf-2.70-MSYS2-specifics-config.guess.patch
* 0003-autotest-remove-cr-from-prog-stdout-stderr.patch:
  - replaced by approach in 0001-autoconf-2.70-MSYS2-Autotest-Unify-EOLs.patch


lazka:
Thanks!

lazka:
I wonder what would a good package to test this with.

Biswa96:
The "Required By" section in [autoconf][1].

[1]: https://packages.msys2.org/package/autoconf?repo=msys&variant=x86_64

jannick0:
The "Required By" section is a subset of the packages really using `autoconf` (those requiring `base-devel` are not listed here, are they?). `pactree` doesn't provide more information, either.  I am wondering if `pacman` processes indirect dependencies to package groups well ...   

The `GNU` team shippes  `autoconf`, why not looking at other `GNU` packages which use the `autoconf` machinery (mandatorily I think), maybe the real basic ones like `coreutils`, `bison` and all their good close friends (technically `flex` is not `GNU`, but used to be, thus it is kinda, so include it to the list)?  
Get bonus points: Watch out that `check` runs with results as with `autoconf 2.69`. :o)

I primarily thought about manual tests when a new package release comes along.  A systematic CI test would be another and more difficult story, of course.

lazka:
coreutils fails with:

```
checking whether gcc accepts -g... yes
checking for gcc option to enable C11 features... none needed
../coreutils-8.32/configure: line 6430: syntax error near unexpected token `ac_cv_prog_cc_stdc=$ac_cv_prog_cc_c89'
../coreutils-8.32/configure: line 6430: `                      ac_cv_prog_cc_stdc=$ac_cv_prog_cc_c89'
==> ERROR: A failure occurred in build().
```

bison fails with:

```
checking for gcc option to enable C11 features... none needed
./configure: line 6259: syntax error near unexpected token `ac_cv_prog_cc_stdc=$ac_cv_prog_cc_c89'
./configure: line 6259: `                      ac_cv_prog_cc_stdc=$ac_cv_prog_cc_c89'
```

jannick0:
Many thanks for doing this!

This is not an issue for `autoconf 2.70`, but an upstream-upstream issue from GnuLib. This GnuLib [patch](http://git.savannah.gnu.org/gitweb/?p=gnulib.git;a=commitdiff;h=a3b3fc85e3e632374811b27cb2111e50fa177e36) should be expected to be streamed up into both `coreutils` and `bison` and potentially other GNU packages.
 
I have not checked if the above really works, but so far no need to update `autoconf 2.70`. I strongly suggest to wait for `autoconf 2.70.1` expected for early next year to merge.  I am marking this PR as WIP.

Xref: [Bug report](https://lists.gnu.org/archive/html/bug-autoconf/2020-12/msg00036.html) at bug-autoconf.

jannick0:
Mimique **by brute force** the patch mentioned above by removing the file `std-gnu11.m4` from the source code running
 ```
cd <src dir>
find . -type f -name 'std-gnu11.m4' -delete
autoreconf -vfi
bash -n configure  # check if configure script is syntactically correct
```
Checked for `bison 3.7.2`: This produces a syntactically correct `configure`.

lazka:
I see, thanks. Let's wait for the dust to settle a bit then (until there are new bison releases etc.)

lazka:
If you create a new package like 2.13 then we can include it earlier. Not sure if that makes sense etc..

lazka:
Arch has updated to 2.70 now

jannick0:
Thanks.  However, I would suggest to still wait for the next release which is supposed to come out soon early next year, since 2.70 has some flaws and is not fully compatible with gnulib - good luck with GNU packages which are not adapted yet.  ;)

Separately, is there any good way to have two release versions of the same package in the msys2 repo (here `autoconf 2.69` and `autoconf 2.7x`? [Defining a new `PKGBUILD` for `autoconf 2.69` with, say, `provides=(autoconf=2.69)` works well for `pacman` when running `makepkg` against a package with `makedepends=(autoconf=2.69)` ... except for a hick-up in the end phase when upgrading to `autoconf 2.7x` again, while the downgrade from `autoconf 2.7x` in the initial dependency check phase works. Any ideas?] 

BTW: The same might apply to `bison` which introduced features explicitly not backward compatible, i.e. `.y` files might not be understood by the latest `bison` version.  This might raise the need to have a tagged version and the current version of bison in the msys2 repo.  The `bison` team is in favor of the position that tarballs should be independent of running `bison` since its products are shipped with the tarball.   My personal view is that this makes the development of packages difficult if msys2 would not be able to ship tagged versions of specific packages (like `autoconf`, `bison` and hopefully not many more).

lazka:
ok, let's wait for the next release, test bison/coreutils and if that works just update.

re bison, I haven't had any problems with bison breaking (there were some 2.x -> 3.x things). We are a rolling distro, if dependencies build fine everything can be updated.

lazka:
Closing in favor of #2325

PR #2254, opened 2020-12-11
curl: rebuild with libzstd; fix+rebuild pacman

lazka:
somehow the libcurl.pc was missing libraries in Libs.private, but rebuilding it locally helped.. let's try again in CI.


lazka:
Time to compare build logs I guess...

PR #2256, opened 2020-12-11
autoconf2.13: move share dir from 'autoconf' to 'autoconf-2.13'

jannick0:
Not sure if this package is still needed.  

@lazka   This PR is a preparation step for keeping various autoconf version at the same time. We could later have packages `autoconf-2.69` and `autoconf` (=`autoconf-2.70.x`). Agree - or some other idea?  I am open for anything, but I think it would be helpful to keep `autoconf-2.69` in the play for some longer period of time to avoid massive complaints when packages are not prepared for `autoconf 2.70.x`.  If ok I can look to prepare the same for `autoconf 2.69`.
  
---
This commit aims at avoiding any interference with another autoconf release
package by keeping the library directory separate from other autoconf
release packages - WITHOUT breaking package functionality.

This approach might be consistently applied to other AUTOCONF versions
other than the latest one at the time.  Then the user can choose which
major AUTOCONF release version she wants to apply.

* PKGBUILD:
  - add and apply patch 0001-autoconf-2.13-version-specific-share-directory.patch
  - add missing dependency to PERL
* 0001-autoconf-2.13-version-specific-share-directory.patch:
  - add patch fixing library path

### DIFF
--- a/autoconf2.13/0001-autoconf-2.13-version-specific-share-directory.patch
+++ b/autoconf2.13/0001-autoconf-2.13-version-specific-share-directory.patch
@@ -1,0 +1,28 @@
+From 64c18ba1272c0158bb53459ba78306ab0e531b90 Mon Sep 17 00:00:00 2001
+Date: Fri, 11 Dec 2020 16:51:45 +0100
+Subject: [PATCH] autoconf-2.13-version-specific-share-directory
+
+* Makefile.in:
+  - ensure that datadir = '/usr/share/autoconf-2.13' is CONSISTENTLY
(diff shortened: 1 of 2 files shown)


lazka:
> Not sure if this package is still needed.

How can we figure this out? I'd prefer to remove it if possible.

lazka:
According to https://www.archlinux.org/packages/extra/any/autoconf2.13/ only firefox stuff. We might want to add js78 in the future, so let's keep it then.

lazka:
Thanks!

PR #2265, opened 2020-12-19
bash: Update to 5.1.004

lazka:
This is only moving changes around so they apply again. Not sure how to test this.. besides dogfooding it.


Alexpux:
Check first on local computer that it works. I try several times to update bash to 5.0 and it builds fine but not work

lazka:
It works fine here.

There was one problem where patch did a fuzzy match, inserting a change at the wrong place resulting in "cannot make pipe for command substitution" on start. This was the only problem I saw, but was easily fixed.

lazka:
Got two confirmations on gitter that it works, so merging.

PR #2266, opened 2020-12-19
Use zstd for all packages now

lazka:
It's been long enough and everyone should have a pacman that supports it


Biswa96:
> everyone should have a pacman that supports it

Let's have a Twitter poll.

lazka:
I think it was broken in the meantime anyway, since pacman depends on gnupg/curl and they use zst

jeremyd2019:
> It's been long enough and everyone should have a pacman that supports it

Maybe not, #2300

mati865:
Always there will be somebody trying to update from ancient version...

PR #2274, opened 2020-12-21
pacman: update to 6.0.0

lazka:
ParallelDownloads seem broken (they download sequentially here and the later ones hit the timeout limit). Maybe it's sourceforge, but if all fails we can disable it by default.


lazka:
It's sourceforge blocking multiple connections, so we need to default to non-parallel downloads (it doesn't add much anyway..)

lazka:
rebased on master, disabled parallel downloads per default because of sourceforge

naveen521kk:
sourceforge?  is it a thing to worry about now?

lazka:
no

lazka:
6.0 is out. I'll try to rebase this in the coming days.

One regression came up on the pacman IRC re corrupt downloads on broken mirrors, so we should wait for a fix before replacing 5.x.

lazka:
> One regression came up on the pacman IRC re corrupt downloads on broken mirrors, so we should wait for a fix before replacing 5.x.

pacman got released without this so I guess not that important

PR #2278, opened 2020-12-22
mingw-w64-cross-compiler-rt and -clang i686

jeremyd2019:
Simple loop to build both x86_64 and i686 target libraries, as is done in -crt.  I have not actually looked at the resulting binaries.

### DIFF
--- a/mingw-w64-cross-clang/PKGBUILD
+++ b/mingw-w64-cross-clang/PKGBUILD
@@ -4,7 +4,7 @@ _realname=clang
 _mingw_suff=mingw-w64-cross
 pkgname=("${_mingw_suff}-${_realname}")
 pkgver=11.0.0
-pkgrel=2
+pkgrel=3
 pkgdesc="Cross Clang for the MinGW-w64"
 groups=("${_mingw_suff}-${_realname}-toolchain")
 arch=('i686' 'x86_64')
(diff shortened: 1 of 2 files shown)


mati865:
The problem with i686 is we cannot use 32bit LLD to link big packages (like LLVM) : https://bugs.llvm.org/show_bug.cgi?id=42388#c6

jeremyd2019:
> The problem with i686 is we cannot use 32bit LLD to link big packages (like LLVM)

☹️ 

Does this PR look OK though?

mati865:
Yeah, PR itself looks fine.

jeremyd2019:
I built this locally on my i686 machine last night, and built a quick test c++ program with it just now that succeeded and ran properly.

lazka:
It doesn't complicate things too much so seems fine I guess, even if it doesn't work in the long run.

@mati865 ok to merge?

mati865 (approved):

jeremyd2019:
I was thinking/hoping that this should make adding arm/aarch64 pretty straightforward too.  Though I don't have any such hardware to try it myself.

mati865:
Does Cygwin work on ARM/Aarch64?

jeremyd2019:
My understanding is that it at least runs under emulation.  So it seems to me it could be used to bootstrap a native toolchain.  Theoretically.  Something interesting for those with the interest and hardware to try, anyway.  I think I remember @Biswa96 talking about this at one point.

mati865:
I think Cygwin taps too deeply into the system for emulation to work reliably.
Cross compilation will be possible though.

PR #2286, opened 2020-12-28
pacman: makepkg-mingw improvements

### LINKED ISSUE #2285: makepkg-mingw: behaves differently in different shells
CONFIG_SITE, MSYSTEM_CARCH, MSYSTEM_CHOST are different when called in msys shell vs a mingw shell.

imo there shouldn't be a difference

We currently use a mingw shell in PR CI but a msys shell in autobuild, so sometimes things pass in CI and then fail in autobuild.

lazka:
* Fix various shellcheck issues; like for spaces in arguments etc
* Merge makepkg-clang into makepkg-mingw. No need to have an extra tool for starters.
  MINGW_INSTALLS="clang32 clang64" will work now.
* Run makepkg in a new login shell for the right MSYSTEM. This reduces
  the differences between running it in a msys shell vs a mingw shell,
  see #2285

Fixes #2285


lazka:
@Alexpux feedback welcome

PR #2288, opened 2020-12-29
pacman: change clang package prefix; sync clang hardening flags

lazka:


lazka:
@mati865

pacman/makepkg_clang32.conf
```diff
@@ -59,9 +59,9 @@ DXSDK_DIR=${MINGW_PREFIX}/${MINGW_CHOST}
 CPPFLAGS="-D__USE_MINGW_ANSI_STDIO=1"
 CFLAGS="-march=i686 -mtune=generic -O2 -pipe"
 CXXFLAGS="-march=i686 -mtune=generic -O2 -pipe"
-LDFLAGS="-pipe"
-# Uncomment to enable hardening (ASLR, DEP)
-#LDFLAGS="-pipe -Wl,--dynamicbase,--nxcompat"
+LDFLAGS="-pipe -Wl,--dynamicbase,--nxcompat,--no-seh"
```
mati865, line 62: LLD silently ignores `--nxcompat` because it's always enabled.
Do we still want to include it?

Biswa96: Maybe we can keep that just for similarity with makepkg_mingw* one 🤷

pacman/makepkg_clang64.conf
```diff
@@ -59,9 +59,9 @@ DXSDK_DIR=${MINGW_PREFIX}/${MINGW_CHOST}
 CPPFLAGS="-D__USE_MINGW_ANSI_STDIO=1"
 CFLAGS="-march=x86-64 -mtune=generic -O2 -pipe"
 CXXFLAGS="-march=x86-64 -mtune=generic -O2 -pipe"
-LDFLAGS="-pipe"
-# Uncomment to enable hardening (ASLR, High entropy ASLR, DEP)
-#LDFLAGS="-pipe -Wl,--dynamicbase,--high-entropy-va,--nxcompat"
+LDFLAGS="-pipe -Wl,--dynamicbase,--high-entropy-va,--nxcompat,--default-image-base-high"
```
mati865, line 62: Like above `--high-entropy-va` and `--nxcompat` are silently ignored since they are always enabled by LLD.

Upstream LLD doesn't understand `--default-image-base-high`, I have patched our LLD to silently ignore this option since it does the right thing by the default. I don't want to keep that patch for longer than necessary so IMO this specific flag is actually harmful with no benefits.

pacman/makepkg_clang32.conf
lazka: ok, I'll remove it. I made sure lld doesn't fail with them, but I didn't know if they were useful.

pacman/makepkg_clang64.conf
lazka: ok, sounds good

pacman/makepkg_clang32.conf
lazka: @Biswa96 we want to get rid of them in makepkg_mingw as well eventually, once they are default in binutils, so this is fine imo

mati865 (approved):

PR #2293, opened 2020-12-30
Add guile 3.0 (3.0.4) but in separate package dir than guile(2)

h-c-mueller:
This allows parallel installation of libguile and does not break existing
dependencies to libguile 2.2. Therefore the libguile package name includes
the version 3.0, i.e. libguile3.0-3.0.4-1-i686.pkg.tar.zst


lazka:
guile is potentially used by remake, autogen, elinks-git.

What's preventing us from just updating 2.2 to 3.x?

h-c-mueller:
I'm not aware of any general problem to directly update from 2.2 to 3.0. But this requires recompilation of all dependent packages. My initial idea was to allow for a smother transition without the need to abruptly change all dependent packages from 2.2 to 3.0.
But if you request, I can rewrite the patch to directly update from 2.2 to 3.0.

PR #2296, opened 2021-01-01
filesystem: add support for MSYSTEM=CLANG{32,64}

jeremyd2019:
I was playing with the https://github.com/msys2/CLANG-packages stuff, and found that I needed to adjust `/etc/msystem` *and* `/etc/profile` in addition to the `makepkg_clang*.conf`.  Otherwise `MINGW_CHOST` was getting unset.

Regardless, the handling in `/etc/profile` was duplicative, and this seems to make things easier moving forward, just requiring changes to `/etc/msystem` for new MINGW or CLANG `MSYSTEM` values.


Biswa96:
But `$MINGW_PREFIX` is different for mingw* and clang* ones.

Biswa96:
But `$MINGW_PREFIX` is different for mingw* and clang* ones i.e. /mingw* and /clang* respectively. Aren't they?

jeremyd2019:
Yes, but that's set in /etc/msystem, sourced from /etc/profile.

lazka:
lgtm

jeremyd2019:
OK.  I figured it wouldn't make sense to bump pkgver/pkgrel without completing adding the CLANG options to msystem.

PR #2298, opened 2021-01-03
rpcsvc-proto: Add package for libvirt build

Biswa96:
This probably fixes the build issue of new libvirt, https://github.com/msys2/MINGW-packages/issues/6384 /cc @Alexpux


Biswa96:
Do not merge without Alexpux comment.

PR #2299, opened 2021-01-04
p7zip: update to 17.03

Astrum-polaris:
switch to maintained fork like arch did


lazka:
Thanks! lib/7zCon.sfx is missing compare to the old version. Is this a problem?

Astrum-polaris:
The old version uses `make all3`, which is essentially sfx+7z+7za+7zr, that's why sfx is missing in the previous commit
Restored that part should resolve this problem.

lazka:
Thanks!

PR #2304, opened 2021-01-10
[RFC] pacman: patch for makepkg MINGW_PREFIX stripping from package function name

jeremyd2019:
If a normally named package function `package_$pkgname` is not found, try to find one with MINGW_PREFIX stripped from the pkgname.

https://github.com/msys2/MSYS2-packages/discussions/2289#discussioncomment-270164

This turned out to be more invasive that I had hoped, a surprising number of places knew that package functions were named `package_$pkgname`.  I replaced those with a call to a function.


lazka:
Random though, what about injecting prefixed wrappers where it is sourced instead? (Not sure if this is easier)

jeremyd2019:
> Random though, what about injecting prefixed wrappers where it is sourced instead? (Not sure if this is easier)

My concern with that would be that it may not "know" enough that early.  Especially with all of the lint functions, it may not even be safe to assume the PKGBUILD is well-formed yet.

lazka:
hm, right, I'm slightly in favor of https://github.com/msys2/MINGW-packages/blob/03fb106a6e6db501fc9d0350b342f4b940bc8074/mingw-w64-gtk3/PKGBUILD#L138 because it is less hidden magic

jeremyd2019:
That would be OK with me.  One thing that this patch just turned up is that the linter for package function variables detected that `depends` is not an array [here](https://github.com/msys2/MINGW-packages/blob/a42579195ecee52e464528687d2272e0a3b7fcbd/mingw-w64-clang/PKGBUILD#L435) and [here](https://github.com/msys2/MINGW-packages/blob/a42579195ecee52e464528687d2272e0a3b7fcbd/mingw-w64-clang/PKGBUILD#L455) in mingw-w64-clang.  It apparently couldn't do that when those functions were wrapped, because the build succeeded rather than erroring out in the 'lint' functions.

lazka:
uh, good point. "MINGW_INSTALLS=mingw64 makepkg-mingw --printsrcinfo" fails to pick up depends on both cases

ok, even more hacky, but seems to handle all cases:

```bash
for _name in "${pkgname[@]}"; do
  _short="package_${_name#${MINGW_PACKAGE_PREFIX}-}"
  _func="$(declare -f "$_short")"
  eval "${_func/#$_short/package_$_name}"
done
```

jeremyd2019:
and that's still preferable to patching `makepkg`?

lazka:
not sure, I tend towards yes, but I'd leave this open.

jeremyd2019:
This patch injects wrappers in `source_buildfile`.  One concern is that the build file is potentially sourced again later, if there is a `pkgver` function, after it updates the `pkgver` variable.  I believe this to be safe, as the contents of the functions should not change (absent some contrived situation like conditionally defining functions based on the `pkgver` value).  Another concern is that error messages and output could be confusing, referring to functions that don't appear to exist in the build file.

jeremyd2019:
Rebased to pick up latest pacman changes, and pkgrel bumped again

Biswa96:
`pkgdesc` is removed?

jeremyd2019:
Fixed merge conflict resolution fail

jeremyd2019:
I am trying to come up with a good way to avoid duplicating that loop.  I took a look at `makepkg-template`, but it seems to rely on symlinks in a way that doesn't seem like it will work on msys2.  I noticed https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-PKGBUILD-common - perhaps a file in there could contain the loop, and be sourced by split package PKGBUILDs... Would that be acceptable?

jeremyd2019:
I think the downside of sourcing a script from PKGBUILDs would be that the sourced file wouldn't be included in the source package, so it wouldn't be self-contained.  I think `makepkg-template` could actually work, if always referring to templates by name AND version, so it wouldn't try to resolve the symlink to the 'latest' version.

jeremyd2019:
It turns out that `makepkg-template` does work with version included.  msys2/MINGW-packages#7931

PR #2320, opened 2021-01-26
pacman: Advise clang not toolchain for MINGW_INSTALL=clang*

mingwandroid:


mati865:
It's still not correct though, `/clangXY` subsystem or whatever it will be will use it's own package instead of `/mingwXY/bin/clang`.

jeremyd2019:
Also sha256sums need to be updated

mingwandroid:
Yeah. This is more of a crossing thing for now. Happy to drop this til the clang dust settles!

PR #2321, opened 2021-01-27
pacman: clean up makepkg-mingw changes

jeremyd2019:
I believe this addresses the issues with #2320 @mingwandroid @mati865

The package names are derived from the `MINGW_PREFIX`es from makepkg_clang{32,64}.conf.  It looks like the paths would come out right because `${_mingw}` is what is being switched on.


mingwandroid:
Thanks

PR #2325, opened 2021-02-05
Update autoconf and automake1.16

Astrum-polaris:
I cherrypicked the commit in #2251 as the base and add more commits.
Tested bison and coreutils locally, both build successfully with autoconf 2.71


lazka:
@jannick0 does this look OK to you?

lazka:
let's assume it is

jannick0:
@lazka Sorry for the late reply, was kind of out of order. Changes look fine and appear to work, since no complaints so far.

PR #2327, opened 2021-02-06
mintty: fix argv problem when launching from launcher

erw7:
If the command to be executed is not specified in the command line option, the following code will determine the command to be executed. 

https://github.com/mintty/mintty/blob/2506d2be13a0fe6889d0f815ea9192d7cb6e54b6/src/winmain.c#L5215-L5220
```c
    cmd = getenv("SHELL");
    cmd = cmd ? strdup(cmd) :
#if CYGWIN_VERSION_DLL_MAJOR >= 1005
      (pw && pw->pw_shell && *pw->pw_shell) ? strdup(pw->pw_shell) :
#endif
      "/bin/sh";
```

Nevertheless, the MSYS2 launcher passes `{ "/usr/bin/bash", "--login", NULL }` as `argv` to `execvp()`. `zsh` will run in compatibility mode if the command name in the `argv[0]` value starts with &#x60;b', &#x60;s', or &#x60;k'. As a result, when `zsh` is invoked from laucher by user configuration, `zsh` will run in compatibility mode. `/etc/profile` sets a value in `PS1` that is not treated as a special character in `zsh` compatibility mode. It has the following disastrous consequences. This PR fixes that problem.

image

### DIFF
--- a/mintty/0002-add-msys2-launcher.patch
+++ b/mintty/0002-add-msys2-launcher.patch
@@ -26,12 +26,12 @@ registry value (or just the entire MuiCache directory) should help, at
 least for the current Windows 10.
 ---
  src/appinfo.h  |   2 +-
- src/launcher.c | 129 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+ src/launcher.c | 115 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++
  src/launcher.h |  12 ++++++
  src/res.h      |   4 ++
- src/res.rc     |  16 ++++++-
- src/winmain.c  |  34 ++++++++++++---
- 6 files changed, 189 insertions(+), 8 deletions(-)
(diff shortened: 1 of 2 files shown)


lazka:
Would dropping that whole patch also help you?

lazka:
just wondering: Why are you using the mintty launcher? We were thinking about dropping this patch alltogether... :/

erw7:
> just wondering: Why are you using the mintty launcher? We were thinking about dropping this patch alltogether... :/

I don't usually use launcher, but when I launched `mintty` without specifying any command line options to see how it worked, I noticed that the prompt for `zsh` launched from launcher was strange. Well, I didn't pursue the cause until now because I rarely encounter it, but it was getting depressing, so I fixed it.

It looks like there are plans to drop this entire patch, so this will be closed.

PR #2337, opened 2021-02-10
filesystem: Avoid trying to chmod /dev/{shm,mqueue}

elieux:
Although the default and most used way to install MSYS2 is to have it
user-writable, some users prefer to have the root writable only to
admins.  In that case, the 01-devices.post script would always print out
two failures on "login" when trying to chmod /dev/{shm,mqueue}.  Since
the only supported configuration for MSYS2 is to mount `/` with noacl
where chmod() is a no-op, the failure is purely cosmetic, as is the
chmod invocation itself.  For these reasons, let's not chmod at all.

I'm aware the non-user-writable root configuration itself is not
officially supported either, but I've been using it like that for years
now and this is the only issue I know of, so I think it's reasonable to
resolve this.

### DIFF
--- a/filesystem/01-devices.post
+++ b/filesystem/01-devices.post
@@ -52,8 +52,6 @@ maybe_create_devs ()
       echo "POSIX semaphores and POSIX shared memory will not work"
       echo
     fi
-  else
-    chmod 1777 "${DEVDIR}/shm"
   fi
 
   # Check for ${DEVDIR}/mqueue directory (for POSIX message queues)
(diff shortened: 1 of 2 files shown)


lazka (approved):

mingwandroid:
LGTM

elvisdukaj:
> I'm aware the non-user-writable root configuration itself is not officially supported either, but I've been using it like that for years now and this is the only issue I know of, so I think it's reasonable to resolve this.

Can I ask how did you resolved it? The Msys2 version in Conan has to be immutable. One way is not to have this folders at all but this brings to annoying messages that `/dev/shm` and `/dev/mqueu` are read-only warning messages. What would be the best way to avoid them completely?

PR #2345, opened 2021-02-18
binutils: update to 2.36.1

Astrum-polaris:
Remove testsuite patches to reduce maintenance effort. 

I guess they are not needed since we didn't run the testsuite.

### DIFF
--- a/binutils/0100-binutils-2.36-msys2.patch
+++ b/binutils/0100-binutils-2.36-msys2.patch
@@ -1,6 +1,6 @@
-diff -Naur binutils-2.30.orig/bfd/acinclude.m4 binutils-2.30/bfd/acinclude.m4
---- binutils-2.30.orig/bfd/acinclude.m4	2018-01-13 16:31:15.000000000 +0300
-+++ binutils-2.30/bfd/acinclude.m4	2018-06-13 08:19:13.199163100 +0300
+diff -Nbadur binutils-2.36.1.orig/bfd/acinclude.m4 binutils-2.36.1/bfd/acinclude.m4
+--- binutils-2.36.1.orig/bfd/acinclude.m4	2021-01-09 18:47:33.000000000 +0800
++++ binutils-2.36.1/bfd/acinclude.m4	2021-02-18 17:57:15.468229900 +0800
 @@ -21,7 +21,7 @@
  [AC_REQUIRE([AC_CANONICAL_TARGET])
  case "${host}" in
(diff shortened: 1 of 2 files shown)


jeremyd2019:
This will be interesting.  I never noticed if anyone made a conscious decision as to what the default dll characteristics should be on "cygwin"

lazka:
I built binutils itself with this and things seemed to work. I hope that rules out any major breakage.

jeremyd2019:
binutils is making msys2/cygwin binaries with dynamicbase now by default, including `msys-2.0.dll` in `msys2-runtime`.  I kind of expected this to break badly, but a quick test is showing that it does seem to work.

lazka:
thanks for testing. hmm.. google gives me https://cygwin.cygwin.narkive.com/zhDhwcA8/two-probable-basing-issues-causing-fork-failures-1-cygreadline7-dll-has-aslr-enabled-2-default-base- which seems to suggest that we mught just be.lucky. Maybe we should ask on the cygwin mailing list?

lazka:
I've asked here: https://cygwin.com/pipermail/cygwin/2021-February/247922.html

lazka:
I've created #2356 to change the defaults

lazka:
ok, upstream agrees. I'm still not sure about nxcompat though.

jeremyd2019:
As I recall, there are already #ifdefs in those files for cygwin (ie, for base addresses), so this would just be one more around the `DEFAULT_DLL_CHARACTERISTICS`.

That also calls into question the concerns about the higher base addresses that was brought up on the list, as the default base addresses for cygwin should not have changed.

As far as the "almost works, then doesn't until reboot" behavior (with cygwin/msys2), I have seen that on 32-bit.   I had *hoped* that 64-bit would not be subject to that (or at least much less likely to, to the point of not worrying about.  I could write a long description why I think that should be the case, but it'd probably be better served on the cygwin list if anywhere).

jeremyd2019:
Untested patch against upstream master.  Hopefully try sending it upstream and/or get comment from cygwin folks after some testing.  Obviously would require fixing up the -msys2 patch here

```diff
From 39bde2f11a1eb97503bae9cf15f5fc05640e5251 Mon Sep 17 00:00:00 2001
From: Jeremy Drake <sourceware-bugzilla@jdrake.com>
Date: Sun, 28 Feb 2021 15:49:08 -0800
Subject: [PATCH] ld: revert default dll characteristics for Cygwin.

Mail thread from
https://cygwin.com/pipermail/cygwin/2021-February/247922.html suggests
these flags will NOT work for Cygwin, which relies on stable address
layouts for their fork() emulation.

In the process, renamed move_default_addr_high shell variable to
cygwin_beahior, as the old name wasn't quite accurate anymore and I
wanted to use it choose which dll characteristics flags to use by
default.

Also copied that switch to pe.em, as it was only in pep.em before but
32-bit also needed to switch defaults for Cygwin.
---
 ld/emultempl/pe.em  | 13 ++++++++++++-
 ld/emultempl/pep.em | 40 +++++++++++++++++++++-------------------
 2 files changed, 33 insertions(+), 20 deletions(-)

diff --git a/ld/emultempl/pe.em b/ld/emultempl/pe.em
index 748a6b49412..9f757cc31dc 100644
--- a/ld/emultempl/pe.em
+++ b/ld/emultempl/pe.em
@@ -5,6 +5,16 @@ if [ -z "$MACHINE" ]; then
 else
   OUTPUT_ARCH=${ARCH}:${MACHINE}
 fi
+
+case ${target} in
+  *-*-cygwin*)
+    cygwin_behavior=1
+    ;;
+  *)
+    cygwin_behavior=0;
+    ;;
+esac
+
 rm -f e${EMULATION_NAME}.c
 (echo;echo;echo;echo;echo)>e${EMULATION_NAME}.c # there, now line numbers match ;-)
 fragment <<EOF
@@ -104,7 +114,8 @@ fragment <<EOF
 #define DEFAULT_PSEUDO_RELOC_VERSION 1
 #endif

-#define DEFAULT_DLL_CHARACTERISTICS    (IMAGE_DLL_CHARACTERISTICS_DYNAMIC_BASE \
+#define DEFAULT_DLL_CHARACTERISTICS    (${cygwin_behavior} ? 0 : \
+                                          IMAGE_DLL_CHARACTERISTICS_DYNAMIC_BASE \
                                         | IMAGE_DLL_CHARACTERISTICS_NX_COMPAT)

 #if defined(TARGET_IS_i386pe) || ! defined(DLL_SUPPORT)
diff --git a/ld/emultempl/pep.em b/ld/emultempl/pep.em
index ff49c15c002..69f5bd04655 100644
--- a/ld/emultempl/pep.em
+++ b/ld/emultempl/pep.em
@@ -8,10 +8,10 @@ fi

 case ${target} in
   *-*-cygwin*)
-    move_default_addr_high=1
+    cygwin_behavior=1
     ;;
   *)
-    move_default_addr_high=0;
+    cygwin_behavior=0;
     ;;
 esac

@@ -99,45 +99,47 @@ fragment <<EOF
 #define DLL_SUPPORT
 #endif

-#define DEFAULT_DLL_CHARACTERISTICS    (IMAGE_DLL_CHARACTERISTICS_DYNAMIC_BASE \
+#define DEFAULT_DLL_CHARACTERISTICS    (${cygwin_behavior} ? 0 : \
+                                          IMAGE_DLL_CHARACTERISTICS_DYNAMIC_BASE \
                                         | IMAGE_DLL_CHARACTERISTICS_HIGH_ENTROPY_VA \
                                         | IMAGE_DLL_CHARACTERISTICS_NX_COMPAT)

+
 #if defined(TARGET_IS_i386pep) || ! defined(DLL_SUPPORT)
 #define        PE_DEF_SUBSYSTEM                3
 #undef NT_EXE_IMAGE_BASE
 #define NT_EXE_IMAGE_BASE \
-  ((bfd_vma) (${move_default_addr_high} ? 0x100400000LL \
-                                       : 0x140000000LL))
+  ((bfd_vma) (${cygwin_behavior} ? 0x100400000LL \
+                                : 0x140000000LL))
 #undef NT_DLL_IMAGE_BASE
 #define NT_DLL_IMAGE_BASE \
-  ((bfd_vma) (${move_default_addr_high} ? 0x400000000LL \
-                                       : 0x180000000LL))
+  ((bfd_vma) (${cygwin_behavior} ? 0x400000000LL \
+                                : 0x180000000LL))
 #undef NT_DLL_AUTO_IMAGE_BASE
 #define NT_DLL_AUTO_IMAGE_BASE \
-  ((bfd_vma) (${move_default_addr_high} ? 0x400000000LL \
-                                       : 0x1C0000000LL))
+  ((bfd_vma) (${cygwin_behavior} ? 0x400000000LL \
+                                : 0x1C0000000LL))
 #undef NT_DLL_AUTO_IMAGE_MASK
 #define NT_DLL_AUTO_IMAGE_MASK \
-  ((bfd_vma) (${move_default_addr_high} ? 0x1ffff0000LL \
-                                       : 0x1ffff0000LL))
+  ((bfd_vma) (${cygwin_behavior} ? 0x1ffff0000LL \
+                                : 0x1ffff0000LL))
 #else
 #undef  NT_EXE_IMAGE_BASE
 #define NT_EXE_IMAGE_BASE \
-  ((bfd_vma) (${move_default_addr_high} ? 0x100010000LL \
-                                       : 0x10000LL))
+  ((bfd_vma) (${cygwin_behavior} ? 0x100010000LL \
+                                : 0x10000LL))
 #undef NT_DLL_IMAGE_BASE
 #define NT_DLL_IMAGE_BASE \
-  ((bfd_vma) (${move_default_addr_high} ? 0x110000000LL \
-                                       : 0x10000000LL))
+  ((bfd_vma) (${cygwin_behavior} ? 0x110000000LL \
+                                : 0x10000000LL))
 #undef NT_DLL_AUTO_IMAGE_BASE
 #define NT_DLL_AUTO_IMAGE_BASE \
-  ((bfd_vma) (${move_default_addr_high} ? 0x120000000LL \
-                                       : 0x61300000LL))
+  ((bfd_vma) (${cygwin_behavior} ? 0x120000000LL \
+                                : 0x61300000LL))
 #undef NT_DLL_AUTO_IMAGE_MASK
 #define NT_DLL_AUTO_IMAGE_MASK \
-  ((bfd_vma) (${move_default_addr_high} ? 0x0ffff0000LL \
-                                       : 0x0ffc0000LL))
+  ((bfd_vma) (${cygwin_behavior} ? 0x0ffff0000LL \
+                                : 0x0ffc0000LL))
 #undef  PE_DEF_SECTION_ALIGNMENT
 #define        PE_DEF_SUBSYSTEM                2
 #undef  PE_DEF_FILE_ALIGNMENT
--
2.30.1.windows.1
```

lazka:
thanks!

PR #2348, opened 2021-02-22
fish: Fix msystem variables to be exported

BasixKOR:
I noticed fish doesn't get all of its supposed PATH when used as login shell, and I discovered /etc/msystem is not sourced correctly.

Unfortunately I couldn't find a way to source it nicely, I created msystem port for fish. I tested this on my machine.


lazka:
so, why doesn't /etc/msystem work?

1480c1:
> so, why doesn't /etc/msystem work?

I am assuming it's because it's running it in its own `sh`, but that doesn't cause it to pull the child process' environment into fish, and it seems fish does not adhere to posix shell scripting syntax

lazka:
That source part makes sense, I was just wondering about the syntax differences. But seems that it's wildly different, so I guess this makes sense.

PR #2354, opened 2021-02-27
pacman-mirrors: Enable the main repo again

elieux:
We have migrated repo.msys2.org to a dedicated server, thanks to
~Appfleet.io~Appfleet.com.


lazka:
lgtm

PR #2363, opened 2021-03-05
makepkg-mingw: Rename MINGW_INSTALLS to MINGW_ARCH

lazka:
In addition to this we want to add a `mingw_arch=(mingw32 mingw64)` for PKGBUILDs
in the future which tells us which type of builds a PKGBUILD supports.

The goal is to be more consistent and avoid confusion.

Other ideas welcome (we can change anything..)


lazka:
@elieux

PR #2365, opened 2021-03-09
Update to git 2.30.2, which fixes a security vulnerability

mapitman:
https://github.com/git/git/security/advisories/GHSA-8prw-h3cq-mghm


lazka:
Thanks, see #2366

PR #2369, opened 2021-03-13
pixz package

fshp:
The original PKGBUILD was taken from the archlinux with small changes


Biswa96:
Do you want this as a msys2 package? Then the program will depend on msys2 DLL (fork of cygwin).

1480c1:
Considering the support windows patch, I would assume MINGW-packages would be a better location

pixz/PKGBUILD
```diff
@@ -0,0 +1,32 @@
...
+pkgname=pixz
+pkgver=1.0.7
+pkgrel=2
+pkgdesc="Parallel, indexed xz compressor"
+arch=('i686' 'x86_64')
+url="https://github.com/vasi/pixz"
+license=('custom')
+depends=('libarchive')
+makedepends=('asciidoc' 'libarchive-devel')
+source=(https://github.com/vasi/pixz/releases/download/v$pkgver/${pkgname}-$pkgver.tar.gz)
```
1480c1, line 12: consider deriving this link from the url variable you provided, in case this URL changes to another gh repo, it would be easier to update just one line

pixz/PKGBUILD
```diff
@@ -0,0 +1,32 @@
+# Contributor: Maxim Moseychuk ("fshp") <franchesko.salias.hudro.pedros@gmail.com>
+
+pkgname=pixz
+pkgver=1.0.7
+pkgrel=2
```
1480c1, line 5: since this is still a "new" package in the eyes of the repo, I don't think pkgrel should be 2

fshp:
By bad.
I first compiled the package with a mingw. And a patch was required.
But for the msys the patch is not needed.
I think the option without patches is better. Or pure exe is better?

pixz/PKGBUILD
```diff
@@ -0,0 +1,32 @@
...
+pkgdesc="Parallel, indexed xz compressor"
+arch=('i686' 'x86_64')
+url="https://github.com/vasi/pixz"
+license=('custom')
```
1480c1, line 9: https://github.com/vasi/pixz/blob/master/LICENSE
seems to indicate that it is BSD rather than custom

1480c1:
> I think the option without patches is better. Or pure exe is better?

If it compiles fine and works as expected without patch, then without patch is fine

pixz/PKGBUILD
fshp: I changed. But I keep the license file
https://wiki.archlinux.org/index.php/PKGBUILD#license

fshp:
Try the latest pixz version

lazka:
I tried git master. but maybe I missed something..

goyalyashpal:
> pixz _(https://man.archlinux.org/man/pixz.1)_
> 
> [NAME](https://man.archlinux.org/man/pixz.1#NAME)
> pixz - parallel, indexed xz compressor
>
> [DESCRIPTION](https://man.archlinux.org/man/pixz.1#DESCRIPTION)
pixz compresses and decompresses files using multiple processors. If the input looks like a tar(1) archive, it also creates an index of all the files in the archive. This allows the extraction of only a small segment of the tarball, without needing to decompress the entire archive.

hmm, nice 😃

PR #2385, opened 2021-03-18
Add neofetch

mapitman:
Neofetch is a command-line system information tool written in bash 3.2
https://github.com/dylanaraps/neofetch

### DIFF
--- a/neofetch/PKGBUILD
+++ b/neofetch/PKGBUILD
@@ -1,0 +1,21 @@
+
+_realname=neofetch
+pkgname=${_realname}
+pkgver=7.1.0
+pkgrel=1
+pkgdesc="neofetch - A command-line system information tool written in bash 3.2+"
+arch=('any')
+license=('MIT')
+url='https://github.com/dylanaraps/neofetch'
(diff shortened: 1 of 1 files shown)


Biswa96:
Though I'm not against this, I have used netofetch in Linux distribuition, but what does someone want to know in Windows?

mapitman:
It's kind of a novelty. Today a friend sent me a screenshot of the "about" window from macOS to show me how much RAM, etc. his new Mac has. I responded with a screenshot  of neofetch. I first checked if there was a package, but I had to clone the repo and install it. I just thought maybe someone else may find it useful.

mapitman:
I noticed on another PR that "new packages are not accepted into MSYS2-packages". Does that mean I should close this and open a new PR against MINGW-packages?

mapitman:
I created a [pull request](https://github.com/msys2/MINGW-packages/pull/9499) in msys2/MINGW-packages as that seems like the more correct place for this.

lazka:
if it's bash only then this is the correct repo

PR #2395, opened 2021-03-21
Add mingw-w64-cross-ucrt-headers/crt

lazka:


Biswa96:
I was thinking of adding a flag like variable in already cross toolchain pkgbuild files, like mingw-w64-qt5 has static option.

lazka:
yeah, de-duplication would be nice.

(One problem is that we currently use the checksum of the PKGBUILD to detect changes, so sourcing things breaks that, unless you change a comment every time or something)

PR #2401, opened 2021-03-24
cmake: Update to 3.20.0

orgads:


lazka:
Could you briefly describe your patch changes? why/what etc..

orgads:
Sure. `perl-libs` is unused. I merged `disable-response-files` into the large patch, applied it on top of cmake 3.20, resolved conflicts and recreated it with Git (and renamed it because the version number was misleading).

lazka:
I see, thanks!

orgads:
I also filed a [MR for cmake](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5933) with this patch, to ease later updates.

PR #2414, opened 2021-03-30
msys2-runtime: upgrade to v3.2.0

dscho:
Apart from setting `pkgver` and re-setting `pkgrel`, all of the work was performed by the `./update-patches.sh` script, picking up the patches I had prepared in https://github.com/msys2/msys2-runtime/pull/30.


dscho:
Cc: @lazka

lazka:
Thanks!

lazka:
I'm gonna let it sit in staging for a day or two.

lazka:
One difference is that now those errors like "missing DLLs" pop up a Windows dialog and block, like

image

and

image

On the plus side this gives better info, but it also just hangs in CI forever if things fail. I wonder why that is.

dscho:
> On the plus side this gives better info, but it also just hangs in CI forever if things fail.

That's _bad_. I scoured the commits that came in since v3.1.7, and I have a hunch that the root cause for that new behavior might https://github.com/msys2/msys2-runtime/commit/21ec498d7f91285018e74f9767fe46cebace682a

dscho:
Oh, this is https://github.com/msys2/msys2-runtime/pull/18!

dscho:
How about hiding this behavior behind `winjitdebug`, as @jeremyd2019 suggested [here](https://github.com/msys2/msys2-runtime/pull/18#issuecomment-724880792)?

dscho:
> 
> 
> How about hiding this behavior behind `winjitdebug`, as @jeremyd2019 suggested [here](https://github.com/msys2/msys2-runtime/pull/18#issuecomment-724880792)?

I implemented this and opened the PR here: https://github.com/msys2/msys2-runtime/pull/35

jeremyd2019:
Has/should this unfortunate behavior been reported upstream to cygwin?  Perhaps they would reconsider an option.

dscho:
> 
> 
> Has/should this unfortunate behavior been reported upstream to cygwin? Perhaps they would reconsider an option.

I haven't had the time to send this patch (with corroborating rationale) to the Cygwin mailing list. Feel free to do that!

PR #2415, opened 2021-03-31
msys2-runtime: make default Windows error reporting an opt-in

dscho:
With this change we won't show that annoying message box e.g. when a DLL is missing, unless `MSYS` contains `winjitdebug`.

This PR is essentially forwarding https://github.com/msys2/msys2-runtime/pull/35 to MSYS2-packages.

Cc: @lazka @jeremyd2019


mingwandroid (commented):
LGTM. Thank you all!

PR #2417, opened 2021-04-01
Update packages

Biswa96:


neheb:
Curious. Why does file not include libmagic?

Biswa96:
> Curious. Why does file not include libmagic?

Answer https://en.wikipedia.org/wiki/File_(command)

neheb:
On Wed, Mar 31, 2021 at 10:33 PM Biswapriyo Nath ***@***.***>
wrote:

> Curious. Why does file not include libmagic?
>
> Answer https://en.wikipedia.org/wiki/File_(command)
>
Sure but libmagic is a separate library that other programs use. It doesn't
seem packaged with file.

Biswa96:
file(1) already has those things in source which is compiled as msys-magic-1.dll.

neheb:
Ah I see. It's MSYS only. No MinGW.

PR #2420, opened 2021-04-01
Add Windows terminal for MinGW

pchemguy:
Some programs to be built by MinGW expect Windows terminal (see https://cmake.org/cmake/help/latest/generator/MinGW%20Makefiles.html). This commit adds a launch mode based on the standard Windows terminal with MinGW bin in the path.

### DIFF
--- a/filesystem/mingw32noMsys.bat
+++ b/filesystem/mingw32noMsys.bat
@@ -1,0 +1,2 @@
+@echo off
+msys2_shell.cmd -mingw32 -defterm -shell cmd
--- a/filesystem/mingw64noMsys.bat
+++ b/filesystem/mingw64noMsys.bat
@@ -1,0 +1,2 @@
+@echo off
+msys2_shell.cmd -mingw64 -defterm -shell cmd
--- a/filesystem/msys2_shell.cmd
+++ b/filesystem/msys2_shell.cmd
@@ -145,7 +145,7 @@ set MSYSCON=
 if /I "%LOGINSHELL%"=="cmd" (
   set SHELLPATH=%LOGINSHELL%
   if exist "%~dp0%MSYSTEM%\bin\gcc.exe" (
-    set Path=%~dp0%MSYSTEM%\bin;%Path%
+    set Path=%~dp0%MSYSTEM%\bin;%~dp0%usr\bin;%Path%
   )
 ) else (
   set SHELLPATH=%WD%\%LOGINSHELL%
--- a/filesystem/msys2_shell.cmd
+++ b/filesystem/msys2_shell.cmd
@@ -144,6 +144,9 @@ exit /b %ERRORLEVEL%
 set MSYSCON=
 if /I "%LOGINSHELL%"=="cmd" (
   set SHELLPATH=%LOGINSHELL%
+  if exist "%~dp0%MSYSTEM%\bin\gcc.exe" (
+    set Path=%~dp0%MSYSTEM%\bin;%Path%
+  )
 ) else (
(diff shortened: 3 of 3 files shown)


lazka:
hm.. not sure I follow. Why would the terminal you use to invoke "make" change anything?

jeremyd2019:
and how does this differ from `-defterm`?

pchemguy:
> hm.. not sure I follow. Why would the terminal you use to invoke "make" change anything?

If you open the link in my initial comment, you will see that cmake can generate  
*makefiles for use with mingw32-make under a Windows command prompt.
Use this generator under a Windows command prompt with MinGW (Minimalist GNU for Windows) in the PATH and using **mingw32-make as the build tool. The generated makefiles use cmd.exe as the shell to launch build rules. They are not compatible with MSYS or a unix shell.***
While cmake can also make Msys compatible make files, project, which do not rely on Msys-MinGW, but on plain MinGW and cmd.exe as a shell, will fail to build under Msys.

> and how does this differ from -defterm?
Not by much, except that it is only possible to specify the shell name, and msys2_shell.cmd forces the login shell to be "%WD%\%LOGINSHELL%". It is impossible to start it with cmd.exe shell, required by make files designed for  mingw32-make.

lazka:
> The generated makefiles use cmd.exe as the shell to launch build rules. They are not compatible with MSYS or a unix shell.

yes, but that doesn't mean you have to use cmd.exe to launch it.

mingwandroid:
Using mingw32-make its a terrible idea imho. Buggy, incapable of parrallel jobs without falling over (if at all).

That advice is just bad overall..

pchemguy:
> > The generated makefiles use cmd.exe as the shell to launch build rules. They are not compatible with MSYS or a unix shell.
> 
> yes, but that doesn't mean you have to use cmd.exe to launch it.

I feel this discussion is a bit abstract. Let me give an example. Here is a sample repo https://github.com/TransmissionZero/MinGW-DLL-Example.
It builds just fine with mingw32-make command as is supposed to under plain MinGW from cmd.exe shell (when mingw is installed independently or from Msys/MinGW using the proposed change in msys_shell.cmd. It fails to build under `msys2_shell.cmd -mingw32 -defterm` shell. It also fails to build, if I start cmd.exe shell from bash shell first. I suppose it might be possible to come up with a way to get it to work somehow, but I should not have to!

pchemguy:
> Using mingw32-make its a terrible idea imho. Buggy, incapable of parallel jobs without falling over (if at all).
> 
> That advice is just bad overall..

I am not advocating for it. But there are existing projects that use it already, and providing appropriate environment (MinGW/cmd.exe) is not a big deal.

lazka:
> I feel this discussion is a bit abstract. Let me give an example. Here is a sample repo https://github.com/TransmissionZero/MinGW-DLL-Example.

thanks

tanzislam:
For using mingw64's GCC in a plain Windows Command Prompt, I just add `C:\msys64\mingw64\bin` to the system PATH.

stahta01:
You do know about this make option, right?

`mingw32-make SHELL=cmd.exe`

Tim S.

pchemguy:
Probably, I did not

GitMensch:
So is this to be closed ("pass `make` the appropriate `SHELL=cmd.exe` if necessary= or should it be rebased and be integrated?

goyalyashpal:
@pchemguy ping

pchemguy:
I haven't actively used msys2 for a while, and I wouldn't be able to keep discussing this issue.

PR #2422, opened 2021-04-01
cmake: Simplify MSYS support patch

### LINKED ISSUE #2419: CMake: A suggestion for minimizing the support patch
The CMake package contains a [very large patch file](https://github.com/msys2/MSYS2-packages/blob/master/cmake/0001-Support-MSYS.patch) for supporting MSYS runtime environment.

I have a suggestion for minimizing this patch. The approach is similar to what is done on gcc.

I started an attempt to [upstream this patch](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5933), but since it is so large, I was asked to split it. When I reviewed the patch content, I've noticed that most of it can be removed.

The current patch defines `MSYS` only. Then, on each `CYGWIN` condition you have to add `MSYS` as well.

On gcc, the compiler defines *both* `__CYGWIN__` and `__MSYS__` macros. This way, existing code with `__CYGWIN__` conditionals works out of the box, and MSYS-specific code can use `__MSYS__`.

I suggest to do the same for CMake - define *both* `CYGWIN` and `MSYS`, then leave almost all the existing conditions untouched.

Pros:
* Easy to maintain
* Should be fairly easy to upstream
* New conditions in CMake will just work, without the need to adjust them
* The only important difference I found between these platforms are in `MSYS.cmake` itself - library prefix (`cyg` vs `msys-`)

Cons:
* Need to carefully review each existing condition, and see if it requires special handling for MSYS.
* Might break existing packages in msys2, if for example they have something like:
```cmake
if(CYGWIN)
# ...
elseif(MSYS)
# ...
endif()
```

What do you think?

orgads:
Instead of setting only MSYS, set both MSYS and CYGWIN. This enables
using existing CYGWIN conditions without adapting them. It also
makes the patch maintenance much easier.

Fixes #2419.


orgads:
There is a permission test that fails because MSYS uses `noacl` in fstab by default.

```
$ ctest -R 'kwsys.testSystemTools' -V
...
test 5
    Start 5: kwsys.testSystemTools

5: Test command: /f/Projects/build-cmake-msys/Source/kwsys/cmsysTestsCxx "testSystemTools"
5: Test timeout computed to be: 1500
5: SetPermissions failed to set permissions (2) for: /f/Projects/build-cmake-msys/Source/kwsys/testSystemToolsNewDir/testNewFile.txt: actual = 32768; expected = 511
5: Failed to write file: file_b
5: Failed to write file: file_b
1/1 Test #5: kwsys.testSystemTools ............***Failed    0.07 sec
```

I tried to change `fullMask` to `S_IWRITE`, or to 0444, but it still didn't pass. Ideas are welcome (this should not hold back the patch anyway, because it probably fails with the existing package as well).

lazka:
Thanks!

piotrvv:
This link may be the answer to the question why ctest does not work http://cygwin.com/ml/cygwin/2013-11/msg00018.html. The symlink shortcut to the /bin /lib /share directories does not work properly in the cchache. You do not need it in cygwin because there is a virtual directory /usr mounted in binary mode. So we need a better version of gcc with the fortran -static option working. And parsing in grep sed.

piotrvv:
The following tests FAILED:
          5 - kwsys.testSystemTools (Failed)
         26 - FindPackageTest (Failed)
         82 - Module.CheckIPOSupported-C (Failed)
         83 - Module.CheckIPOSupported-CXX (Failed)
        130 - CPackComponents (Failed)
        136 - CPackComponentsPrefix (Failed)
        137 - X11 (Failed)
        145 - ExternalProject (Failed)
        203 - CTest.UpdateSVN (Failed)
        270 - Fortran (Failed)
        272 - FortranC.Flags (Failed)
        353 - RunCMake.Make (Failed)
        370 - RunCMake.FileAPI (Failed)
        374 - RunCMake.GenerateExportHeader (Failed)
        381 - RunCMake.GenEx-TARGET_FILE (Failed)
        415 - RunCMake.MaxRecursionDepth (Failed)
        442 - RunCMake.file (Failed)
        443 - RunCMake.file-CHMOD (Failed)
        444 - RunCMake.find_file (Failed)
        445 - RunCMake.find_library (Failed)
        446 - RunCMake.find_package (Failed)
        447 - RunCMake.find_path (Failed)
        463 - RunCMake.project_injected (Failed)
        480 - RunCMake.configure_file (Failed)
        485 - RunCMake.SymlinkTrees (Failed)
        508 - RunCMake.CommandLine (Failed)
        510 - RunCMake.install (Failed)
        515 - RunCMake.ExternalProject (Failed)
        519 - RunCMake.CPackSymlinks (Failed)
        523 - RunCMake.ClangTidy (Failed)
        524 - RunCMake.IncludeWhatYouUse (Failed)
        526 - RunCMake.Cppcheck (Failed)
        527 - RunCMake.MultiLint (Failed)
        533 - RunCMake.CPack_TGZ (Failed)
        553 - CMake.GetFilenameComponentRealpath (Failed)
        557 - CMake.ImplicitLinkInfo (Failed)
        558 - CMake.ModuleNotices (Failed)
Errors while running CTest
make: *** [Makefile:148: test] Error 8

orgads:
I now found that `chmod` doesn't work when `umask` is set. That's very strange...

It's the same on Cygwin when `noacl` is set.

Example:
```c++
#include <unistd.h>
#include <sys/stat.h>
#include <stdlib.h>
#include <fcntl.h>
#include <iostream>

const char fileName[] = "foo";
const int mask = S_IWRITE;

void get()
{
  struct stat st;
  stat(fileName, &st);
  std::cout << (st.st_mode) << std::endl;
}

void set(int mode) {
  chmod(fileName, mode);
}

int main() {
  std::cout << std::oct;
  close(open(fileName, O_WRONLY | O_CREAT));
  get();
  umask(mask);
  set(0);
  get();
  set(mask);
  get();
}
```

orgads:
Reported to cygwin: https://cygwin.com/pipermail/cygwin/2021-April/248208.html

orgads:
I found the root caused. Reported here: https://cygwin.com/pipermail/cygwin/2021-April/248241.html

The problem is that when `umask` is set, `stat` applies it on its result. So it's not `chmod` that is broken, but `stat`. Worked around it in the patch (moved the `umask` call to where it's needed).

orgads:
Do you prefer to rename the patch file back to `0001-Support-MSYS.patch`, so you can review the differences?

orgads:
Looks okish now. Some tests fail, but they fail also on cygwin. See my last comment on https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5969.

lazka:
I've built brotli and clang with this and all seems fine.

Thanks!

orgads:
That was a long journey \o/

Thanks for the careful review.

lazka:
It hasn't been _that_ careful ;) as my cmake knowledge is limited, but you seem to know your way around and I agree with the patch goal.

PR #2425, opened 2021-04-07
Update mingw-w64 packages to commit 1c7738.

jeremyd2019:
This should include fixes for msys2/CLANG-packages#6 and
msys2/CLANG-packages#21

### DIFF
--- a/mingw-w64-cross-clang-crt/PKGBUILD
+++ b/mingw-w64-cross-clang-crt/PKGBUILD
@@ -4,7 +4,7 @@ _realname=crt
 _mingw_suff=mingw-w64-cross-clang
 pkgname=("${_mingw_suff}-${_realname}")
 conflicts=("${_mingw_suff%-*}-${_realname}")
-pkgver=9.0.0.6128.07922837
+pkgver=9.0.0.6158.1c773877
 pkgrel=1
 pkgdesc='MinGW-w64 CRT for cross-compiler'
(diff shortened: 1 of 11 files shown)


jeremyd2019:
CI failed due to a blip cloning from sf's git server.  But, CI fails anyway due to conflicts between the clang, ucrt, and 'normal' versions of the mingw-w64-cross packages, because the CI script installs all packages after building.  If desired, I can split this into 3 pull requests that can each pass CI.

Biswa96:
Passing CI here would not make any difference. The can be merged as is because mingw-w64 toolchain will not fail to build 🤞

jeremyd2019:
Well, I've been doing some building with this revision, and ran into a regression in ucrt: https://sourceforge.net/p/mingw-w64/mailman/message/37257645/

Namely, fseterr.c from gnulib now fails to compile, because it tries to mess with the _flag field in the FILE*

jeremyd2019:
Issue I ran into is now fixed.  PR updated to newer commit that fixes it.

jeremyd2019:
Once this is merged (or at least the mingw-w64-cross-clang packages, I can split them if that's more palatable) I intend to update/rebase #2294, and once that and #2426 are merged I can at least get msys2/CLANG-packages#2 building an aarch64 hosted and targeting clang in CI.

lazka:
We need to move CI to use a temporary repository instead of installing everything in the PR.

PR #2426, opened 2021-04-08
add CLANGARM64 MSYSTEM/MINGW_ARCH

jeremyd2019:
This may be a bit premature (there are no packages yet), but having support in the repo would help with automation.


pacman/makepkg_mingw.conf
```diff
@@ -80,6 +80,16 @@ elif [[ "$MSYSTEM" == "CLANG32" ]]; then
...
   LDFLAGS="-pipe -Wl,--dynamicbase,--no-seh"
+elif [[ "$MSYSTEM" == "CLANGARM64" ]]; then
+  CARCH="aarch64"
+  CHOST="aarch64-w64-mingw32"
+  MINGW_CHOST="aarch64-w64-mingw32"
+  MINGW_PREFIX="/clangarm64"
+  MINGW_PACKAGE_PREFIX="mingw-w64-clang-aarch64"
+  CPPFLAGS="-D__USE_MINGW_ANSI_STDIO=1"
+  CFLAGS="-march=aarch64 -mtune=generic -O2 -pipe"
+  CXXFLAGS="-march=aarch64 -mtune=generic -O2 -pipe"
```
Biswa96, line 91: Shouldn't this be `-march=armv8-a`?

jeremyd2019: Possibly.  I just double-checked, and I have actually been running with no -march or -mtune.

PR #2431, opened 2021-04-11
libtool: support `llvm-objdump -f` output

mati865:
More info in https://github.com/msys2/CLANG-packages/issues/9

### DIFF
--- a/libtool/0014-Support-llvm-objdump-f-output.patch
+++ b/libtool/0014-Support-llvm-objdump-f-output.patch
@@ -1,0 +1,39 @@
+From 1a4e74ef284f0944f474df5c28afd6c1737f66b6 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Mateusz=20Miku=C5=82a?= <mati865@gmail.com>
+Date: Sat, 10 Apr 2021 00:39:05 +0200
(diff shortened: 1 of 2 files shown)


lazka:
adding a validpgpkeys array should fix that

mati865:
Tests will fail if they are ran on the CI though and I couldn't figure what is wrong with them.

mati865:
Oh, `check()` is not ran on the CI.

jeremyd2019:
I don't know why, but gmp seems to be trying to use a linker script now on i686.

jeremyd2019:
It kind of looks like the ltmain.sh installed in the package is lacking some of the other clang-related patches that were already there

jeremyd2019:
Ah, the new patch patches ltmain.sh, making it newer than ltmain.in which the other patches patched, causing ltmain.sh not to be regenerated from ltmain.in.

mati865:
Okay, now I hate autotools even more than usually. So reordering the patches would help?

jeremyd2019:
Just patch ltmain.in instead of ltmain.sh

mati865:
Ah, I have modified `.sh` instead of `.in`. Nice catch.

PR #2434, opened 2021-04-12
Update packages

Biswa96:


Biswa96:
GitHub Actions can not download source from invisible-mirror.net of diffstat. Need manual build.

mksh/PKGBUILD
```diff
@@ -1,10 +1,10 @@
 # Maintainer: Masanori Shirayama <mashir43@gmail.com>
 
 pkgname=mksh
-_ver=59
+_ver=59c
```
lazka, line 4: needs "59.c" since "59" > "59c"

Biswa96: In next line, `_ver` is converted to correct `pkgver`.

Biswa96:
Ping!

mksh/PKGBUILD
lazka: oops

lazka:
Pong!

PR #2437, opened 2021-04-12
jo: add new package

1480c1:


jo/PKGBUILD
```diff
@@ -0,0 +1,34 @@
...
+
+build() {
+  rm -rf "${srcdir}/build-${CARCH}"
+  mkdir -p "${srcdir}/build-${CARCH}"
+  cd "${srcdir}/build-${CARCH}"  || return 1
+
+  ../${pkgname}-${pkgver}/configure \
+    --prefix=/usr \
+    --build="${CARCH}" \
+    --host="${CARCH}"
```
lazka, line 21: `${CHOST}`?

1480c1: Accidentally replaced that with the wrong variable

lazka:
Why do you need a msys version of this as well?

1480c1:
Was mainly doing this for the bash-completion, but I do notice that jq doesn't have a msys2 version, so probably actually unnecessary

lazka:
I've created #2438 for bash-completion

PR #2447, opened 2021-04-17
Canadian Mirror added

ZachBacon:
The ucrt64 and x86_64 repo is only supported in my repo for now due to space constraints at the moment, however, I'll solve that in a few weeks time. Also at the time of creating this PR, the x86_64 repo is still syncing with rsync, I'll make a comment on this pr when it's done.

### DIFF
--- a/pacman-mirrors/PKGBUILD
+++ b/pacman-mirrors/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=pacman-mirrors
 pkgver=20210417
-pkgrel=2
+pkgrel=1
 pkgdesc="MSYS2 mirror list for use by pacman"
 arch=('any')
 url="https://www.msys2.org/dev/mirrors/"
--- a/pacman-mirrors/PKGBUILD
+++ b/pacman-mirrors/PKGBUILD
@@ -1,8 +1,8 @@
 # Maintainer: Ray Donnelly <mingwandroid@gmail.com>
 
 pkgname=pacman-mirrors
-pkgver=20210325
-pkgrel=1
+pkgver=20210417
+pkgrel=2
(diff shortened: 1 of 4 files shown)


Biswa96:
Where are the other repos?

ZachBacon:
Again due to disk space I'm only dealing with the two for now, once I can allocate more disk space I'll be adding the rest.

Biswa96:
If this is allowed _by the law of msys2_ you need to update the checksums by running `updpkgsums` and change the `pkgver` to current date format.

ZachBacon:
Call me ignorant, but that has to do with a mirror setup how?

lazka:
The files you changed are part oft the package pacman-mirrors, without correct checksums it will fail to build and without a pkgver bump no one will get the update.

ZachBacon:
And that's what I get for trying to do things super late at night

ZachBacon:
sync has completed, and I'll maintain the sync at least twice daily

Biswa96:
As `pkgver` is increased `pkgrel` need to be reset to 1.

PKGBUILD 101 ;)

ZachBacon:
> As `pkgver` is increased `pkgrel` need to be reset to 1.
> 
> PKGBUILD 101 ;)

You'll have to forgive me, I mainly do packaging for a different set of rules, Solus has a thing that regardless of a new version the rel is bumped regardless. My kinda defaulted to that without realizing. I'll fix that when I'm better awake and can actually care better

1480c1:
> As `pkgver` is increased `pkgrel` need to be reset to 1.
> 
> PKGBUILD 101 ;)

Arrays start at 0! /s

PR #2449, opened 2021-04-18
filesystem, gnupg, msys2-keyring: avoid keeping gpg-agent/dirmngr around

lazka:
They get started in the background and stay around which comes with various problems:
* The files can't be deleted until they are stopped
* Some CI systems wait until all programs are stopped
* On a cygwin update we have to restart everything or forking will be broken
  (though pacman takes care of that now by killing everything cygwin-like on a core update)

Ideally though they would just exit in a controlled way after we no longer need them.

The reason this uses GNUPGHOME instead of passing --home to gpgconf is that dirmngr started by
"pacman-key --refresh-keys" somehow couldn't be killed that way.
Maybe because gpg doesn't forward --homedir in that case. Setting GNUPGHOME helped.

### DIFF
--- a/filesystem/07-pacman-key.post
+++ b/filesystem/07-pacman-key.post
@@ -2,9 +2,11 @@ maybe_init_keyring ()
 {
   if [ ! -d /etc/pacman.d/gnupg ]
   then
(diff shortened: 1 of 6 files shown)


lazka:
Let's give this a try...

PR #2455, opened 2021-04-23
filesystem: set XDG_DATA_DIRS for mingw bash-completion

### LINKED ISSUE #2438: [bash-completion] make it work with mingw packages
The following makes it work, maybe something we can set on start:

```
XDG_DATA_DIRS=$MINGW_PREFIX/share:/usr/local/share/:/usr/share/ bash -l
```

lazka:
This adds a new profile.d/ script where we can add stuff without changing
/etc/profile and triggering a pacman backup.

Fixes #2438

### DIFF
--- a/filesystem/profile.000-msys2.sh
+++ b/filesystem/profile.000-msys2.sh
@@ -1,0 +1,6 @@
+# Setting XDG_DATA_DIRS allows bash-completion to find completion files
+# installed in the mingw package prefixes as well
+if [ ! "${MINGW_PREFIX}" = "" ]; then
+    XDG_DATA_DIRS="/usr/local/share/:/usr/share/"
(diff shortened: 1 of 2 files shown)


filesystem/profile.000-msys2.sh
```diff
@@ -0,0 +1,6 @@
+# Setting XDG_DATA_DIRS allows bash-completion to find completion files
+# installed in the mingw package prefixes as well
+if [ ! "${MINGW_PREFIX}" = "" ]; then
+    XDG_DATA_DIRS="/usr/local/share/:/usr/share/"
+    export XDG_DATA_DIRS="$MINGW_PREFIX/share:$XDG_DATA_DIRS"
+fi
```
naveen521kk, line 6: Add a new line at the end of the file
```suggestion
fi

```

PR #2456, opened 2021-04-23
pacman: require databases to be signed

### LINKED ISSUE #2453: [pacman] Remove SigLevel=DatabaseOptional
We copied this (I guess) from Arch Linux which doesn't sign databases.

But we do sign them, so we shouldn't allow DBs without a signature.

In theory this (as it is now) would allow a mirror/server attacker to modify the DB to provide old packages without security fixes (the package signatures still have to work out..)

open question: Do we need to adjust the staging repo example docs if we change this?

lazka:
We used the defaults from Arch, but Arch doesn't sign its databases.
We do sign them, so there is no reason to not enforce this.

Fixes #2453

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=pacman
 pkgver=5.2.2
-pkgrel=20
+pkgrel=21
 pkgdesc="A library-based package manager with dependency support (MSYS2 port)"
 arch=('i686' 'x86_64')
 url="https://www.archlinux.org/pacman/"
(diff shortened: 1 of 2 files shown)


elieux (approved):
We do sign both packages and databases in the repo maintaining scripts.  I have been running like that for years now and I don't remember any issues specific to this change.  In conclusion, I don't see anyone negatively affected by this and it does provide some additional security.  Those who have added their own or third-party unsigned repositories presumably know how to change the settings (and the configuration is not even overwritten automatically).

PR #2461, opened 2021-04-24
pacman-mirrors: Backup mirrorlist.*

naveen521kk:
They just get overwritten but they shouldn't.
https://wiki.archlinux.org/index.php/PKGBUILD#backup

### DIFF
--- a/pacman-mirrors/PKGBUILD
+++ b/pacman-mirrors/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=pacman-mirrors
 pkgver=20210423
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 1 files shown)


lazka:
@elieux thoughts on this?

naveen521kk:
repo.msys2.org sometimes doesn't work(it's sometimes very slow or blocks my connection) for me and I use another mirror, and it isn't nice to change it every time.

lazka:
yeah, I agree

jeremyd2019:
What I have been doing is adding a `Server =` line before the `Include =` line in pacman.conf, which was already protected in this way.

PR #2463, opened 2021-04-25
ca-certificates: Update to 20210119

lazka:


Biswa96:
Should this be updated? The "work" folder names feels like unstable.

lazka:
The source comes from here: https://packages.debian.org/source/bullseye/ca-certificates imo this is fine

PR #2466, opened 2021-04-25
Update some PKGBUILD config backups

cbrt64:
Self-explanatory, I hope. If there are others in `/etc` needing backup I couldn't find them.

### DIFF
--- a/msys2-launcher/PKGBUILD
+++ b/msys2-launcher/PKGBUILD
@@ -3,11 +3,11 @@
 _realname="msys2-launcher"
 pkgname=("${_realname}")
 pkgver=1.3
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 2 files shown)


filesystem/PKGBUILD
```diff
@@ -14,7 +14,7 @@ provides=('msys2-base')
 replaces=('msys2-base')
 backup=('etc/fstab' 'etc/shells' 'etc/profile'
         etc/bash.bash{rc,_logout} etc/skel/.bash{rc,_profile,_logout}
-        etc/nsswitch.conf)
+        etc/nsswitch.conf msys2_shell.cmd)
```
lazka, line 17: This isn't a config file but a program, so I'd rather not see it backuped.

lazka: thanks

PR #2468, opened 2021-04-26
clang: clean up msysize patches

lazka:
cmake defines CYGWIN now and in C we have `__CYGWIN__` too


lazka:
@mati865 in case you are interested. Ideally this shouldn't change anything and just reduce patches

mati865 (approved):
Nice

PR #2473, opened 2021-04-29
msys2-runtime: Backport fix for symlink resolving

orgads:
Reported here: https://cygwin.com/pipermail/cygwin/2021-April/248307.html


lazka:
Thanks, msys2-runtime is managed here: https://github.com/msys2/msys2-runtime/pulls

After it gets merged there https://github.com/msys2/MSYS2-packages/blob/master/msys2-runtime/update-patches.sh takes care of syncing all the patches.

lazka:
Somewhat related, I wonder when we can expect v3.2.1...

orgads:
Done

PR #2477, opened 2021-05-04
gdb: fix i686 build

jeremyd2019:
the msysize patch had a hunk that was missed that needed to be updated for i686.

### DIFF
--- a/gdb/PKGBUILD
+++ b/gdb/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=gdb
 pkgver=10.2
-pkgrel=1
+pkgrel=2
 _gcc_ver=10.2.0
 pkgdesc="GNU Debugger (MSYS2 version)"
 arch=('i686' 'x86_64')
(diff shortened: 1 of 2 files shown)


gdb/PKGBUILD
```diff
@@ -2,7 +2,7 @@
 
 pkgname=gdb
 pkgver=10.2
-pkgrel=1
+pkgrel=2
 _gcc_ver=10.2.0
```
mati865, line 6: While bumping pkgrel could you also fix `_gcc_ver`?
```suggestion
_gcc_ver=10.3.0
```

jeremyd2019: MSYS2 gdb is still 10.2.0

mati865: Sorry, eyeballed wrong package.

jeremyd2019: I checked and mingw-w64-gdb does not seem to contain corresponding variable/file.  So, is that supposed to be mingw-w64-gcc version or msys2 gcc version?

mati865: It's correct right now. Github doesn't let me resolve this comment for some reason.

PR #2478, opened 2021-05-05
mm-common: new package in support of gnome mm.

jeremyd2019:
This includes autoconf m4 macros necessary to autoreconf gnome
<whatever>mm projects.  Since autoconf is msys2, this is too.

PKGBUILD largely lifted [from arch](https://github.com/archlinux/svntogit-packages/blob/13ac84ebfee343a2a452207d6c029e165676e972/trunk/PKGBUILD).

### DIFF
--- a/mm-common/PKGBUILD
+++ b/mm-common/PKGBUILD
@@ -1,4 +1,4 @@
-# Contributor: Ionut Biru <ibiru@archlinux.org>
+# Maintainer: Jeremy Drake <github@jdrake.com>
 
 pkgname=mm-common
 pkgver=1.0.2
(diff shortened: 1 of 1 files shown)


jeremyd2019:
It builds, but I have not tested it yet

Biswa96:
AFAIK, mm-common is not required when maintainer mode is disabled, see cairomm16.

jeremyd2019:
yes, however, cairomm is old enough to apparently not have meson build, and autotools needs an autoreconf to get the new libtool fixes for clang.  this package contains m4 macros necessary to complete autoreconf.

jeremyd2019:
Confirmed that this package allows autoreconf of mingw-w64-cairomm to succeed, which allows it to build under clang64

Biswa96:
BTW, would you like to add yourself as maintainer in PKGBUILD?

PR #2483, opened 2021-05-07
spdlog: add

neheb:
Signed-off-by: Rosen Penev <rosenp@gmail.com>


Biswa96:
Why? We want to keep less packages in msys2 as we focus on mingw packages.

neheb:
Porting a Linux package to Windows. MinGW seems too hard (lots of POSIX APIs used).

Biswa96:
spdlog is already available in mingw packages https://packages.msys2.org/base/mingw-w64-spdlog

lazka:
We only add packages here that we need for building mingw ones.

neheb:
OK

Issue #2487, opened 2021-05-08
python: broken with child_info_fork::abort: address space needed


lazka:
* Rebuild Python as is
* Build glib2 (which uses meson and thus Python)

```
2021-05-08T21:44:59.3747015Z ==> Starting build()...
0 [main] python 22250 child_info_fork::abort: address space needed by '_ssl.cpython-38-x86_64-msys.dll' (0x190000) is already occupied
0 [main] python 22251 child_info_fork::abort: address space needed by '_ssl.cpython-38-x86_64-msys.dll' (0x190000) is already occupied
0 [main] python 22252 child_info_fork::abort: address space needed by '_ssl.cpython-38-x86_64-msys.dll' (0x190000) is already occupied
```

CI run: https://github.com/lazka/MSYS2-packages/runs/2536195524

jeremyd2019:
This looks like the algorithm that Windows uses when the address space that an image wants to occupy is already occupied, it will start filling them in at fairly low addresses, but not in a way that necessarily loads them at the same place in each process.  So I would look for any DLLs that would overlap imagebase + length.

It's odd that the compile of .py modules at the end of python build didn't fail, that's where I had a lot of issues with i686.  That involves lots of forking from multiprocessing module, and loads all the extension dlls.

jeremyd2019:
(I wish the "MSYS2 DLL dependencies"/ldd listing would include the image base of the module itself in addition to those of its dependencies... maybe even the lengths too)

jeremyd2019:
Here's yer overlap:
```
  /c/_/python/pkg/python/usr/lib/python3.8/lib-dynload/_ssl.cpython-38-x86_64-msys.dll                base 0x0005d76d0000 size 0x00029000 *
  /c/_/python/pkg/python/usr/lib/python3.8/lib-dynload/_struct.cpython-38-x86_64-msys.dll             base 0x0005d76e0000 size 0x0000f000 *
```

lazka:
Previously:
```
/usr/lib/python3.8/lib-dynload/_ssl.cpython-38-x86_64-msys.dll base 0x0004210a0000 size 0x00029000
/usr/lib/python3.8/lib-dynload/_struct.cpython-38-x86_64-msys.dll base 0x0004cef70000 size 0x0000f000
```

jeremyd2019:
[...]

PR #2488, opened 2021-05-09
ci: list dll bases and lengths

jeremyd2019:
By calling `rebase -i`.  Trying to help diagnose #2487.

### DIFF
--- a/.ci/ci-build.sh
+++ b/.ci/ci-build.sh
@@ -55,6 +55,7 @@ for package in "${packages[@]}"; do
 
     echo "::group::[dll check] ${package}"
     execute 'Checking dll depencencies' list_dll_deps ./pkg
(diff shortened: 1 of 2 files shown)


jeremyd2019:
Output for python: https://github.com/jeremyd2019/MSYS2-packages/runs/2536413793?check_suite_focus=true#step:7:10798
I am not sure about the grep coloring, it doesn't do much for python cause most of its dlls don't start with `msys-`

jeremyd2019:
It might make sense to highlight lines ending with `*` because those indicate an overlap.  But exes are expected to overlap, because they aren't typically loaded into the same address space so they aren't generally subject to the `--enable-auto-image-base` hash-based image bases.

lazka:
Thanks!

PR #2494, opened 2021-05-10
binutils: increase the address range for auto-image-base

lazka:
We get conflicts in Python modules now. Their paths contain the cygwin
version and that happened to trigger a base image collision when being hashed
since the last update.

This change uses a bit more of the path based hash to increase the possible base addresses
and make collisions less likely.

Hopefully without any negative side effects.


jeremyd2019:
With this change, the range for auto image bases increases from `0x0400000000-0x05ffff0000` to `0x0400000000-0x13ffff0000`.  I don't know if that's a problem or not, but just pointing it out.  If the mask were `0x7ffff0000` the range would go to `0x0bffff0000` which doesn't spill into the next hex digit.

lazka:
Looks good. Fingers crossed.

PR #2495, opened 2021-05-10
rhash: disable --dynamicbase etc.

lazka:
This can't be good for cygwin magic, so disable.


jeremyd2019:
Actually it's *usually* fine, but once in a blue moon 💣

lazka:
Yeah, reading up on all this it seems Windows will try to keep the same "random" base to share memory between processes ... usually.

PR #2496, opened 2021-05-11
python: pretend to be cygwin in more places

lazka:
The main change here is that sys.platform == 'cygwin', which allows
us to drop various changes for extending cygwin checks. Fewer patches
and less likely that we miss to patch a new cygwin check on updates.

If one really needs to check for msys Python then this still works:

sysconfig.get_platform().startswith("msys")


lazka:
This will require a review of patches for other packages which use sys.platform.

PR #2497, opened 2021-05-13
filesystem: remove empty mingw root directories

lazka:
I don't think they are actually needed anymore


Alexpux:
Why you think so? This folders are creating and handling by this package

lazka:
pacman will just create them when needed, no? Like for everything else.

lazka:
I'm going to revert this, but for other reasons -> #2502

pacman gets confused by the direcotry being named like the executable. While this doesn't prevent the upgrade, only the downgrade I'd like to check that out first.

PR #2501, opened 2021-05-14
pacman: export CC and CXX variables explicitly

MehdiChinoune:


jeremyd2019:
I thought of this, but didn't know what if anything it might break.  I do know that python appreciates knowing that the compiler is named clang, and turns on proper support for clang pgo/lto, so I set those vars explicitly in python's PKGBUILD

Biswa96:
This seems to be same change as the other PR in mingw-packages. What happen if some projects use clang even in mingw64 or mingw32? e.g. zig, ghld etc. Why do we need this change?

MehdiChinoune:
> This seems to be same change as the other PR in mingw-packages. What happen if some projects use clang even in mingw64 or mingw32? e.g. zig, ghld etc. Why do we need this change?

If they need clang in mingw64 or gcc in clang64, they have to explicitly export CC and CXX in the PKGBUILD file.
We need this change to make buildsystems know that they are using clang instead of gcc, as some of them deduce the compiler from its name.

Alexpux:
Setting this variables in conf files will not export them to build. You must also export this variables in /usr/share/makepkg/buildenv/buildflags.sh

MehdiChinoune:
The CI failure is not related to my changes.

jeremyd2019:
> The CI failure is not related to my changes.

But that's not good (#2487)

PR #2509, opened 2021-05-18
rebase: update to 4.5.0

jeremyd2019:
- Drop some upstream patches.
- Rebase some other patches.
- Use new --with-posix-shell configure option rather than patching shebang on scripts.
- Includes new pacman hook for i686 (#2108).


jeremyd2019:
Looks like GHA is screwed up again.  So my plan to use it to test is on hold...

jeremyd2019:
Simple tests on i686 and x86_64 worked as expected

PR #2512, opened 2021-05-21
pacman: make makepkg handle CMAKE_BUILD_PARALLEL_LEVEL

jeremyd2019:
As discussed on [discord recently](https://discord.com/channels/792780131906617355/794220101741838401/844763658637672478).

Add `CMAKE_BUILD_PARALLEL_LEVEL` to `makepkg{,_mingw}.conf`.  Teach makepkg to export it, and unset it when `!makeflags` option is used.

This came about because I am concerned the forthcoming switch of mingw-w64-clang to using ninja will make building on my raspberry pi too swappy to be useful, as the pi 3B+ has 4 cores but only 1GB of RAM, and ninja will default to `-j6` [in that case](https://github.com/ninja-build/ninja/blob/0a632d11eb1e8b7becbd9756c3e3439777924f6d/src/ninja.cc#L246).

### DIFF
--- a/pacman/0021-export-build-vars.patch
+++ b/pacman/0021-export-build-vars.patch
@@ -5,7 +5,7 @@
  
  	# ensure all necessary build variables are exported
 -	export CPPFLAGS CFLAGS CXXFLAGS LDFLAGS RUSTFLAGS MAKEFLAGS CHOST
-+	export CC CXX CPPFLAGS CFLAGS CXXFLAGS LDFLAGS RUSTFLAGS MAKEFLAGS CHOST
++	export CC CXX CPPFLAGS CFLAGS CXXFLAGS LDFLAGS RUSTFLAGS MAKEFLAGS CMAKE_BUILD_PARALLEL_LEVEL CHOST
  }
 --- pacman-5.2.2/scripts/libmakepkg/buildenv/buildflags.sh.in.orig	2021-05-15 07:55:29.592543800 +0100
 +++ pacman-5.2.2/scripts/libmakepkg/buildenv/buildflags.sh.in	2021-05-15 07:55:42.773495200 +0100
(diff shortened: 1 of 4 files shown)


jeremyd2019:
I won't be offended if this isn't desired.  I *think* I could just export it from the shell calling makepkg without this change.

Also, I wonder if upstream might be interested in this, as they are probably also moving into a world where make is used less so the `MAKEFLAGS` option is less relevant

Biswa96:
Should this be added to pacman? This looks like just a personal choice/configuration. Just asking 😁

jeremyd2019:
That's kind of what I'm asking too :grin:.  But, the argument for is that this in fact makes it a configuration, on par with `MAKEFLAGS` which is an analogous option that already exists.

jeremyd2019:
Adding this to my `makepkg_mingw.conf` seems to work:
```sh
CMAKE_BUILD_PARALLEL_LEVEL="2"
export CMAKE_BUILD_PARALLEL_LEVEL
```
Doesn't do anything nice like getting disabled when `!makeflags` option is specified, but does the trick without having to patch anything.

mati865:
You can add it to https://github.com/msys2/msys2-pacman/blob/9df3eaa6f16a67e2c564eaed85bcd2f7ecb6c80f/scripts/libmakepkg/buildenv/makeflags.sh.in#L33 so it will get disabled by `!makeflags`.

jeremyd2019:
This pull request does that, and also exports it in the first place like MAKEFLAGS.

jeremyd2019:
It doesn't seem like anybody else wants this, and I have a working solution (minus disabling on `!makeflags`, which may not even be desired as usually setting `!makeflags` means the package wants serial build anyway).

PR #2514, opened 2021-05-22
Update packages

Biswa96:


Biswa96:
Same issue with invisible-mirror.net download in CI.

mm-common/PKGBUILD
```diff
@@ -1,26 +1,20 @@
...
 pkgrel=1
 pkgdesc="Common build files of the C++ bindings"
 url="https://www.gtkmm.org/"
 arch=('i686' 'x86_64')
-makedepends=('git' 'meson')
+makedepends=('meson')
 license=('GPL2')
-_commit=85381b86dcc2490d2b080c178477df9eb65103ca  # tags/1.0.2^0
-source=("git+https://gitlab.gnome.org/GNOME/mm-common.git#commit=$_commit"
+source=("https://download.gnome.org/sources/mm-common/${pkgver%.*}/${pkgname}-${pkgver}.tar.xz"
```
jeremyd2019, line 11: Cool, when I did the initial package I didn't find source tarballs (though to be honest once I found the arch package used git, I didn't look too hard)

mm-common/PKGBUILD
```diff
@@ -43,5 +37,3 @@ package() {
   cd "${srcdir}/build-${CARCH}"
   DESTDIR="${pkgdir}" meson install
 }
-
-# vim:set ts=2 sw=2 et:
```
jeremyd2019, line 47: 😆 is there some style guide about vim/emacs modeline comments.  I added one in the split package template, so they are multiplying.

Biswa96: AFAIK, there is no style guidelines in any msys2 repo.

mm-common/PKGBUILD
Biswa96: Arch uses git repos because git is itself a checksum verification engine (kind of). But that comes at a cost, the git repo is cloned with full history.

jeremyd2019:
> Same issue with invisible-mirror.net download in CI.

They haven't fixed that yet? 😠

PR #2531, opened 2021-06-04
etc-update: new package

jeremyd2019:
Gentoo's tool to manage config file updates.  Based on Arch's PKGBUILD.

etc-update looks at `/etc/os-release` to determine how config file updates are handled.  Add this to `filesystem` package, specifying that MSYS2 is like cygwin and arch (being like arch is how it decides to look at `.pacnew` files).

Word to the wise, you probably want to set the `EDITOR` environment variable.   I was testing this from a fresh install from an old installer, and `sdiff` tried to run `ed`.

### DIFF
--- a/etc-update/PKGBUILD
+++ b/etc-update/PKGBUILD
@@ -1,0 +1,26 @@
+# Maintainer: Jeremy Drake <github@jdrake.com>
+pkgname='etc-update'
+pkgdesc="CLI to interactively merge .pacnew configuration files"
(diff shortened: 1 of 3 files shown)


jeremyd2019:
I've also been thinking it might make sense to include "cygwin" in the `ID_LIKE` list.

lazka:
Worth a try I guess. The current tools are not very user friendly.

PR #2533, opened 2021-06-06
m4: Upgrade to 1.4.19

jannick0:
Surprisingly `m4 1.4.19` implies a dependency to `libiconv` which is burried deep in `GNULIB`. Preferably this dependency could be dropped by using `iconv` replacement functions shipped by the `m4` package, not easy to switch on for me. 

Grateful if someone had an idea how to do use the internal `iconv` replacement functions.  Otherwise the package is ready to be merged from my side.


m4/PKGBUILD
```diff
@@ -1,45 +1,77 @@
...
-  cd ${srcdir}/$pkgname-$pkgver
-  ./configure \
-    --prefix=/usr \
-    --sysconfdir=/etc \
-    --localstatedir=/var \
-    --with-syscmd-shell=/usr/bin/sh
-  make
+  mkdir -p build-${pkgname}-${pkgver}_${MSYSTEM_CHOST}
+  pushd build-${pkgname}-${pkgver}_${MSYSTEM_CHOST}
+    # fool configure that gcc is CYGWIN (which it essentially is)
```
lazka, line 35: oh, that's a nice trick

jannick0: Works as long as gcc's triplet is not used in the build process, i.e. in any Makefile.

PR #2540, opened 2021-06-14
pacutils: new package

jeremyd2019:
The checks pass and everything.

concern: this is definitely *not* statically linked.  As a pacman-related package, should it be? (I notice pacman-contrib is)?


jeremyd2019:
Created andrewgregory/pacutils#47 and andrewgregory/pacutils#48

lazka:
> concern: this is definitely _not_ statically linked. As a pacman-related package, should it be? (I notice pacman-contrib is)?

pacman-contrib used to be part of pacman, so I kept it that way. Skipping it here seems fine.

lazka:
Just wondering, what motivated you to package this?

jeremyd2019:
I was looking at `paccheck` - what I read online was that `pacman -Qkk` only checks metadata of installed files (owner, mode, etc) against the database, while paccheck can verify checksums.  But after reading more of the other tools, some of them seem quite nifty.

lazka:
`pacman -Qkk` also does checksums since 6.0 I think, but it seems broken here

PR #2542, opened 2021-06-14
pacman: use color by default

1480c1:


Biswa96:
1. Can space checking also disabled?
2. Can parallel download be enabled?

jeremyd2019:
> 1. Can space checking also disabled?

https://github.com/msys2/setup-msys2/blob/master/main.js#L252 but for end-users it seems like a useful thing.

> 2. Can parallel download be enabled?

It already is.  https://github.com/msys2/MSYS2-packages/blob/master/pacman/pacman.conf#L36

Biswa96:
> but for end-users it seems like a useful thing.

Everybody disable it in CI.

1480c1:
@lazka any opinion on the disabling checkspace? I personally found it a bit annoying due to how long it takes sometimes

lazka:
> > but for end-users it seems like a useful thing.
> 
> Everybody disable it in CI.

because CI systems don't keep state, so a broken installation doesn't matter.

> @lazka any opinion on the disabling checkspace? I personally found it a bit annoying due to how long it takes sometimes

I remember it being widely different on some systems, it's pretty much instant here.

I'm not sure really.. I'll try testing what happens if there isn't enough disk space.

pacman/PKGBUILD
```diff
@@ -4,7 +4,7 @@
 
 pkgname=pacman
 pkgver=6.0.0
-pkgrel=2
+pkgrel=3
```
lazka, line 7: update checksums

1480c1:
We can probably talk about disabling space check in another PR and keep this just to color

PR #2547, opened 2021-06-18
msys2-runtime: handle Ctrl+C with MINGW executables

dscho:
The idea is that we handle <kbd>Ctrl</kbd>+<kbd>C</kbd> not by terminating the processes (which would reflect `kill -KILL`, not `kill -INT`) but by imitating what Windows does in a CMD window: we inject a remote thread that runs the `CtrlRoutine` function. This lets the `atexit()` handlers and the ConsoleCtrlEvent handlers do their job.

This patch simply applies the patches that were merged in https://github.com/msys2/msys2-runtime/pull/31 and fixes https://github.com/msys2/msys2-runtime/issues/29.

Cc: @naveen521kk


lazka:
(I'll keep this in staging for a day or two so it gets some testing)

naveen521kk:
There is no file diff here which mean `getprocaddr*.exe`  isn't packaged. I will fix it.

dscho:
> 
> 
> There is no file diff in CI build which mean `getprocaddr*.exe` isn't packaged. I will fix it.

Oh? I thought we had tested this... But then, that was the CI build of https://github.com/msys2/msys2-runtime/pull/31... @naveen521kk thank you for being so diligent!

naveen521kk:
well technically your thanks should go to @lazka, he was the one to ping me on Discord and ask why things aren't working after installing from the staging repo and only then I found that those executables weren't there.

PR #2550, opened 2021-06-18
Update packages

Biswa96:


gnutls/gnutls-3.6.3-msysize.patch
```diff
@@ -2,7 +2,7 @@ diff -Naur gnutls-3.6.3-orig/build-aux/ar-lib gnutls-3.6.3/build-aux/ar-lib
 --- gnutls-3.6.3-orig/build-aux/ar-lib	2018-07-13 08:14:46.000000000 +0300
 +++ gnutls-3.6.3/build-aux/ar-lib	2018-07-26 13:56:44.494565900 +0300
 @@ -56,6 +56,9 @@
- 	  CYGWIN*)
+ 	  CYGWIN* | MSYS*)
  	    file_conv=cygwin
  	    ;;
 +	  MSYS*)
```
jeremyd2019, line 8: Seems like this hunk could be removed now

gnutls/gnutls-3.6.3-msysize.patch
```diff
@@ -12,7 +12,7 @@ diff -Naur gnutls-3.6.3-orig/build-aux/ar-lib gnutls-3.6.3/build-aux/ar-lib
  	    file_conv=wine
  	    ;;
 @@ -68,6 +71,9 @@
- 	cygwin)
+ 	cygwin | msys)
  	  file=`cygpath -m "$file" || echo "$file"`
  	  ;;
 +	msys)
```
jeremyd2019, line 18: And this one

gnutls/gnutls-3.6.3-msysize.patch
Biswa96: Ya, I realized that after pushing that. But this may not cause severe issue. I shall remove it next update.

gnutls/gnutls-3.6.3-msysize.patch
Biswa96: Ya, I realized that after pushing that. But this may not cause severe issue. I shall remove it next update.

PR #2551, opened 2021-06-18
msys2-runtime: copy `usr/libexec` folder also

naveen521kk:
it was missed in https://github.com/msys2/MSYS2-packages/pull/2547


dscho (approved):
!!!

Issue #574, opened 2016-04-27
PKGBUILDs for perl-rename and autojump... would adding these be possible?


Todd-Weed:
I like both of these packages and find them beneficial; they do respectively:

1) rename multiple files using Perl regular expressions
2) a faster way to navigate your filesystem from the command line

Their original package sources can be found [here](https://www.archlinux.org/packages/community/any/perl-rename/) and [here](https://www.archlinux.org/packages/community/any/autojump/).

Both are basically scripts so worrying about dependency won't be an issue. Additionally they likely won't go through many updates so maintenance should be light.

The original PKGBUILDS I have edited (for linking), and have built, installed, and tested them:

```bash
# $Id$
# Maintainer: Florian Pritz <flo@xinu.at>

pkgname=perl-rename
pkgver=1.9
pkgrel=1
pkgdesc="Rename multiple files using Perl regular expressions"
arch=(any)
url="http://search.cpan.org/~pederst/rename/"
license=('unknown')
depends=(perl)
options=(!emptydirs)
source=(http://search.cpan.org/CPAN/authors/id/P/PE/PEDERST/rename-$pkgver.tar.gz)
md5sums=('16df2adde955a6867701564e3d7c6a52')

build() {
    cd "$srcdir/rename-$pkgver"

    perl Makefile.PL PREFIX=/usr INSTALLDIRS=vendor
    make
}

package() {
    cd "$srcdir/rename-$pkgver"

    make DESTDIR="$pkgdir"/ install
    mv "$pkgdir"/usr/bin/vendor_perl/{rename,perl-rename}
    mv "$pkgdir"/usr/share/man/man1/{rename,perl-rename}.1p
}
```

```bash
# $Id$
# Maintainer: Jaroslav Lichtblau <svetlemodry@archlinux.org>
# Contributor: graysky <graysky AT archlnux DOT us>
# Contributor: Geoffroy Carrier <geoffroy@archlinux.org>
# Contributor: Jo├½l Schaerer <joel.schaerer@laposte.net>
# Contributor: Daniel J Griffiths <ghost1227@archlinux.us>

pkgname=autojump
pkgver=22.3.0
pkgrel=1
pkgdesc="A faster way to navigate your filesystem from the command line"
arch=('any')
url="https://github.com/wting/autojump"
license=('GPL3')
depends=('python')
source=($pkgname-$pkgver.tar.gz::https://github.com/wting/$pkgname/archive/release-v$pkgver.tar.gz)
[...]

PR #2552, opened 2021-06-18
add perl-rename

Atlas48:
attempt to resolve issue #574, plus I really would like to use this easily.

### DIFF
--- a/perl-rename/PKGBUILD
+++ b/perl-rename/PKGBUILD
@@ -1,0 +1,29 @@
+# $Id$
+# Maintainer: Florian Pritz <flo@xinu.at>
+
+pkgname=perl-rename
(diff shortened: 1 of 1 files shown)


FJEagle:
Yes, adding this rename package would be appreciated

Aerocatia:
Would this still be possible to merge?

striezel:
> Would this still be possible to merge?

@Aerocatia, @Atlas48: Not sure.

I've just updated the package to version 1.16 and made some other minor adjustments. However, I have no idea whether the newer version still works as intended. Package seems to build fine, but I'd prefer if someone could actually download and test the package from the CI build and play around with it a bit to see whether the package is OK or needs further adjustments.

PR #2553, opened 2021-06-18
add autojump

Atlas48:
attempt to resolve issue #574


Biswa96:
New packages are not accepted in msys2 except if it is required to build packages in mingw. MSYS2 focuses on building native software built against the Windows APIs.

PR #2557, opened 2021-06-21
msys2-runtime: patch to fix ctrl-c handling

jeremyd2019:
msys2/msys2-runtime#48


dscho (approved):
Better late than never :-)

PR #2573, opened 2021-07-06
Add IAA mirror to pacman-mirrors

jtanx:
The Internet Association of Australia has kindly agreed to mirror msys2 - both via https/rsync:

https://mirror.internet.asn.au/pub/msys2
rsync://mirror.internet.asn.au/msys2

It has been set up to sync every 4 hours.

### DIFF
--- a/pacman-mirrors/PKGBUILD
+++ b/pacman-mirrors/PKGBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Ray Donnelly <mingwandroid@gmail.com>
 
 pkgname=pacman-mirrors
-pkgver=20210703
+pkgver=20210706
 pkgrel=1
 pkgdesc="MSYS2 mirror list for use by pacman"
 arch=('any')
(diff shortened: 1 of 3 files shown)


jeremyd2019:
nice, there was just a discussion about how mirrors were lacking for Australia.

lazka:
Thanks!

lazka:
Also added to mirrorbits https://github.com/msys2/msys2-main-server/commit/a4e55bbbcd07c5c5699cbf9dd123f0b2eb9dcafe -> https://mirror.msys2.org/distrib/msys2-x86_64-latest.exe?mirrorlist&fromip=1.0.7.255

jtanx:
Ah, nice thanks!

Any clue why it's listed as unreachable on that link? https://mirror.internet.asn.au/pub/msys2/distrib/msys2-x86_64-latest.exe works for me.

Also any objections to them posting about this on their social media links?

Biswa96:
The mirror is now green in the map. BTW, if you zoom in the map here https://packages.msys2.org/mirrors, the IAA mirror is shown in the middle of Brisbane river 😕

jeremyd2019:
> BTW, if you zoom in the map here https://packages.msys2.org/mirrors, the IAA mirror is shown in the middle of Brisbane river 😕

I think MS was experimenting with datacenters in shipping containers under the ocean, for cooling, maybe this is related?

lazka:
actually all locations returned by geoip2 seem to be in water or public parks sport areas. I'd guess they do it on purpose. (privacy reasons maybe? or reduced accuracy for the free api?)

lazka:
as for green vs non green. mirrorbits pings the mirrors every 5 minutes for status updates, could be it timed out once.

jeremyd2019:
There was a story a while back about how geoip coordinates for the US (when it didn't know anything more than the country) wound up on some property in Kansas, and how police kept showing up to that propery looking for all kinds of crimes going on, because they plugged an IP address in and got those coordinates.  I recall they 'fixed' it by moving the coordinates to a body of water or something, to try to save the property owners the hassle.  Maybe they did the same across the board

jtanx:
Hmm, yeah odd, I see it still intermittently going between unreachable and available on https://mirror.msys2.org/?mirrorstats

But whenever I check myself, it seems up.

PR #2602, opened 2021-08-13
WIP: Port libarchive to CMake

ongun-kanat:
This is my first contribution to the repository so I may need some changes. This is a WIP PR. I want to make sure the build works and meets the expectations before merging it. 

Since the CMakeLists provided the option for dynamically linking `bsdtar` and `bsdcpio` with libarchive, I enabled that setting. Therefore `bsdtar` and `bsdcpio` now depend on `libarchive` which provides all the other dependencies.

Unfortunately I cannot build autotools version of libarchive. Some autoconf rules seem to be missing in my default MSYS2 installation with `base-devel` group of packages (missing `makedepends()` ?). Therefore, I cannot make a 1-to-1 comparison. 

I plan to add  a `check()` function to PKGBUILD in order to run `make test`. Currently a few of them fails/segfaults. The fails seem to be actual programming errors in libarchive and they may need to be patched.

---
Long logfile of failed tests
[LastTest.log](https://github.com/msys2/MSYS2-packages/files/6980117/LastTest.log)

### DIFF
--- a/libarchive/PKGBUILD
+++ b/libarchive/PKGBUILD
@@ -2,15 +2,15 @@
 
 pkgname=('libarchive' 'libarchive-devel' 'bsdcpio' 'bsdtar')
 pkgver=3.5.1
-pkgrel=1
+pkgrel=2
 pkgdesc="library that can create and read several streaming archive formats"
 arch=('i686' 'x86_64')
 url="https://libarchive.org/"
 license=('BSD')
 groups=('libraries')
 depends=('gcc-libs' 'libbz2' 'libiconv' 'libexpat' 'liblzma' 'liblz4' 'libnettle' 'libxml2' 'libzstd' 'zlib')
-makedepends=('libbz2-devel' 'libiconv-devel' 'libexpat-devel' 'liblzma-devel' 'liblz4-devel' 'libnettle-devel' 'libxml2-devel' 'libzstd-devel' 'zlib-devel')
-options=('!strip' 'debug' 'libtool')
+makedepends=('libbz2-devel' 'libiconv-devel' 'libexpat-devel' 'liblzma-devel' 'liblz4-devel' 'libnettle-devel' \
+             'libxml2-devel' 'libzstd-devel' 'zlib-devel' 'cmake')
 source=("https://github.com/libarchive/libarchive/releases/download/${pkgver}/${pkgname}-${pkgver}.tar.xz"
         'libarchive-3.3.2-bcrypt-fix.patch'
         'libarchive-3.3.1-msys2.patch')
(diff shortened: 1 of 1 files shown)


lazka:
> Therefore, I cannot make a 1-to-1 comparison.

The CI output shows a filesystem level diff at least.

libarchive/PKGBUILD
```diff
@@ -2,15 +2,15 @@
...
+pkgrel=2
 pkgdesc="library that can create and read several streaming archive formats"
 arch=('i686' 'x86_64')
 url="https://libarchive.org/"
 license=('BSD')
 groups=('libraries')
 depends=('gcc-libs' 'libbz2' 'libiconv' 'libexpat' 'liblzma' 'liblz4' 'libnettle' 'libxml2' 'libzstd' 'zlib')
-makedepends=('libbz2-devel' 'libiconv-devel' 'libexpat-devel' 'liblzma-devel' 'liblz4-devel' 'libnettle-devel' 'libxml2-devel' 'libzstd-devel' 'zlib-devel')
-options=('!strip' 'debug' 'libtool')
+makedepends=('libbz2-devel' 'libiconv-devel' 'libexpat-devel' 'liblzma-devel' 'liblz4-devel' 'libnettle-devel' \
```
Biswa96, line 12: No need to add `\`

libarchive/PKGBUILD
```diff
@@ -24,46 +24,34 @@ prepare() {
   # msysize patch
   #patch -Np1 -i "${srcdir}/libarchive-3.3.1-msys2.patch"
   patch -Np1 -i "${srcdir}/libarchive-3.3.2-bcrypt-fix.patch"
-  autoreconf -ivf
 }
 
 build() {
-  export lt_cv_deplibs_check_method='pass_all'
-  cd "${pkgname}-${pkgver}"
+  local build_type=Release
```
Biswa96, line 30: local custom variables should start with underscore `_` Link https://man.archlinux.org/man/PKGBUILD.5

lazka: hm, why?

Biswa96: "This will prevent any possible name clashes with internal makepkg variables."

lazka: yes, but we are in a function scope here, so that shouldn't apply.

Biswa96: 👍

libarchive/PKGBUILD
```diff
@@ -24,46 +24,34 @@ prepare() {
...
-    --without-libiconv-prefix \
-    --without-xml2 \
-    --without-cng \
-    --without-lzo2
+  mkdir -p "${pkgname}-${pkgver}/build-${CHOST}"
+  cd "${pkgname}-${pkgver}/build-${CHOST}"
 
-# CNG breaks pacman on Vista due to BCryptDeriveKeyPBKDF2
+  cmake \
+    -G"Unix Makefiles" \
```
1480c1, line 41: Have you tried using Ninja?

libarchive/PKGBUILD
ongun-kanat: Fixed

libarchive/PKGBUILD
ongun-kanat: I didn't since I wasn't sure `DESTDIR` worked with Ninja. Apparently it works. However this issue with Ninja is annoying: https://github.com/ninja-build/ninja/issues/1482

jeremyd2019: Yes it is.  `CMAKE_BUILD_PARALLEL_LEVEL` can be used with `cmake --build` (instead of calling ninja directly)

ongun-kanat:
> 
> 
> > Therefore, I cannot make a 1-to-1 comparison.
> 
> The CI output shows a filesystem level diff at least.

libarchive seems to use a unique way to give "sonames" to their libraries: https://github.com/libarchive/libarchive/blob/3.5.1/CMakeLists.txt#L74 So it changes file name to `*-18.dll`

libarchive/PKGBUILD
ongun-kanat: Hmm I should amend the commit with that. Thanks for the info @jeremyd2019

jeremyd2019:
If soname changes, all dependencies need to be rebuilt.  Luckily there aren't too many: https://packages.msys2.org/package/libarchive?repo=msys&variant=x86_64

ongun-kanat:
> 
> 
> If soname changes, all dependencies need to be rebuilt. Luckily there aren't too many: https://packages.msys2.org/package/libarchive?repo=msys&variant=x86_64 (see "Required by" section)

I know :) I haven't investigated libarchive source code history to have enough confidence in ABI stability of their interfaces. If they want to increment soname at each release, though, I wouldn't get my hopes high. BTW it seems the dependents of libarchive use static libs instead of dlls. It's done to not bork a working installation, I guess?

lazka:
https://github.com/libarchive/libarchive/issues/1065 :D

lazka:
So either we fix https://github.com/libarchive/libarchive/blob/2d57843d3c1bd6061df94d2554d3349ef60bf8e3/CMakeLists.txt#L78 to not add the minor version (it's pretty clear this is wrong as it is now imo), or we rebuild (bump pkgrel of all packages in this PR)

ongun-kanat:
> So either we fix https://github.com/libarchive/libarchive/blob/2d57843d3c1bd6061df94d2554d3349ef60bf8e3/CMakeLists.txt#L74 to not add the minor version (it's pretty clear this is wrong as it is now imo), or we rebuild (bump pkgrel of all packages in this PR)

From libarchive/libarchive#1065 I infer that the soname in the CMake is the correct one. So every minor version is a soname change.

PR #2609, opened 2021-08-22
pacman: switch mingw i686 -march to pentium4

jeremyd2019:
We had discussed this on Discord a while ago.  `-march=pentium4` seemed like a reasonable minimum, requiring instruction sets through sse2.  This has been the default 32-bit instruction set for Visual Studio starting in Visual Studio 2012, and is also the default in clang (though the `-march=i686` option here has been overriding it to target Pentium Pro instead).

My personal recollection of the last CPU I had to deal with that didn't have SSE2 was the Athlon XP, circa 2005.  All 64-bit capable CPUs support SSE2.

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=pacman
 pkgver=6.0.0
-pkgrel=5
+pkgrel=6
 pkgdesc="A library-based package manager with dependency support (MSYS2 port)"
 arch=('i686' 'x86_64')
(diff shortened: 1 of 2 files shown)


lazka:
For completeness, https://www.ghacks.net/2018/06/21/windows-7-support-dropped-for-cpus-without-sse2/ Win7 requires SSE2, and since we require Win7 this should be fine.

jeremyd2019:
Ok.  I had to run Win7 on Athlon XP CPUs, but not as recently as 2018.

PR #2623, opened 2021-08-29
pacman: try to improve makepkg-mingw defaults

lazka:
* Print MINGW_ARCH info when passing --help/-h
* Change behaviour when MINGW_ARCH isn't set
  * Default to MSYSTEM if in a mingw environment
  * Default to mingw64 otherwise (but warn)
* Print some feedback on what is getting built right away.
  makepkg can take some time at the start and this makes
  things clear right away.

The goal of the MINGW_ARCH change is to make it more likely to do the right thing by
default and to not special case any environment.

This has the potential of breaking existing users if MINGW_ARCH isn't set
and they want exactly "mingw32/mingw64" to be built like before. This will now build
only mingw64 if called in a msys environment and either mingw32/64 if called
in those environments.

To restore the old behavior set MINGW_ARCH="mingw64 mingw32" first.

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=pacman
 pkgver=6.0.1
-pkgrel=1
+pkgrel=2
 pkgdesc="A library-based package manager with dependency support (MSYS2 port)"
 arch=('i686' 'x86_64')
 url="https://www.archlinux.org/pacman/"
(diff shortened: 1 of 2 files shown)


elieux:
Cool, cool.  Only things I would point out is to improve on the help screen by printing the available environments and the current default.

elieux:
And...  How does this interact with *mingw_arch* in the recipes?  Maybe we should think about a way to build all the supported arches in one command?

lazka:
> And... How does this interact with _mingw_arch_ in the recipes? Should we think about a way to build all the supported arches in one command?

nothing atm. A "all" thing would be possible, but what to do about clangarm64?

PR #2633, opened 2021-09-05
Add localepurge package

orgads:
Imported from Arch[1], with minor adaptations.

[1] https://aur.archlinux.org/packages/localepurge


lazka:
What motivated you to propose this? thanks

orgads:
Clear disk space. See also https://sourceforge.net/p/msys2/mailman/message/35989985/

This is far from perfect though.

1. It has to be executed manually (on debian it is installed as a trigger, which is automatically executed after installing new packages).
2. I get many "missing files" warnings when uninstalling a package.

lazka:
I see, thanks

PR #2635, opened 2021-09-06
CI: merge with MINGW-packages

jeremyd2019:


jeremyd2019:
Test run: https://github.com/jeremyd2019/MSYS2-packages/actions/runs/1204607008

jeremyd2019:
I didn't try to look at the python 'check' job, I don't really know if that would translate to MSYS2.

.ci/ci-library.sh
```diff
@@ -114,29 +73,18 @@ execute(){
...
+# Get changed packages in correct build order
 list_packages() {
-    local _packages
-    _list_changes _packages '*/PKGBUILD' '%/PKGBUILD' --pretty=format: --name-only || return 1
-    for _package in "${_packages[@]}"; do
-        local find_case_sensitive="$(find -name "${_package}" -type d -print -quit)"
-        test -n "${find_case_sensitive}" && packages+=("${_package}")
-    done
-    return 0
+    _as_list "packages" "*" "" "$($DIR/ci-get-build-order.py)"
```
jtanx, line 83: One thing I didn't realise when switching to use `_as_list`, it assumes that no output is a failure, but in this case it should actually be allowed (see line 27 in ci-build.sh) - that's why the build is failing right now.

jeremyd2019: msys2/MINGW-packages#9545

PR #2650, opened 2021-09-27
pacman: Add RUSTFLAGS to mingw_mingw.conf

MehdiChinoune:


mati865:
Note: the default release level for Rust is `opt-level=3`.

MehdiChinoune:
> Note: the default release level for Rust is `opt-level=3`.

You mean cargo release profile.
We have the same situation with CMake/C/CXX where the default optimization with Release Configuration is `-O3`

jeremyd2019:
It looks like upstream makepkg.conf uses `opt-level=2` and `debuginfo=2`
https://github.com/msys2/msys2-pacman/blob/fd4c8f1f9fd7add827bceb512ac68dee97e81577/etc/makepkg.conf.in#L44

lazka:
It's not used in Arch though: https://github.com/msys2/msys2-pacman/blob/fd4c8f1f9fd7add827bceb512ac68dee97e81577/etc/makepkg.conf.in#L44

What does this change give us?

MehdiChinoune:
> It's not used in Arch though: https://github.com/msys2/msys2-pacman/blob/fd4c8f1f9fd7add827bceb512ac68dee97e81577/etc/makepkg.conf.in#L44
> 
> What does this change give us?

It gives us default compile flags for rust-based packages? like C/C++ !

lazka:
I have nothing against the change in general, I'm just curious why it is needed.

* Why does Arch not need to set them, but we do?
* Wouldn't setting opt-level=2 be a downgrade if 3 is the default?

MehdiChinoune:
> I have nothing against the change in general, I'm just curious why it is needed.
> 
> * Why does Arch not need to set them, but we do?

Maybe because most of rust packages use cargo as a buildsystem?
That's up to you to decide.
> * Wouldn't setting opt-level=2 be a downgrade if 3 is the default?

`opt-level=3` is default with cargo release profile (like CMake release build-type).
https://doc.rust-lang.org/rustc/codegen-options/index.html#opt-level
the default is `opt-level=0`, while `-O` is an alias for `opt-level=2`.

mati865:
AFAICT opt-level in RUSTFLAGS will override one from Cargo profile.

MehdiChinoune:
> AFAICT opt-level in RUSTFLAGS will override one from Cargo profile.

You are right. https://doc.rust-lang.org/cargo/reference/environment-variables.html

`cargo build` default to dev profile unless it gets `--release` passed.
https://doc.rust-lang.org/cargo/reference/profiles.html#profile-selection

PR #2653, opened 2021-10-02
ca-certificates: drop expired "DST Root CA X3"

rimrul:
Let's encrypt signs current certificates based on their root
certificate "ISRG Root X1". They provide two different versions
of "ISRG Root X1", though. One is self-signed using "ISRG Root X1"
and one is cross-signed using "DST Root CA X3" [1]. Some systems
still distribute certificate chains with the cross-signed version.
"DST Root CA X3" expired on 2021-09-30. This can lead to valid
certificates being considered invalid on devices with the expired
certificate still in their CA certificate bundle.

Ubuntu[2] and RedHat[3][4] recently dropped this certificates from their
`ca-certificates` packages for the same reasons.

The blacklist code in certdata2pem is copied from the debian version
of the file as it comes with ca-certificates_20210119.tar.xz.

[1] https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021
[2] https://ubuntu.com/security/notices/USN-5089-1
[3] https://access.redhat.com/errata/RHBA-2021:3649
[4] https://access.redhat.com/articles/6338021

### DIFF
--- a/ca-certificates/PKGBUILD
+++ b/ca-certificates/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=ca-certificates
 pkgver=20210119
-pkgrel=2
+pkgrel=3
 pkgdesc='Common CA certificates'
 arch=('any')
 url='https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/'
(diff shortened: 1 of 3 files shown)


lazka:
thanks

PR #2658, opened 2021-10-05
Add MINGW-x86_64-v3 subsystem

MehdiChinoune:
see: https://github.com/msys2/MINGW-packages/discussions/9456


Alexpux:
Why we need this?

MehdiChinoune:
> Why we need this?

I prefer to continue discussion in msys2/MINGW-packages#9456

PR #2666, opened 2021-10-18
pactoys-git: use fork 'msys2/pactoys'

eine:
Consume pactoys from https://github.com/msys2/pactoys, instead of the outdated https://github.com/renatosilva/pactoys.

### DIFF
--- a/pactoys-git/PKGBUILD
+++ b/pactoys-git/PKGBUILD
@@ -2,8 +2,8 @@
 
 _realname='pactoys'
 pkgname="${_realname}-git"
-pkgver=r2.07ca37f
-pkgrel=2
+pkgver=r51.43181b1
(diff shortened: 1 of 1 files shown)


eine:
I thought the meaning of `*-git` packages is that they retrieve the master/main branch at build time, while the packages without appendix do consume a pinned commit (either through git or through a tarball). For instance:

- https://github.com/msys2/MSYS2-packages/blob/master/easyoptions-git/PKGBUILD
- https://github.com/msys2/MSYS2-packages/blob/master/colormake-git/PKGBUILD
- https://github.com/msys2/MSYS2-packages/blob/master/libtre-git/PKGBUILD

Shall I pin the commit here and rename the package from `pactoys-git` to `pactoys`?

PR #2667, opened 2021-10-18
Open62541(git) package added

Samega7Cattac:


Samega7Cattac:
Seems to be a issue with the compiler on CI, tested multiple times on my machine and compiled without issues

Issue #2657, opened 2021-10-04
Missing stdbuf from coreutils package


JasonQiu1:
`coreutils` package is not installing `stdbuf`. 

The output of `pacman -Ql coreutils` shows that `stdbuf` cannot be found in `/usr/bin`, and its manual cannot be found in `/usr/share/man/man1`. 

This may have something to do with [issue #541](https://github.com/msys2/MSYS2-packages/issues/541) and the [corresponding successful pull request #1019](https://github.com/msys2/MSYS2-packages/pull/1019) or one of the following cleanup commits.

To reproduce this problem, install MSYS2 and upgrade everything with `pacman -Syuu` twice. Try running `stdbuf`, and you will get the following:
```
$ stdbuf
bash: stdbuf: command not found
```

Running `stdbuf` should instead produce help text regarding the program, as it is a part of the gnu coreutils package,

PR #1019, opened 2017-09-30
coreutils: fix libstdbuf.so.exe problem once and for all

dscho:
The `stdbuf.exe` program links to a `libstdbuf.so` file, so if that does
not exist, there is a problem.

In d9773f0a (coreutils: Update to 8.24, 2015-08-28), a sed call was
introduced to ensure that `libstdbuf.so` is built and installed under
that name, without `.exe` suffix.

As reported in https://github.com/Alexpux/MSYS2-packages/issues/541,
however, that fix no longer works, the `libstdbuf.so.exe` file is in
coreutils 8.26-1.

Let's just rename the `libstdbuf.so.exe` file after it has been
installed into the `pkgdir` tree.

Note: we still have to muck around with `Makefile.in` so that the build
does not break with

	make[2]: *** No rule to make target 'src/libstdbuf.so', needed
	by 'all-am'.  Stop.

Signed-off-by: Johannes Schindelin <johannes.schindelin@gmx.de>

Issue #541, opened 2016-04-08
stdbuf doesn't work due to missing libstdbuf.so


elieux:
I think this is caused by makepkg pre-packaging renames.

Problem:
```
$ stdbuf.exe -oL -eL -i0 sh
      0 [main] sh 9992 C:\msys64\usr\bin\sh.exe: *** fatal error - error while loading shared libraries: /usr/lib/coreutils/lib stdbuf.so: cannot open shared object file: No such file or directory
Hangup
```

Workaround:
```
$ mv /usr/lib/coreutils/libstdbuf.so.exe /usr/lib/coreutils/libstdbuf.sox
$ cp /usr/lib/coreutils/libstdbuf.sox /usr/lib/coreutils/libstdbuf.so
$ mv /usr/lib/coreutils/libstdbuf.sox /usr/lib/coreutils/libstdbuf.so.exe
```

larsxschneider:
With the latest Git for Windows 2.14.1 we get the same error using `stdbuf`:
```
$ stdbuf -oL ls
      0 [main] ls 5460 C:\Program Files\Git\usr\bin\ls.exe: *** fatal error - error while loading shared libraries: /usr/lib/coreutils/libstdbuf.so: cannot open shared object file: No such file or directory
Hangup
```

If we rename `C:\Program Files\Git\usr\lib\coreutils\libstdbuf.so.exe` to `C:\Program Files\Git\usr\lib\coreutils\libstdbuf.so` then it works.

@Alexpux / @dscho : Can you give me a hint where this could be fixed? Thanks!

dscho:
@larsxschneider `pacman` to the rescue:

```sh
$ pacman -Qo /usr/lib/coreutils/libstdbuf.so.exe
/usr/lib/coreutils/libstdbuf.so.exe is owned by coreutils 8.26-1
```

So the culprit is that something in `coreutils` is built incorrectly.

Looking at the corresponding PKGBUILD file, I see that there is *supposedly* a fix for it already: https://github.com/Alexpux/MSYS2-packages/blob/e6d639a60034244bf3be0b2e61e1d910f7894c21/coreutils/PKGBUILD#L38

However, that fix obviously does not work, or it works and nobody bothered to rebuild the package since. The latter seems not to be the case, as the latest change to `PKGBUILD` was b6fa9fd7e2a62e5837614c4167699d9b07326172 that updated to  8.26-1, which is the current version, so that `sed` call must have already been in there.

[...]

PR #2669, opened 2021-10-21
coreutils: Re-add disabled stdbuf; some cleanup

cbrt64:
Fix for #2657. 

`stdbuf` has been in and out of MSYS2/coreutils since 2014. Here's a short history for reference:

| coreutils | what happened | stdbuf status (presumed?) |
| --- | --- | --- |
| 8.23-1 | 1) A local [patch](https://github.com/msys2/MSYS2-packages/blob/5c64d6da84b6a3ad9cc651a33031e33ff335334c/coreutils/003-fix-libstdbuf-link.patch) was added to fix a build problem (I assume introduced by upstream putting libstdbuf in pkglibexec_PROGRAMS) | (worked?) |
| 8.24 upstream | 2) Building stdbuf on Windows was [disabled](https://git.savannah.gnu.org/cgit/coreutils.git/commit/?id=0cf66cbe85009213bcf2608eceeee1355f367667) for the same problem | |
| 8.23-2 | 1) was removed and 2) was downstreamed [here](https://github.com/msys2/MSYS2-packages/commit/61ff57c053088cc7240d83acaef8cf2f392fe609) | (no stdbuf?) |
| 8.24-1 | 2) was reverted [locally](https://github.com/msys2/MSYS2-packages/commit/d9773f0a8cdb2ce28dbe6f2088cbd65256afb9d6), and a [different build fix](https://github.com/msys2/MSYS2-packages/blame/d9773f0a8cdb2ce28dbe6f2088cbd65256afb9d6/coreutils/PKGBUILD#L38) added | (works(?), but not OOB due to wrong file extension) |
| 8.25-1 | #541 | ditto (I tested this version on an old setup) |
| 8.26-2 | `libstdbuf.so.exe` fixed in #1019 | (works?) |
| 8.30-1 | reversion of 2) was removed (accidentally?) in #1416 | gone again |

This PR:
1. Uses the original fix from 8.23, instead of from 8.24. Adding rather than removing `$EXEEXT` makes more sense to me because
A. upstream's choice of `pkglibexec_PROGRAMS` [creates executables, not libraries](https://github.com/msys2/MSYS2-packages/issues/541#issuecomment-333266581) (note that libstdbuf is built as a DLL either way),
B. it's compatible with upstream, and
C. it can be a patch instead of a sed script in PKGBUILD
2. Converts remaining post-autoreconf stuff to a patch
3. Renames libstdbuf.so to libstdbuf.dll because why not

I skipped checking after my first build run, since `check()` still fails as in #1416.

### DIFF
--- a/coreutils/002-coreutils-8.32-enable-stdbuf.patch
+++ b/coreutils/002-coreutils-8.32-enable-stdbuf.patch
@@ -1,0 +1,31 @@
+--- coreutils-8.32/configure.ac
++++ coreutils-8.32/configure.ac
+@@ -509,7 +509,7 @@
+   ],
+   [stdbuf_supported=yes])
+ AC_MSG_RESULT([$stdbuf_supported])
+-if test "$stdbuf_supported" = "yes" && test -z "$EXEEXT"; then
++if test "$stdbuf_supported" = "yes"; then
+   gl_ADD_PROG([optional_bin_progs], [stdbuf])
+ fi
+ CFLAGS=$ac_save_CFLAGS
+@@ -530,7 +530,7 @@
+ # Note it adding to pkglibexec_PROGRAMS, $(transform) in src/local.mk
+ # may need to be updated accordingly.
+ case " $optional_bin_progs " in
+-  *' stdbuf '*) pkglibexec_PROGRAMS='src/libstdbuf.so';;
++  *' stdbuf '*) pkglibexec_PROGRAMS='src/libstdbuf.so$(EXEEXT)';;
+   *) pkglibexec_PROGRAMS='';;
+ esac
+ 
+--- coreutils-8.32/src/stdbuf.c
++++ coreutils-8.32/src/stdbuf.c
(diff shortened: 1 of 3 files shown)


cbrt64:
BTW, I tested `stdbuf` actually works by watching the output of
```C
/* stdprint.c */
#include <stdio.h>
#include <unistd.h>

int main (void) {
  while (1) {
    fprintf (stderr, "E");
    fprintf (stdout, "O");
    usleep (1);
  }
}
```
```C
/* passthru.c */
#include <stdio.h>

int main (void) {
  while (!feof (stdin))
    putchar (getchar ());
}
```

```bash
$ stdbuf -o0 -e64 ./stdprint  # vs
$ stdbuf -o64 -e0 ./stdprint

$ stdbuf -o0 ./stdprint | ./passthru  # vs
$ stdbuf -o0 ./stdprint | stdbuf -i0 ./passthru
```

PR #2671, opened 2021-10-22
u-boot-tools: Update to v2021.10

lbmeng:


lazka:
I'd guess because we only support symlinks when the target already exists, so it can copy it. So by extracting twice the targets get extracted, on the next run the symlinks get copied.

Not sure why I switched it back then, maybe because using official tarballs is nice + signatures...

PR #2700, opened 2021-11-15
filesystem: Avoid trying to chmod /dev/{shm,mqueue} during mkdir

mitchhentges:
When launching a custom msys2 environment, and it does "first-run"
operations, it complains that: "cannot change permissions of
'/dev/shm': Permission denied".

This is related to the previous work in this area that happened
in #2337: Since "/" is mounted with noacl and chmod() is a no-op,
this error is cosmetic and can be removed.

However, in #2337, the access modifying behaviour was not removed
from the "mkdir" branch, which this patch aims to address.

### DIFF
--- a/filesystem/01-devices.post
+++ b/filesystem/01-devices.post
@@ -44,7 +44,7 @@ maybe_create_devs ()
   # Create it if necessary
   if [ ! -e "${DEVDIR}/shm" ]
   then
-    mkdir -m 1777 "${DEVDIR}/shm"
+    mkdir "${DEVDIR}/shm"
     if [ ! -e "${DEVDIR}/shm" ]
     then
       echo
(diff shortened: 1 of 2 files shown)


mitchhentges:
I'm not entirely clear how this `cannot change permissions of '/dev/shm': Permission denied` error doesn't crop up in regular MSYS2 usage: it doesn't show up in the installer "details".

However, doing `mkdir -m 1777 <dir>` in an MSYS2 shell fails with the error message as expected 🤔.

lazka:
lgtm, but needs a pkgrel bump of the package and the checksums need to be updated (run `updpkgsums`)

mitchhentges:
Sounds good, thanks!
Thanks for your good work on MSYS2, I'm currently working to port [MozillaBuild](https://wiki.mozilla.org/MozillaBuild) over and it's been a good experience so far 😁

PR #2714, opened 2021-11-28
msys2-launcher: Update to 1.4

lazka:


podsvirov:
When about to update launcher for `i686` repo? It's strange because we can't launch `CLANG32` terminal on 32-bit platform.

Biswa96:
msys2 launcher [package](https://packages.msys2.org/package/msys2-launcher?repo=msys&variant=x86_64) now has clang32.exe. If you mean 32 bit msys2 repo it was deprecated long ago https://www.msys2.org/news/#2020-05-17-32-bit-msys2-no-longer-actively-supported

podsvirov:
I know the repository is deprecated, but that doesn't mean that no one else is using it. I rebuilt this package for myself. But not all users know how to do this.

lazka:
I updated the repo once since the announcement to help the arm64 effort, but that has since switched to 64bit and x64 emulation on win11.

jeremyd2019:
I do still update https://github.com/jeremyd2019/msys2-build32/releases/repo/, but no guarantees how long I can keep that going.  I'll probably give up once msys2-runtime updates to a version without 32-bit support.

PR #2721, opened 2021-12-04
libffi: Update to 3.4.2

lazka:


lazka:
private key fetching is broken by the libffi so bump in CI (dirmngr fails to start), so I guess we try in autobuild..

PR #2732, opened 2021-12-10
Add an autotools meta package

lazka:
The goal of this package is to depend on everything that is (usually)
needed for running autogen/autoreconf if needed,
configure and make, but not a compiler.

If packages would makedepend on this then we could make base-devel smaller and remove all the autotools things from there. So if a package builds with meson or cmake, it doesn't install autoconf in CI.

autotools is then just another build system

### DIFF
--- a/autotools/PKGBUILD
+++ b/autotools/PKGBUILD
@@ -1,0 +1,19 @@
+# Maintainer: Christoph Reiter <reiter.christoph@gmail.com>
+
+pkgname=autotools
+pkgver=2021.12.10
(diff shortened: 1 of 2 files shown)


Biswa96:
Do all the packages, that need autoconf, require to change makedepends?

lazka:
> Do all the packages, that need autoconf, require to change makedepends?

yes, eventually.

lazka:
> yes, eventually.

done here for msys2-packages: 38126d7

PR #2745, opened 2021-12-17
Remove autotools from base-devel

lazka:
All packages depend on it explicitly now.

### DIFF
--- a/autoconf/PKGBUILD
+++ b/autoconf/PKGBUILD
@@ -3,12 +3,11 @@
 
 pkgname=autoconf
 pkgver=2.71
-pkgrel=1
+pkgrel=2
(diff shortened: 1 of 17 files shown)


Biswa96:
Will base-devel be removed? Every script which depends on base-devel has to change to autotools?

lazka:
> Will base-devel be removed?

No, I'm just cleaning it up.

> Every script which depends on base-devel has to change to autotools?

At least in our repo that's already the case.

Biswa96:
CI has this error in clang while building vala:

```
/usr/bin/m4: internal error detected; please report this bug to <bug-m4@gnu.org>: Segmentation fault
```

lazka:
reverted the m4 rebuild for now

PR #2747, opened 2021-12-17
gettext: only put gettext itself into base-devel

lazka:


lazka:
similar crash to sed.. need to investigate first..

PR #2757, opened 2021-12-27
Add back a autoconf2.69 package

lazka:
Useful for rebuilding configure files when upstream only supports 2.69 such as bintuils and gcc


Biswa96:
If only a handful project depends on the older version couldn't we update the configure.ac file? Or ask upstream to update autoconf files?

lazka:
In this case I wanted to update gdb/binutils. There is no clear error message why the 2.71 produces broken files and I'd rather not deal with autoconf. But feel free :)

PR #2761, opened 2021-12-28
python: backport patch to not link against crypt

### LINKED ISSUE #2759: python: python-config still tries to link to `-lcrypt`
This is breaking gdb's configure.  It uses distutils.sysconfig to find what flags to use, and `python-config --libs` output includes `-lcrypt`.  So either `python-devel` needs to depend on `libcrypt-devel` again, or python-config (and sysconfig) needs to be patched to not try to link in `-lcrypt`.

lazka:
Fixes #2759


jeremyd2019:
The module base addresses didn't even overlap!

lazka:
yes, I was scared too :)

PR #2770, opened 2021-12-29
pacman-mirrors: add BFSU

alissa-tung:


lazka:
is there rsync support?

alissa-tung:
Sorry I am not sure. https://mirrors.bfsu.edu.cn/status/
Other mirrors are rsyncing, if BFSU MSYS2 is not maybe a issue can be filed

lazka:
We prefer mirrors which provide rsync access so we can monitor them. All others go to tier 2.

I've created #2775. I'll take it from here, so closing this for now.

Thanks!

alissa-tung:
That's reasonable. (BTW, I filed this MR because BFSU mirror is a mirror of Tuna mirror which is sometime faster and more stable). Maybe I gonna connect Tuna for help if I can.

lazka:
Turns out rsync is supported (I somehow did something wrong when testing for it I guess), so this is fine as is.

lazka:
Thanks!

PR #2771, opened 2021-12-29
msys2-runtime: update patches

### LINKED ISSUE #2752: pacman: sometimes hangs validating GPG signatures
I first saw this on Windows on ARM, but it seems to have shown up on msys2-autobuild recently too (https://github.com/msys2/msys2-autobuild/runs/4552699834?check_suite_focus=true and https://github.com/msys2/msys2-autobuild/runs/4575182521?check_suite_focus=true).

What I saw on ARM was that `pstree` showed two pacman processes, one a child of the other.  If I killed the child, via its winpid and taskkill, pacman would complain about a GPGME error and things would move on (with a failed invocation of pacman).  If I killed the parent, the child would still hang around as an apparent zombie.

I tried to debug it once a while back, but I seem to recall that attempting to attach a debugger to it caused it to 'unblock' and exit, which made trying to figure out what was going on pretty difficult.

jeremyd2019:
Fixes msys2/msys2-runtime#58.  Fixes #2752

### DIFF
--- a/.ci/ci-build.sh
+++ b/.ci/ci-build.sh
@@ -50,7 +50,7 @@ for package in "${packages[@]}"; do
         for pkg in *.pkg.tar.*; do
             pkgname="$(echo "$pkg" | rev | cut -d- -f4- | rev)"
             echo "::group::[install] ${pkgname}"
-            grep -qFx "${package}" "$(dirname "$0")/ci-dont-install-list.txt" || pacman --noprogressbar --upgrade --noconfirm $pkg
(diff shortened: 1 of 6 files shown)


jeremyd2019:
CI for msys2-runtime went 'funny' now that msys2-runtime-devel is no longer installed normally.  Trying to figure that out.

jeremyd2019:
The file diff for `msys2-runtime-devel` is wrong, but at least the whole run isn't failing now.

PR #2798, opened 2022-01-09
filesystem: fix nested `msys2_shell.cmd -full-path` calls

dscho:
When calling `msys2_shell.cmd -full-path` from _within_ an MSYS2 Bash, the `ORIGINAL_PATH` variable is set already, but incorrectly. Let's not use it.

This fixes https://github.com/msys2/msys2-runtime/issues/43.


lazka:
What about doing `ORIGINAL_PATH="${PATH}` here instead? https://github.com/msys2/MSYS2-packages/blob/d2c89935e769e901c3127a64cc355f496b8f83f0/filesystem/profile#L38

dscho:
> What about doing `ORIGINAL_PATH="${PATH}"` here instead?

This would not work, as we're in a `.cmd` script.

Also, I believe that trying to solve this at the `/etc/profile` level instead is prone to fail, as we might be fed a Windows-style path list in `ORIGINAL_PATH` at that point.

A possible alternative would be to apply https://github.com/git-for-windows/msys2-runtime/commit/f5523f7cdb472c6b40356dcdc9e757028b20b976 to the original MSYS2 runtime. I will prepare a PR for that.

dscho:
> A possible alternative would be to apply [git-for-windows/msys2-runtime@f5523f7](https://github.com/git-for-windows/msys2-runtime/commit/f5523f7cdb472c6b40356dcdc9e757028b20b976) to the original MSYS2 runtime. I will prepare a PR for that.

Here is the PR: https://github.com/msys2/msys2-runtime/pull/79

lazka:
hm, is ORIGINAL_PATH public API in some way? It seemed like an implementation detail to me.

dscho:
> hm, is ORIGINAL_PATH public API in some way? It seemed like an implementation detail to me.

It is an implementation detail, but by virtue of being an environment variable that is not only set by `/etc/profile`, but also used, it _is_ public.

> we are never using/reading `ORIGINAL_PATH` with that change, so nothing can go wrong imo.

Oh, but we are! If you call the MSYS2 Bash from any non-MSYS program directly after setting `MSYS2_PATH_TYPE=inherit`, which people do, it totally is read.

> I'll try to reproduce the issue.

It's very easy. Just call `/msys2_shell.cmd -full-path` in your MSYS2 Bash, and in the new window, call `echo $PATH` and observe the Windows-style paths (including colons after the drive letter, which won't work with Unix-style path lists, where a colon is used as path separator).

lazka:
> > hm, is ORIGINAL_PATH public API in some way? It seemed like an implementation detail to me.
> 
> It is an implementation detail, but by virtue of being an environment variable that is not only set by `/etc/profile`, but also used, it _is_ public.

fair enough, we could still try to not advertise them by using them in other files though.

> > we are never using/reading `ORIGINAL_PATH` with that change, so nothing can go wrong imo.
> 
> Oh, but we are! If you call the MSYS2 Bash from any non-MSYS program directly after setting `MSYS2_PATH_TYPE=inherit`, which people do, it totally is read.

With my patch above it is never read (by MSYS2 at least). But the problem I see is that if you create multiple nested shells that inherit then PATH gets longer and longer. I don't know if that can be a problem re hitting length and performance limits though, so maybe that's why the code tries to re-use the initially inherited PATH.

> > I'll try to reproduce the issue.
> 
> It's very easy. Just call `/msys2_shell.cmd -full-path` in your MSYS2 Bash, and in the new window, call `echo $PATH` and observe the Windows-style paths (including colons after the drive letter, which won't work with Unix-style path lists, where a colon is used as path separator).

ok, that runtime patch looks good then (moving patches upstream is always good). thanks!

Can we close this then?

dscho:
> Can we close this then?

I would like to wait for confirmation by @alanlivio that https://github.com/msys2/MSYS2-packages/pull/2805 fixes the issue, too (once the new `msys2-runtime` version is available in MSYS2).

lazka:
Sounds good

alanlivio:
@lazka, I tested the PR and, for me, it can be merged and the issue closed.

Issue #2801, opened 2022-01-10
make: 4.3-2 now uses posix_spawn, which seems to break on i686


jeremyd2019:
I know that i686 msys2 is no longer supported, but I also know at least @dscho may also hit this, so I thought I should track this in an issue rather than just my observations in Discord.

When make was rebuilt recently, it turned on using posix_spawn instead of vfork/exec for running commands.  This seems to be working OK on x86_64, but for some reason on i686 I am seeing I think *every* attempt to run something from make fail with something like:
```
make: /bin/sh: Invalid argument
```

I was able to narrow this down by adding `--disable-posix-spawn` configure argument to make's PKGBUILD, which allows i686 make to work again.  I used gdb to set breakpoints on posix_spawn and vfork, and from that can verify that both i686 and x86_64 4.3-2 use posix_spawn, while 4.3-1 used vfork.

I have not yet figured out what is going wrong in the posix_spawn path though, but it is concerning because I don't see an obvious reason why i686 vs x86_64 should matter for it.

dscho:
> When make was rebuilt recently

According to https://github.com/git-for-windows/git-sdk-32/commits/HEAD/usr/bin/make.exe, it wasn't rebuilt for i686, though. And at least I do not have any plans to build new a `make` package for i686. Do you?

jeremyd2019:
I try to build all package updates for i686, just for my own usage (and anybody else who may have a 32-bit machine and happens to want to use my packages).  I figured you were building i686 updates as well in support of git-sdk-32.

lazka:
For reference, I tried to build make in cygwin 32 bit and it has the same problem:

* rebuilding `cygport make.cygport all` results in a broken make
* Adding `CYGCONF_ARGS="--disable-posix-spawn"` to the cygport file and rebuilding again results in a good make

jeremyd2019:
https://cygwin.com/pipermail/cygwin/2022-January/250453.html

jeremyd2019:
https://cygwin.com/pipermail/cygwin/2022-January/250459.html

C reproducer:
```c
#include <stdlib.h>
#include <stdio.h>
#include <string.h>

#include <unistd.h>
#include <spawn.h>
#include <sys/wait.h>

extern char **environ;

int main()
{
    pid_t pid;
    char *argv[] = {"sh", "-c", "echo hi", NULL};
    posix_spawnattr_t attr;
    int status;
    short flags = POSIX_SPAWN_RESETIDS;

[...]

PR #2803, opened 2022-01-10
make: rebuild without posix_spawn support

lazka:
It makes problems on 32bit, so disable to fix that and just to
be safe.

See https://github.com/msys2/MSYS2-packages/issues/2801

### DIFF
--- a/make/PKGBUILD
+++ b/make/PKGBUILD
@@ -3,7 +3,7 @@
 
 pkgname=make
 pkgver=4.3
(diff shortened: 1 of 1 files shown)


lazka:
@jeremyd2019 I was curious and did some benchmarks. A fresh build of lapack takes 153s with posix_spawn and 122s without (repeated two times, about the same result). So posix_spawn() is slower

PR #2806, opened 2022-01-14
base-devel: remove patchutils/intltool

lazka:
patchutils:

* It mostly contains tools that are useful to developers, but are not used in
   a PKGBUILD directly.

intltool:

* It's long abandoned and everyone should be using gettext
  and we want to get rid of it in the long run.
* Many packages that need it depend on it anyway

### DIFF
--- a/base-devel/PKGBUILD
+++ b/base-devel/PKGBUILD
@@ -22,7 +22,6 @@ depends=(
   'groff'
   'texinfo'
   'texinfo-tex'
-  'intltool'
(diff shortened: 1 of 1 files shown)


Biswa96:
> It's long abandoned and everyone should be using gettext

Many mingw packages with meson build system search intltool. Should it be removed from there also?

lazka:
It depends, in theory they could manually call intltool in their meson.build files.

Issue #2807, opened 2022-01-15
[mingw-w64-cross-binutils] locks up while building


stahta01:
I am trying to build mingw-w64-cross-binutils under Window 7 64 bit and it locks up during
ld/genscripts.sh in x86_64-pc-mingw32 target.

I edited just the pkgrel and tried to see if server can build it. It has been running 4.5 hours when does it give up?

Tim

stahta01:
It failed at 6 hours running on the github CI server

jeremyd2019:
GHA hosted runners time out jobs at 6 hours.

I wonder if this is related to msys2/msys2-runtime#77?

stahta01:
Could be related; but, I am not sure. I am trying to build the package locally using an upgraded version of binutils and if that works; I would say it is not the same problem. Be likely a few hours before build locks up or fails. Tim S.

stahta01:
The build worked using version 2.37; so, it is not the same problem.

Tim S.

PR #2808, opened 2022-01-15
mingw-w64-cross-binutils: Update to version 2.37

stahta01:
The building of this package locks for me without this upgrade.
I have not used to package; but, plan to build the rest of these cross mingw packages (not doing the clang ones). I figure if they build then this package works.
See #2807 

Tim S.


stahta01:
Built and installed these packages to test mingw-w64-cross-binutils:
  mingw-w64-cross-headers-git
  mingw-w64-cross-crt-git
  mingw-w64-cross-windows-default-manifest
  mingw-w64-cross-winpthreads-git
  mingw-w64-cross-gcc
  mingw-w64-cross-zlib
  mingw-w64-cross-winstorecompat-git
  mingw-w64-cross-tools-git

I think that is the best testing I can figure out for now.
Removing work in progress from changes at this time.

Tim S.

PR #2826, opened 2022-01-28
docbook: update the install scripts and add docbook-xsl-ns

dscho:
The install scripts might have picked up the wrong `xmlcatalog` and therefore produced incorrect paths in `/etc/xml/catalog`.

Also, the `docbook-xsl-ng` package is required to build Git's HTML pages. Let's offer that package.


jeremyd2019:
I don't know if this is important or not in this particular case, but install scripts should avoid using absolute paths.  They are run with the cwd of `/` so instead of `/usr/bin/...` use `usr/bin/...`.  This allows them to run the correct thing when using an alternate `--root`.

lazka:
Both ways work, but without "/" is the convention for some reason: https://bbs.archlinux.org/viewtopic.php?pid=913891#p913891

not sure..

dscho:
Thank you for the review and the explanation, I changed it.

lazka:
thanks.

I found two more places, 8800ddb6b027, I hope that's OK.

dscho:
> thanks.
> 
> I found two more places, [8800ddb](https://github.com/msys2/MSYS2-packages/commit/8800ddb6b027afc8356c50fc1845099ab3161e8d), I hope that's OK.

D'oh! I thought I had checked all instances, but I must simply have forgotten that because...

```console
$ git grep xmlcatalog | grep -ve usr/bin/xmlcatalog -e /PKGBUILD:
docbook-mathml/docbook-mathml.install:    xmlcatalog --noout --create etc/xml/catalog
docbook-mathml/docbook-mathml.install:  xmlcatalog --noout --add "public" \
docbook-mathml/docbook-mathml.install:  xmlcatalog --noout --add "system" \
docbook-mathml/docbook-mathml.install:  xmlcatalog --noout --del "-//OASIS//DTD DocBook MathML Module V1.1CR1//EN" etc/xml/catalog
docbook-mathml/docbook-mathml.install:  xmlcatalog --noout --del "http://www.oasis-open.org/docbook/xml/mathml/1.1CR1/dbmathml.dtd" etc/xml/catalog
```

PR #2833, opened 2022-01-30
filesystem: warn users that Windows 7/8 support is going away

lazka:
This will be shown on a login shell, so the first two lines when
users open a MSYS2 terminal.

image

This is the linked website: https://www.msys2.org/docs/windows_support/

~~Maybe I should try set an en var, so it doesn't get shown again on nested login shells.~~ done

### DIFF
--- a/filesystem/PKGBUILD
+++ b/filesystem/PKGBUILD
@@ -4,8 +4,8 @@
 # Contributor: Alethea Rose <alethea@alethearose.com>
 
 pkgname=filesystem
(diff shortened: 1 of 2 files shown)


mati865 (approved):

naveen521kk (approved):

dscho:
Thank you for your consideration! I think we're fine in Git for Windows. AFAIR we're overriding most of MSYS2's Bash profile.

PR #2836, opened 2022-01-31
Fix the install script of `docbook-mathml`, too

dscho:
This is in the same spirit as https://github.com/msys2/MSYS2-packages/pull/2826 (where I really should have `git grep`ed for similar instances, but forgot).

Note: after this, `docbook/PKGBUILD` still contains calls to `xmlcatalog` without any `/usr/bin/` prefix. This is intentional because the `PATH` should be set correctly when _building_ MSYS packages, anyway (in contrast to the scenario where `.install` is called, i.e. when the packages are _installed_).

So this is the rule: use `usr/bin/xmlcatalog` in the `.install` scripts, and `xmlcatalog` in `PKGBUILD`.

### DIFF
--- a/docbook-mathml/PKGBUILD
+++ b/docbook-mathml/PKGBUILD
@@ -12,7 +12,7 @@ depends=('libxml2')
 install=docbook-mathml.install
 source=("https://www.docbook.org/xml/mathml/${pkgver}/dbmathml.dtd" 'LICENSE')
 sha256sums=('70da7140035621330f1b5ab6926197c3c3af467f2207d55a41f6396d9ad96abd'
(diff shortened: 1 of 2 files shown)


dscho:
Tsk, tsk: https://github.com/msys2/MSYS2-packages/pull/2836/commits/567230acbec84c9863cc55b9d3ad93e823c3ac74

dscho:
> Tsk, tsk: [567230a](https://github.com/msys2/MSYS2-packages/commit/567230acbec84c9863cc55b9d3ad93e823c3ac74)

Turns out that's yet another fall-out of 7f7316ba9b01db7f1854ed760b653537225d9d19.

lazka:
thanks!

Issue #2841, opened 2022-02-02
pacman failed retrieving file tree-1.8.0-1-x86_64 Maximum file size exceeded


michkot:
## Describe the issue

pacman failed retrieving file tree-1.8.0-1-x86_64 Maximum file size exceeded

## Steps to Reproduce the Problem

```$ pacman -S tree
resolving dependencies...
looking for conflicting packages...

Packages (1) tree-1.8.0-1

Total Installed Size:  0.07 MiB

:: Proceed with installation? [Y/n]
:: Retrieving packages...
 tree-1.8.0-1-x86_64 is up to date
error: failed retrieving file 'tree-1.8.0-1-x86_64.pkg.tar.xz?viasf=1.sig' from master.dl.sourceforge.net : Maximum file size exceeded
warning: failed to retrieve some files
error: failed to commit transaction (download library error)
Errors occurred, no packages were upgraded.
```

seems to be the same underyling problem like #1010, #1025 , ...

Biswa96:
Try that after deleting the SourceForge link from every files in /etc/pacman.d directory.

lazka:
what's the pacman version?

lazka:
no response, so closing. our sourceforge mirror is gone now too.

Issue #2575, opened 2021-07-07
pacman -Syu -> failed retrieving file


SloBloLabs:
## Describe the issue

updating my repo today and running into the following issue (See full log below):
error: failed retrieving file 'mingw-w64-x86_64-python-astroid-2.4.2-1-any.pkg.tar.zst?viasf=1.sig' from master.dl.sourceforge.net : Maximum file size exceeded
warning: failed to retrieve some files
error: failed to commit transaction (download library error)
Errors occurred, no packages were upgraded.

## Steps to Reproduce the Problem
$ pacman -Syu
:: Synchronizing package databases...
 mingw32 is up to date
 mingw64 is up to date
 ucrt64                                                                                                                               1468.7 KiB  1834 KiB/s 00:01 [####################################################################################################] 100%
 clang64                                                                                                                              1095.6 KiB  1246 KiB/s 00:01 [####################################################################################################] 100%
 msys is up to date
:: Starting core system upgrade...
 there is nothing to do
:: Starting full system upgrade...
resolving dependencies...
looking for conflicting packages...
warning: dependency cycle detected:
warning: mingw-w64-x86_64-libwebp will be installed before its mingw-w64-x86_64-libtiff dependency
warning: dependency cycle detected:
warning: mingw-w64-x86_64-harfbuzz will be installed before its mingw-w64-x86_64-freetype dependency

Packages (202) asciidoc-9.1.0-2  autoconf-2.71-1  automake1.16-1.16.3-1  binutils-2.36.1-4  bison-3.7.6-1  brotli-1.0.9-2  bsdcpio-3.5.1-1  bsdtar-3.5.1-1  ca-certificates-20210119-1  cmake-3.20.4-1  curl-7.76.1-1  dash-0.5.11.4-1  diffstat-1.64-1  dtc-1.6.0-4
               expat-2.3.0-1  file-5.40-2  findutils-4.8.0-1  gdb-10.2-3  gdbm-1.19-1  git-2.32.0-1  glib2-2.68.1-1  gnupg-2.2.28-3  help2man-1.48.3-1  icu-68.2-1  lemon-3.36.0-1  less-581-1  libarchive-3.5.1-1  libassuan-2.5.5-1  libcrypt-2.1-3  libcurl-7.76.1-1
[...]

Issue #2564, opened 2021-06-26
meld Maximum file size exceeded


sdbbs:
## Describe the issue

Cannot install meld

## Steps to Reproduce the Problem

```
$ pacman -S mingw-w64-x86_64-meld3
resolving dependencies...
looking for conflicting packages...
warning: dependency cycle detected:
warning: mingw-w64-x86_64-libwebp will be installed before its mingw-w64-x86_64-libtiff dependency
warning: dependency cycle detected:
warning: mingw-w64-x86_64-harfbuzz will be installed before its mingw-w64-x86_64-freetype dependency

Packages (68) mingw-w64-x86_64-adobe-source-code-pro-fonts-2.030ro+1.050it-2
...
              mingw-w64-x86_64-meld3-3.21.0-3

Total Installed Size:  563.43 MiB

:: Proceed with installation? [Y/n] y
:: Retrieving packages...
 mingw-w64-x86_64-meld3-3.21.0-3-any is up to date
error: failed retrieving file 'mingw-w64-x86_64-meld3-3.21.0-3-any.pkg.tar.zst?viasf=1.sig' from master.dl.sourceforge.net : Maximum file size exceeded
warning: failed to retrieve some files
error: failed to commit transaction (download library error)
Errors occurred, no packages were upgraded.
```

```
OS Name:                   Microsoft Windows 10 Home
OS Version:                10.0.19042 N/A Build 19042
OS Manufacturer:           Microsoft Corporation
```

lazka:
The sourceforge mirror is long gone, so closing this.

Issue #2178, opened 2020-10-07
Server downtime: Lessons learned


lazka:
Some things learned or things to do:

* We should try to not let the repo size grow indefinitely as that makes things hard to transfer/recover-> https://github.com/msys2/MSYS2-packages/issues/2174
* We should sign the installers, so they can safely be recovered from a third party mirror -> TODO
* We should make sure the sourceforge mirror is regularely/properly synced, or remove it
* We should document how to switch mirrors locally (using rankmirrors maybe)
* We should document the contact info for our existing mirrors

I've created a docker based server setup to quickly get a server running, though this isn't used right now because the server was already set up differently: https://github.com/msys2/msys2-repo-server Maybe it can be useful in the future.

Anything else?

1480c1:
Locally, I am just running

```bash
for list in mingw{32,64} msys; do
    cp /etc/pacman.d/mirrorlist.$list{,.orig}
    rankmirrors -r $list -v /etc/pacman.d/mirrorlist.$list.orig > /etc/pacman.d/mirrorlist.$list.ranked
    cp /etc/pacman.d/mirrorlist.$list{.ranked,}
done
```
for all of my msys2 installs so I get the best one either way

jeremyd2019:
I was thinking something ranking mirrors post-install would be helpful, but then I thought that might cause more issues/annoyance when pacman-mirrors is updated.

lazka:
@carrotIndustries are you the selfnet.de mirror contact? Seems rsync sync is broken, but I can't see why right now, maybe you can see some rsync logs or investigate this?

lazka:
hm, I think the file permissions are wrong

lygstate:
```
downloading mingw-w64-x86_64-python-setuptools-47.1.1-1-any.pkg.tar.zst...
downloading mingw-w64-x86_64-binutils-2.35.1-2-any.pkg.tar.zst...
downloading mingw-w64-x86_64-headers-git-8.0.0.6001.98dad1fe-1-any.pkg.tar.zst...
downloading mingw-w64-x86_64-crt-git-8.0.0.6001.98dad1fe-1-any.pkg.tar.zst...
downloading mingw-w64-x86_64-isl-0.22.1-1-any.pkg.tar.xz...
error: failed retrieving file 'mingw-w64-x86_64-windows-default-manifest-6.4-3-any.pkg.tar.xz' from sourceforge.net : Maximum file size exceeded
[...]

PR #2844, opened 2022-02-05
pacman-mirrors: remove sourceforge

lazka:
it's no longer in sync

### DIFF
--- a/pacman-mirrors/PKGBUILD
+++ b/pacman-mirrors/PKGBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Ray Donnelly <mingwandroid@gmail.com>
 
 pkgname=pacman-mirrors
-pkgver=20220126
+pkgver=20220205
 pkgrel=1
 pkgdesc="MSYS2 mirror list for use by pacman"
 arch=('any')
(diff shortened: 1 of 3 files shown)


cbrt64:
Re: #2564, #2575, #2841, https://github.com/msys2/MSYS2-packages/issues/2178#issuecomment-705824207, and https://github.com/msys2/MINGW-packages/issues/9183, among others, it occurs to me that there are some users who have customized repo configs (maybe really old ones with sf still at the top), so this PR by itself wouldn't do anything for them, and won't prevent more such reports.

Not sure what else could be done, just my two cents. In my case I only recently discovered `pacdiff` (from you on [gitter](https://gitter.im/msys2/msys2?at=6072f67fa2ac0d38e7ac6203), so thanks for that).

lazka:
> Re: #2564, #2575, #2841, [#2178 (comment)](https://github.com/msys2/MSYS2-packages/issues/2178#issuecomment-705824207), and [msys2/MINGW-packages#9183](https://github.com/msys2/MINGW-packages/issues/9183), among others, it occurs to me that there are some users who have customized repo configs (maybe really old ones with sf still at the top), so this PR by itself wouldn't do anything for them, and won't prevent more such reports.

Not that it helps, or we can do anything about it, but we have some stats for this: https://github.com/msys2/MSYS2-packages/discussions/2832 vs https://sourceforge.net/projects/msys2/files/REPOS/stats/timeline?dates=2021-02-06%20to%202022-02-05&period=daily

-> Sourceforge is ~0.5% of our current traffic.

> Not sure what else could be done, just my two cents. In my case I only recently discovered `pacdiff` (from you on [gitter](https://gitter.im/msys2/msys2?at=6072f67fa2ac0d38e7ac6203), so thanks for that).

In theory we made those mirror lists marked as "backup", so they sta around if modified only after we switched away from sourceforge, so all active users should have been migrated.

And, yes, it's a shame pacdiff isn't better integrated into the update flow, like in Debian for example.

cbrt64:
Ah, good point. Backups c51d60ccc10d15afe507e6eaf88fd82c1f6adb29 came before sf was bumped to Tier 2 3343a82cb9865cde32410d3e446249848c8fab77, but that's only two months so not nearly as bad as I was thinking. Thanks for the answer!

PR #2846, opened 2022-02-05
autoconf2.69: Add 0006-autoconf2.5-2.69-perl-5.22-autoscan.patch

stahta01:
This patch was copied from cygwin.

This patch fixes check so it finishes without error.

Tim S.


lazka:
thanks

PR #2847, opened 2022-02-06
Fix txt2tags dependency and grml-zsh-config empty man pages

cbrt64:
The `/usr/bin/txt2tags` stub uses `pkg_resources` from python-setuptools, which was present on the older build system.
So on CI, without python-setuptools by default, `txt2tags` fails, generating empty man files.
And building grml-zsh-config succeeds even when `txt2tags` fails (I'm working on an upstream PR to address their Makefile, though there's no need here to wait for that).

Arch Community has done [the same thing](https://github.com/archlinux/svntogit-community/commit/29a507d5eb40b48cacba77349a0d17657ff46eed) with txt2tags, maybe for the same reason? I'm mostly python-illiterate, if there's a better way to fix this I'm all ears.

This PR includes a grml-zsh-config update "for free".

### DIFF
--- a/grml-zsh-config/PKGBUILD
+++ b/grml-zsh-config/PKGBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Norbert Pfeiler <norbert.pfeiler+aur ät gmail.com>
 
 pkgname=grml-zsh-config
-pkgver=0.19.0
+pkgver=0.19.1
 pkgrel=1
 pkgdesc="grml's zsh setup"
 arch=('any')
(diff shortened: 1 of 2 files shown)


lazka:
thanks, I think upgrading setuptools and rebuilding txt2tags would also remove the pkg_resources

lazka:
uh, now /usr/bin/txt2tags is gone all together.. hm

lazka:
Updated to setuptools to 52.0.0 instead now, which works as before and is new enough for this.

Issue #2852, opened 2022-02-07
csh / tcsh crashes on start when used as login shell


pilonsi:
## Describe the issue

When msys2_shell.cmd is invoked to use csh or tcsh as login shell it crashes because it uses the --login argument which csh does not support. Changing it to -l instead which is supported by all shells fixes the issue. I have done some quick testing using bash, tcsh, zsh, ksh, fish and dash as login shells and all work fine after the change.

I have the fix I use in a local branch, I can submit a pull request if necessary: https://github.com/pilonsi/MSYS2-packages/commit/25d1eb38ab1b8450ee46108f65c79bf99ebbe5d6

## Steps to Reproduce the Problem

1. Run 'msys2_shell.cmd -shell tcsh -mingw64'

## Additional Context: Operating System, Screenshots

- OS: Windows 10 Pro 21H2

1480c1:
You might as well open a PR since you already have a commit and branch ready and discuss it there, the changes seem good to me either way

pilonsi:
Sure! This is fhe first time I report an issue on GitHub so I wasn't 100% sure on how to proceed.

PR #2853, opened 2022-02-08
Fix csh crashing when used as login shell

pilonsi:
Fix csh crashing when used as login shell. 

When msys2_shell.cmd is invoked to use csh or tcsh as login shell it crashes because it uses the `--login` argument which csh does not support. Changing it to `-l` instead which is supported by all shells fixes the issue. I have done some quick testing using bash, tcsh, zsh, ksh, fish and dash as login shells and all work fine after the change.

Follow-up to: https://github.com/msys2/MSYS2-packages/issues/2852


lazka:
* please add the description of the problem/fix from the issue to the commit message.
* bump pkgrel in PKGBUILD
* update checksums using `updpkgsums`

pilonsi:
Done! Thank you.

pilonsi:
I have squashed both commits into a single one with the issue description / fix as message.

PR #2854, opened 2022-02-08
xmlto: workaround -o argument handling with cygpath

elmarco:
Patch sent upstream for review:
https://pagure.io/xmlto/pull-request/10

### DIFF
--- a/xmlto/PKGBUILD
+++ b/xmlto/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=xmlto
 pkgver=0.0.28
-pkgrel=2
+pkgrel=3
 pkgdesc="Convert xml to many other formats"
 arch=('i686' 'x86_64')
 url="https://pagure.io/xmlto/"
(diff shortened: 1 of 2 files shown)


elmarco:
wrong patch, oops

xmlto/PKGBUILD
```diff
@@ -2,16 +2,23 @@
...
 makedepends=('docbook-xsl' 'autotools' 'gcc')
-source=("https://releases.pagure.org/${pkgname}/${pkgname}-${pkgver}.tar.bz2")
-sha256sums=('1130df3a7957eb9f6f0d29e4aa1c75732a7dfb6d639be013859b5c7ec5421276')
+source=("https://releases.pagure.org/${pkgname}/${pkgname}-${pkgver}.tar.bz2"
+        0001-xmlto-convert-output-path-to-unix-path-on-cygwin-msy.patch)
+sha256sums=('1130df3a7957eb9f6f0d29e4aa1c75732a7dfb6d639be013859b5c7ec5421276'
+            '981e200d9470852048df58fa7afbd02264b75d81b8e572c1aba3dc1487ad0f9d')
+
+prepare() {
+  cd "${srcdir}/${_realname}-${pkgver}"
```
mmuetzel, line 19: Should be `cd "${srcdir}/${pkgname}-${pkgver}"`?

lazka:
How/Why is something passing Windows paths to xmlto?

elmarco:
> How/Why is something passing Windows paths to xmlto?

This happens with meson:
```
custom_target(...
            command: [
                xmlto, 'man', '@INPUT@', '-o', meson.current_build_dir()
            ],
)
```

lazka:
I see. Imho adding hacks to every unix tool to not mangle Windows paths is a bit futile and paths should be converted on the non-cygwin side of things.

But this seems easy/manageable enough, so fine with me.

elmarco:
> I see. Imho adding hacks to every unix tool to not mangle Windows paths is a bit futile and paths should be converted on the non-cygwin side of things.
> 
> But this seems easy/manageable enough, so fine with me.

So meson shouldn't generate C:/ windows paths in the first place? Based on what environment should they make that decision?

lazka:
The meson build file could use cygpath when run on Windows, but yeah, mixing cygwin and native is always hacky..

PR #2859, opened 2022-02-12
Enable security keys in OpenSSH

brian-maloney:
I'd like to use [openssh-sk-winhello](https://github.com/tavrez/openssh-sk-winhello) with OpenSSH in my MSYS2 environment, but I've found that security keys are disabled in the build.  I can't find any documentation on why this is disabled, and I am able to successfully build OpenSSH and use FIDO2 security keys by removing this flag.

I'm not sure if something else needs to be added to get this in to a current build without changing ssh version, so please help advise me what else is needed to make this PR successful.

Thank you!

### DIFF
--- a/openssh/PKGBUILD
+++ b/openssh/PKGBUILD
@@ -45,7 +45,6 @@ build() {
     --with-libedit \
     --with-kerberos5=/usr \
     --without-hardening \
-    --disable-security-key \
     --disable-strip
 
   make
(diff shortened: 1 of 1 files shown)


brian-maloney:
Thanks @mmuetzel!  I bumped the `pkgrel` to `2`.  Seems like that's all that should be needed so this PR should be ready for review and merge now.

PR #2870, opened 2022-02-23
filesystem: Improve compatibility of '_warn_deprecated_winver'

ah64316:
'function' causes an error when I execute a shell with 'dash'; removing it seems to restore expected behavior (unable to test the actual function as I don't have win7 or win8)

### DIFF
--- a/filesystem/profile.000-msys2.sh
+++ b/filesystem/profile.000-msys2.sh
@@ -6,7 +6,7 @@ if [ ! "${MINGW_PREFIX}" = "" ]; then
 fi
 
 # Warn the user on the first login shell in case we detect a too old Windows version
-function _warn_deprecated_winver()
+_warn_deprecated_winver()
 {
     if [ "$__MSYS2_WINDOWS_VERSION_WARNING_DONE" = "true" ]; then
         return;
--- a/filesystem/profile.000-msys2.sh
+++ b/filesystem/profile.000-msys2.sh
@@ -15,8 +15,9 @@ _warn_deprecated_winver()
     winver="$(uname -s | sed -ne 's/\([^-]*\)-\([^-]*\).*/\2/p')"
     if [ "$winver" = "6.1" ] || [ "$winver" = "6.2" ]; then
         export __MSYS2_WINDOWS_VERSION_WARNING_DONE="true"
-        echo -e "\e[1;33mThe MSYS2 project is planning to drop active support of Windows 7 and 8.0 sometime during 2022.\e[1;0m" 1>&2;
-        echo -e "\e[1;33mFor more information visit https://www.msys2.org/docs/windows_support\e[1;0m" 1>&2;
+        printf "\e[1;33mThe MSYS2 project is planning to drop active support of Windows 7\e[1;0m\n" 1>&2;
+        printf "\e[1;33mand 8.0 sometime during 2022. For more information visit\e[1;0m\n" 1>&2;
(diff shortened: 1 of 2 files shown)


lazka:
thanks. just wondering, why do you use dash?

ah64316:
Couple of reasons, firstly, my company's security stuff is .. enthusiastic .. and spawning 'dash' is SIGNIFICANTLY faster than spawning 'bash'.  Ironically, most of my work IS actually run under a 'bash' shell but if I want to spawn off a shell to do something quick I don't want to wait the 30 - 60 seconds it takes to launch.  But I also keep by scripts compatible so tend to execute those under 'dash' to check my own work.  I suspect this has little to do with 'bash' spawn time and more to do with the scripts that bash executes when it spawns; unfortunately, time-poverty prevents me looking into it more thoroughly

lazka:
ok, thanks

filesystem/profile.000-msys2.sh
```diff
@@ -15,8 +15,9 @@ _warn_deprecated_winver()
     winver="$(uname -s | sed -ne 's/\([^-]*\)-\([^-]*\).*/\2/p')"
     if [ "$winver" = "6.1" ] || [ "$winver" = "6.2" ]; then
         export __MSYS2_WINDOWS_VERSION_WARNING_DONE="true"
-        echo -e "\e[1;33mThe MSYS2 project is planning to drop active support of Windows 7 and 8.0 sometime during 2022.\e[1;0m" 1>&2;
-        echo -e "\e[1;33mFor more information visit https://www.msys2.org/docs/windows_support\e[1;0m" 1>&2;
+        printf "\e[1;33mThe MSYS2 project is planning to drop active support of Windows 7\e[1;0m\n" 1>&2;
```
ah64316, line 18: I'm not really understanding why 'printf' is better for 'dash'.  Did you intend to use '/usr/bin/printf' to get proper processing of the escape sequences?

lazka: I ran it through shellcheck and it complained that dash doesn't support "-e". Do you have a better idea?

ah64316: Not familiar with spellcheck but yeah, dash doesn't seem to support `-e`.  I would personally remove the escape sequence for baseline stuff and simply print the message but you could either use `/bin/printf` to get the `\e` processing (which I'm guessing was your initial thought) or simply embed the escape sequence directly (e.g. use CTL-V-ESC in 'vi' to get the escape).

I updated my pull request (I think) with the embedded escape sequence; if that's not agreeable then I would simply replace the `printf` with `/bin/printf` or `/bin/echo -e` to bypass the builtins.

I'd also say, I don't personally have any issue with 'dash' being uglier (i.e. outputting the raw `-e \e[1;33mThe MSYS2 project is planning to drop active support of Windows 7 and 8.0 sometime during 2022.\e[1;0m`).

ah64316: PS: Sorry, I also changed the tests to a single test with `-o` rather than two tests.  Pedantic and probably only historically relevant (i.e. when `test` used to execute a `/bin/test`).  Ignore, as you see fit.

ah64316: Nope; don't think it updated the pull-request; it's on the branch though [..]

https://github.com/ah64316/MSYS2-packages/tree/patch-1

[..] although, in hindsight, if I had a project that was important I'd probably prefer to use `\e` rather than embedding non-printable characters.

PR #2875, opened 2022-02-27
openssh: add libcbor and libfido2 to allowed openssh support Windows Hello and FIDO2 security keys

SquallATF:


libcbor/PKGBUILD
```diff
@@ -0,0 +1,43 @@
...
+build() {
+  cd ${pkgname}-${pkgver}
+  cmake . \
+    -Bbuild \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_INSTALL_PREFIX=/usr \
+    -DCBOR_CUSTOM_ALLOC=ON \
+    -DBUILD_SHARED_LIBS=ON 
+  make -C build
+  make -C build DESTDIR=${srcdir}/dest install
```
1480c1, line 24: you may want to consider making these `cmake --build build` calls instead of make to adhere to a user set `CMAKE_GENERATOR`

libfido2/PKGBUILD
```diff
@@ -0,0 +1,70 @@
...
+  depends=("$pkgbase=$pkgver")
+  groups=("development")
+
+  mkdir -p ${pkgdir}/usr/lib
+  cp -rf ${srcdir}/dest/usr/include ${pkgdir}/usr/
+  cp -rf ${srcdir}/dest/usr/lib/pkgconfig ${pkgdir}/usr/lib/ 
+  cp -f ${srcdir}/dest/usr/lib/*.a ${pkgdir}/usr/lib
+}
+
+package_libfido2-docs() {
```
lazka, line 66: This needs `depends=()` or it will inherit it from the toplevel

lazka:
I remember in cygwin IRC that they patch openssh to make winhello work. So to get that to work we probably need to take those patches from them.

SquallATF:
> I remember in cygwin IRC that they patch openssh to make winhello work. So to get that to work we probably need to take those patches from them.

latest libfido2 already support winhello under cygwin, and  openssh 5.9 already support `fido_assert_set_clientdata` which winhello needed

lazka:
ok, thanks

PR #2887, opened 2022-03-16
python-pacdb: add package

jeremyd2019:
poetry conveniently created a setup.py in the source tar.gz


Biswa96:
provides/replaces/conflicts may not require for new python packages.

PR #2920, opened 2022-04-08
texinfo: Don't override MSYSTEM in check for path separator.

### LINKED ISSUE #2901: texi2dvi script cannot find tex (broken detection of path separator)
The `texi2dvi` script doesn't find a working tex installation, even though it is on PATH:

```
$ which tex
/c/Users/tomas/AppData/Local/Programs/MiKTeX/miktex/bin/x64/tex
$ texi2dvi nonexistent.texi
You don't have a working TeX binary (tex) installed anywhere in
your PATH, and texi2dvi cannot proceed without one.
...
```
The problem is caused by that `find_prog` fails to find `tex`, which in turn is because `path_sep` is detected to be `;` while in fact it is `:`. The problem is located here:

```
# Systems which define $COMSPEC or $ComSpec use semicolons to separate
# directories in TEXINPUTS -- except for Cygwin and Msys, where COMSPEC
# might be inherited, but : is used.

# In the case of Msys, uname returns a value derived from MSYSTEM, as
# MSYSTEM is user configurable, it is not so safe to use it to detect
# Msys. It is safer to use OSTYPE, this is why we set MSYSTEM to
# $OSTYPE before calling uname
if test -n "$COMSPEC$ComSpec" \
   && MSYSTEM=$OSTYPE uname | $EGREP -iv 'cygwin|msys' >/dev/null; then
  path_sep=";"
else
  path_sep=":"
fi
```
This code ends up calling `MSYSTEM=msys uname`, but, `uname` does not recognize "msys" (it does recognize "MSYS"), so the check above fails to detect Msys2.

```
$ MSYSTEM=MSYS uname
MSYS_NT-10.0-19044

$ MSYSTEM=msys uname
MINGW64_NT-10.0-19044

$ MSYSTEM=randomthing uname
MINGW64_NT-10.0-19044

$ echo $MSYSTEM
MSYS

$echo $OSTYPE
msys
```
There are work-around possible, including

```
$ OSTYPE=MSYS ./texi2dvi nonexistent.texi
$ COMSPEC= ./texi2dvi nonexistent.texi
./texi2dvi: cannot read ./nonexistent.texi, skipping.
```
But it would be very nice if it could be fixed.

mmuetzel:
Cherrypick upstream patch.

Fixes #2901.

### DIFF
--- a/texinfo/0006-msys2-OS-detection-in-texi2dvi.patch
+++ b/texinfo/0006-msys2-OS-detection-in-texi2dvi.patch
@@ -1,0 +1,31 @@
+From df780e22a5898df4a0f52ceb0ae3579ea7c78646 Mon Sep 17 00:00:00 2001
+From: Gavin Smith <gavinsmith0123@gmail.com>
+Date: Fri, 8 Apr 2022 17:27:00 +0100
+Subject: Msys2 OS detection in texi2dvi
+
+* util/texi2dvi: Do not set MSYSTEM via OSTYPE environment
+variable before calling uname, as this is not working on
+Msys2.  Reverses change on 2015-09-23.  Instead, check the value of
+OSTYPE directly as well as calling uname.
+
+Report and testing by Tomas Kalibera <tomas.kalibera@gmail.com>.
+
+diff --git a/util/texi2dvi b/util/texi2dvi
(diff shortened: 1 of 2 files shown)


lazka:
thanks

PR #2921, opened 2022-04-08
texi2dvi: backport, fix PATH separator detection to find tex

kalibera:
Without this fix, texi2dvi does not find tex on PATH, because it incorrectly detects the PATH separator. This is a backport of an upstream fix.

More at

https://github.com/msys2/MSYS2-packages/issues/2901
https://lists.gnu.org/archive/html/bug-texinfo/2022-04/msg00002.html


Biswa96:
This seems to be same as previous one.

kalibera:
Sorry, yes, Markus already did it. Thanks

PR #2934, opened 2022-04-17
libarchive: Upgrade to v3.6.1 and link with OpenSSL

elieux:
Previously it was linked with WinCrypto.


lazka:
Should we still do this, now that there is a fix for wincrypto?

lazka:
On a second thought, this also means upstream likely doesn't test much with wincrypto, so maybe this is better after all.

elieux:
I'm thinking we prefer unixy things in msys packages.

PR #2944, opened 2022-04-19
u-boot-tools: Update to v2022.04

lbmeng:
U-Boot v2022.04 host tools (mkeficapsule) introduced new build
dependencies to libgnutls-devel and libuuid-devel. No runtime
dependencies are needed as the mkeficapsule executable is for
testing purpose and we didn't ship it since day 1.

### DIFF
--- a/u-boot-tools/PKGBUILD
+++ b/u-boot-tools/PKGBUILD
@@ -1,17 +1,17 @@
 # Maintainer: Bin Meng <bmeng.cn@gmail.com>
 
 pkgname=u-boot-tools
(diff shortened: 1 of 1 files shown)


Biswa96:
Is `libgnutls-devel` required? The dependency list does not show gnutls DLLs.

lbmeng:
> Is `libgnutls-devel` required? The dependency list does not show gnutls DLLs.

Yes, it's required to build "mkeficapsule" exectuable in the upstream U-Boot, however we don't ship it in our package.

PR #2956, opened 2022-04-27
readline: avoid linking the static library to the ncurses DLL

dscho:
The ncurses headers [changes](https://github.com/ThomasDickey/ncurses-snapshots/commit/ba50399b8b9cc69cb56d8a51754c4366b2ae3f7e#diff-e468d12af06630d1a5e6bca47f23638abdcf3eccfda06a629387934c5196bbeaL71-L76), now defaulting to `dllimport` for its symbols.

Seeing as we definitely do not want to do that in the static version of libreadline, we now have to adjust our code. This is particularly important because we use that static library in Bash, where we definitely want to avoid linking to any MSYS DLL except `msys-2.0.dll`.

The symptom of an incorrectly-built library is this:

```console
$ nm /usr/lib/libreadline.a | grep tputs
                 U __imp_tputs
                 U __imp_tputs
```

In the correct case, the symbols would lack the `__imp_` prefix:

```console
$ nm /usr/lib/libreadline.a | grep tputs
                 U tputs
                 U tputs
```

### DIFF
--- a/readline/PKGBUILD
+++ b/readline/PKGBUILD
@@ -5,7 +5,7 @@ pkgname=('libreadline' 'libreadline-devel')
 _basever=8.1
 _patchlevel=002 #prepare for some patches
 pkgver=${_basever}.${_patchlevel}
-pkgrel=1
+pkgrel=2
 pkgdesc="GNU readline library"
 arch=('i686' 'x86_64')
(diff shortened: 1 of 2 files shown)


lazka:
Closing in favor of #2961, since that doesn't require patching and works similarly for all types of packages.

PR #2961, opened 2022-04-29
readline: set NCURSES_STATIC for the static build

lazka:
ncurses has started to default to dllimport by default, which means
NCURSES_STATIC has to be set by anyone trying to statically build against it
to get the right symbol names.

To keep things simple for us just build twice, once dynamically and
once statically with NCURSES_STATIC set.

Alternative to #2956

Also removes the "RPATH" part, which was copied from Arch, but from what I see
doesn't change anything for us.


lazka:
@dscho This is how we handled this in other packages so far.. not sure...

also see #2962

dscho:
Thank you!

PR #2963, opened 2022-05-01
pacman: enable clang32 by default

lazka:
And add commented out versions of clangarm64 and staging while at it,
so they are easy to enable if wanted.


eine (approved):
LGTM

PR #2967, opened 2022-05-06
msys2-runtime: Update to current HEAD of 3.3 branch

lazka:
https://github.com/msys2/msys2-runtime/tree/msys2-3_3-branch


lazka:
@dscho any thoughts on if MSYS2 should update to this too?

dscho:
> @dscho any thoughts on if MSYS2 should update to this too?

I thought that it might be prudent to wait for v3.3.5. Also note that we just discovered problems on Windows 7: https://github.com/git-for-windows/msys2-runtime/pull/43#issuecomment-1115920606 (also [reported to Cygwin](https://cygwin.com/pipermail/cygwin/2022-May/251370.html) who claim that this is a bug in our `readline` build, even if it is fine with the _current_ MSYS2 runtime...)

lazka:
yeah, I saw the cygwin thread. thanks, let's wait then.

lazka:
3.3.5 is imminent, so closing

PR #2983, opened 2022-05-12
filesystem: set `BASH_COMPLETION_COMPAT_DIR`

fauxpark:
Bash completions installed into `$MINGW_PREFIX/etc/bash_completion.d/`, eg. youtube-dl, do not get loaded because the script defaults to looking in `/etc/bash_completion.d/` instead.

### DIFF
--- a/filesystem/PKGBUILD
+++ b/filesystem/PKGBUILD
@@ -5,7 +5,7 @@
 
 pkgname=filesystem
 pkgver=2022.01
-pkgrel=4
+pkgrel=5
 pkgdesc='Base filesystem'
 arch=('i686' 'x86_64')
 license=('BSD')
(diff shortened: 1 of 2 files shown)


lazka:
Wouldn't this break completion for msys commands under /etc/bash_completion.d/ ?

fauxpark:
I don't think so, because it's only in effect if `MINGW_PREFIX` is set (if I have that if statement correct)?

lazka:
ok, that seems manageable.

I wonder though if we shouldn't adjust the youtube-dl package and move the file there to `.../share/bash-completion/completions` instead of the legacy directory.

fauxpark:
Perhaps it would be better to ask upstream whether it can be moved - I don't know whether there is a specific reason it is still installed to the legacy directory rather than lazily loaded.

PR #3036, opened 2022-06-07
Remove easyoptions-git package

lazka:
No users, upstream unmaintained for 7 years.

Because I'm cleaning up ruby r-deps.


lazka:
If anyone needs it feel free to create an issue and we can look into adding it back.

PR #3080, opened 2022-07-04
coreutils: "uname -o" -> Cygwin if MSYSTEM=CYGWIN

lazka:
If MSYSTEM=CYGWIN we output "Cygwin" as the operating system.
This is relevant in build scripts which call "uname -o" and
check for cygwin there.


dscho (approved):
Makes sense, when combined with the corresponding patch to the MSYS2 Runtime.

lazka:
thanks!

PR #3087, opened 2022-07-06
mintty: bump the default window size a bit

lazka:
80x24 is too small for me (and maybe others?) and it's not that obvious
that you can change the size in the settings, so maybe a larger default
helps new users.

### DIFF
--- a/mintty/001-default-size.patch
+++ b/mintty/001-default-size.patch
@@ -1,0 +1,13 @@
+--- mintty-3.6.1/src/config.c.orig	2022-04-24 16:35:43.000000000 +0200
++++ mintty-3.6.1/src/config.c	2022-07-06 14:52:46.761720300 +0200
+@@ -150,8 +150,8 @@
+   .allow_set_selection = false,
+   .selection_show_size = false,
+   // Window
+-  .cols = 80,
+-  .rows = 24,
++  .cols = 100,
++  .rows = 27,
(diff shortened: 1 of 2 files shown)


lazka:
image

Biswa96:
Would it be same if the columns and rows options are provided from launcher instead of patching mintty?

lazka:
> Would it be same if the columns and rows options are provided from launcher instead of patching mintty?

The the user wouldn't be able to change it in the mintty settings.

Biswa96:
> Then the user wouldn't be able to change it in the mintty settings.

How about adding a config file with the those settings? Assuming it won't replace the current one.

lazka:
I could add a /etc/minttyrc, but then any GUI changes will get written there instead of ~/.minttyrc, which means there will be a .pacnew file on every future mintty update.

I'm fine with both ways, but prefer the patch tbh.

Biswa96:
OK, understood.

1480c1:
I personally use something like 120x30 using
```bash
Columns=120
Rows=30
```
in the .minttyrc file

mintty:
80x24 is kind of a traditional standard terminal size.
My preference is 80x33 but I'd not make that a default.
Settings in $HOME are a proper place for personal preferences.
Anyway, mintty 3.7.3 will not save changes to /etc/minttyrc anymore.

PR #3133, opened 2022-08-02
Add unifdef package

jhol:
The `unifdef` utility selectively processes conditional C preprocessor `#if` and `#ifdef` directives. It removes from a file both the directives and the additional text that they delimit, while otherwise leaving the file alone.

### DIFF
--- a/unifdef/PKGBUILD
+++ b/unifdef/PKGBUILD
@@ -1,0 +1,23 @@
+# Maintainer: Joel Holdsworth <jholdsworth@nvidia.com>
+
+pkgname=unifdef
+pkgver=2.12
+pkgrel=1
+pkgdesc="Selectively processes conditional C preprocessor #if and #ifdef directives"
+arch=('any')
+url="https://dotat.at/prog/unifdef/"
+license=("BSD")
+makedepends=("gcc")
+source=("https://dotat.at/prog/${pkgname}/${pkgname}-${pkgver}.tar.xz")
+sha256sums=("43ce0f02ecdcdc723b2475575563ddb192e988c886d368260bc0a63aee3ac400")
+
+build() {
+  cd "${srcdir}"/${pkgname}-${pkgver}
+  make
+}
+
+package() {
+  cd "${srcdir}"/${pkgname}-${pkgver}
+  make DESTDIR="${pkgdir}" prefix=/usr install
+  install -Dm644 COPYING ${pkgdir}/usr/share/licenses/${pkgname}/COPYING
+}


unifdef/PKGBUILD
```diff
@@ -0,0 +1,23 @@
+# Maintainer: Joel Holdsworth <jholdsworth@nvidia.com>
+
+pkgname=unifdef
+pkgver=2.12
+pkgrel=1
+pkgdesc="Selectively processes conditional C preprocessor #if and #ifdef directives"
+arch=('any')
+url="https://dotat.at/prog/unifdef/"
+license=("BSD")
+makedepends=("gcc")
```
Biswa96, line 10:
```suggestion
makedepends=("${MINGW_PACKAGE_PREFIX}-cc")
```

unifdef/PKGBUILD
```diff
@@ -0,0 +1,23 @@
...
+sha256sums=("43ce0f02ecdcdc723b2475575563ddb192e988c886d368260bc0a63aee3ac400")
+
+build() {
+  cd "${srcdir}"/${pkgname}-${pkgver}
+  make CC=${CHOST}-gcc all
+}
+
+package() {
+  cd "${srcdir}"/${pkgname}-${pkgver}
+  make DESTDIR="${pkgdir}" prefix=/usr install
```
Biswa96, line 21:
```suggestion
  make DESTDIR="${pkgdir}" prefix=${MINGW_PREFIX} install
```

unifdef/PKGBUILD
```diff
@@ -0,0 +1,23 @@
...
+arch=('any')
+url="https://dotat.at/prog/unifdef/"
+license=("BSD")
+makedepends=("gcc")
+source=("https://dotat.at/prog/${pkgname}/${pkgname}-${pkgver}.tar.xz")
+sha256sums=("43ce0f02ecdcdc723b2475575563ddb192e988c886d368260bc0a63aee3ac400")
+
+build() {
+  cd "${srcdir}"/${pkgname}-${pkgver}
+  make CC=${CHOST}-gcc all
```
Biswa96, line 16: `CC` is already set from /etc/makepkg-mingw.conf.

```suggestion
  make
```

jhol:
> Is it possible to add the package in mingw-packages repository? The msys2-packages repository is kept at a minimum just with some build tools.

The problem is that unifdef currently contains POSIX-only code. If you try to compile it with MINGW, you get errors relating to missing UNIX features:

```
gcc -march=x86-64 -mtune=generic -O2 -pipe -pipe -o unifdef unifdef.c
In file included from unifdef.c:46:
unifdef.h:29:10: fatal error: err.h: No such file or directory
   29 | #include <err.h>
      |          ^~~~~~~
compilation terminated.
```

Biswa96:
There is a makefile in win32/Makefile.mingw directory and it compiles fine in mingw environment. I was wondering if that can be used instead.

PR #3183, opened 2022-09-11
Update pacman

lazka:
pkgbuild linting is now opt-in


lazka:
Forced PKGBUILD linting in CI now instead: https://github.com/msys2/MSYS2-packages/commit/6277194830 and https://github.com/msys2/MINGW-packages/commit/cc4b64c042eba4274fe587

PR #3194, opened 2022-09-20
libffi: update to 3.4.3 with fixes

h-c-mueller:
libffi-3.4.3 fixes e.g. the guile test-suite/standalone/test-ffi 'procedure->pointer and test-suite/standalone/test-loose-ends.c scm_procedure_to_pointer

MSYS2 specific patches have been included upstream.


Biswa96:
> MSYS2 specific patches have been included upstream.

FYI, those patches were not even required to be in upstream. msys2 specific patches are being remove one-by-one. One of the future plan is to be like cygwin, see https://www.msys2.org/wiki/Devtopics/

lazka:
there is still a lot to do until that is done though

h-c-mueller:
The MSYS2 specific patches typically address output from e.g. `./config.guess`, which is different from Cygwin. But the different `./config.guess` output shall in most cases result in the same behavior as for Cygwin.
So far, to my understanding, this (i.e. `./config.guess`) will not change and requires related patches in subsequent configuration files (e.g. configure.ac). Best to have these patch generalized, so that they can be accepted upstream.
Please correct me if I'm wrong, since this were my intentions so far.

lazka:
the goal is for `./config.guess` to have the same output as on cygwin by default.

PR #3201, opened 2022-09-30
guile: update to 3.0.8

h-c-mueller:
- activate check()
- only set configure options which are not set by default
- fix depends for libguile-devel
- add support for gdb guile auto-load


MehdiChinoune:
> * only set configure options which are not set by default

You know the rule `explicit is better than implicit`

h-c-mueller:
Upgrade to guile 3.0.8 changes API from 2.2 to 3.0. So dependent packages are required to rebuild. 

Dependent package seems only autogen. Probably only the so far two unused patches need to be adjusted and activated again:
autogen-5.18.12-allow-guile-2.2.patch
autogen-5.18.12-fix-guile-version-check.patch

h-c-mueller:
Upgrade to guile 3.0.8 changes API from 2.2 to 3.0. So dependent packages are required to rebuild.

Dependent package seems only autogen. Related pull request to enable 3.0 support is [f111bea](https://github.com/msys2/MSYS2-packages/pull/3202/commits/f111bea5a5ff5cca26f01273c0cbde740762075a).

PR #3222, opened 2022-10-11
makepkg: set _FORTIFY_SOURCE=2 by default

lazka:
This follows what Arch Linux does by default, see
https://github.com/archlinux/svntogit-packages/blob/091c8714716b40241f4259c8e6fdf99508944a2f/trunk/makepkg.conf#L42

This is now possible after https://github.com/msys2/MINGW-packages/issues/13401

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=pacman
 pkgver=6.0.1
-pkgrel=20
+pkgrel=21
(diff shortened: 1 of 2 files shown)


lazka:
There are some known issues for some packages:

* https://github.com/msys2/MINGW-packages/issues/13544
* https://github.com/msys2/MINGW-packages/issues/5803#issuecomment-1276784281

Let's wait with enabling by default a bit then

lazka:
There are no longer any known issues, so we could try this now

PR #3229, opened 2022-10-15
pacman: Set default arch to `x86-64-v2`

lhmouse:
This allows SSE4.2 and CMPXCHG16, the same with RHEL 9.

Reference: https://developers.redhat.com/blog/2021/01/05/building-red-hat-enterprise-linux-9-for-the-x86-64-v2-microarchitecture-level
Signed-off-by: LIU Hao <lh_mouse@126.com>

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=pacman
 pkgver=6.0.1
-pkgrel=18
+pkgrel=19
 pkgdesc="A library-based package manager with dependency support (MSYS2 port)"
 arch=('i686' 'x86_64')
 url="https://www.archlinux.org/pacman/"
@@ -70,7 +70,7 @@ validpgpkeys=('6645B0A8C7005E78DB1D7864F99FFE0FEAE999BD'  # Allan McRae <allan@a
 sha256sums=('SKIP'
             '26d141ead0b586e29ab6c49ffa45cf60eb2689f53f8e90c885ccd6d117e9ab67'
             'c12da01ede663a4924d0817a0d1bd6082b1380383cfb74cc1cea08f9d73e4902'
-            '1b8033d17f23eb00c17b6303d854c4c6eae17cb2f59542f37ddfb83528437d6b'
+            'bf9d1e0133b023fc28be291a1c5520c50cb0064cabc18d6ee95db7710873de7c'
             '98198e1f0f252eae0560d271bee4b9149e127399dd0d3fd5d8d24579d9e0550f'
             '04ac67a8f458a9daea532f87cedbab76f028ccf437069e3d24a583457603e61d'
             '6904ea154e451115c14246ded57737a5ab6c1151d1ac818c37dc849b70a5275f'
--- a/pacman/makepkg_mingw.conf
+++ b/pacman/makepkg_mingw.conf
@@ -42,8 +42,8 @@ if [[ "$MSYSTEM" == "MINGW64" ]]; then
   CC="gcc"
   CXX="g++"
   CPPFLAGS="-D__USE_MINGW_ANSI_STDIO=1"
-  CFLAGS="-march=x86-64 -mtune=generic -O2 -pipe"
-  CXXFLAGS="-march=x86-64 -mtune=generic -O2 -pipe"
+  CFLAGS="-march=core2 -mtune=generic -O2 -pipe"
+  CXXFLAGS="-march=core2 -mtune=generic -O2 -pipe"
   LDFLAGS="-pipe"
 elif [[ "$MSYSTEM" == "MINGW32" ]]; then
   CARCH="i686"
@@ -66,8 +66,8 @@ elif [[ "$MSYSTEM" == "CLANG64" ]]; then
   CC="clang"
   CXX="clang++"
   CPPFLAGS="-D__USE_MINGW_ANSI_STDIO=1"
-  CFLAGS="-march=x86-64 -mtune=generic -O2 -pipe"
-  CXXFLAGS="-march=x86-64 -mtune=generic -O2 -pipe"
+  CFLAGS="-march=core2 -mtune=generic -O2 -pipe"
+  CXXFLAGS="-march=core2 -mtune=generic -O2 -pipe"
   LDFLAGS="-pipe"
 elif [[ "$MSYSTEM" == "CLANG32" ]]; then
   CARCH="i686"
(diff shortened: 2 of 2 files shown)


jeremyd2019:
I am still keeping a couple of machines alive with `core2` (one generation older than `nehalem`, which `x86-64-v2` corresponds to on Intel).  They do run win10 without issue, even if they are not listed as supported.  If there's consensus that these machines are out of luck, that's fine, but wanted to point out that dropping win7 does not automatically mean these older machines can't run anymore

lhmouse:
Do you suggest `-march=core2` instead? Or something even older than that, such as `-march=nocona`?

jeremyd2019:
I don't know about AMD.  I've dealt with an AMD X2 something, that also could run win10, but don't know how that fits into those intel codenames.  Wikipedia says it has sse3 but doesn't mention ssse3

lazka:
I still think we should try to support the same hardware as Windows does, unless there is a good reason.

* With Linux distros users have a choice of using an older LTS for older hardware. With MSYS2 being rolling there is no such choice.
* There are downstream projects which distribute our binaries to Windows users and probably expect them to work every where.

Biswa96:
Would the result be same if a different version of mingw-w64-crt package is added with x86_64-v2?

mati865:
No, it affects all binaries/libraries.

lazka:
https://superuser.com/a/941175 has some info if we want to target 8.1

MehdiChinoune:
Some statistics:
[Steam Hardware Survey](https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam):

| CPU feature | % |
|--|--|
| SSE2 | 100.00 |
| SSE3 | 100.00 |
| LAHF/SAHF | 100.00 |
| CMPXCHG16B | 99.99% |
| SSSE3 | 99.50 |
| SSE4.1 | 99.24 |
| SSE4.2 | 98.95 |
| FCMOV | 97.64 |
| AES | 96.66 |
| AVX | 95.27 |
| AVX2 | 89.05 |
| HyperThreading | 77.12 |

[FireFox Data](https://data.firefox.com/dashboard/hardware)
in the GPU graph you could observe that most users have GPUs >= SandyBridge == Intel Core 2nd Gen

[Arch](https://gitlab.archlinux.org/archlinux/rfcs/-/merge_requests/2)
30-08-2021
image

29-01-2022
image

lhmouse:
The difference between `core2`, `nocona`, `nehalem` etc. is trivial; I don't prefer any of them. But, `x86-64` is different from `nocona` a lot because it doesn't allow `_InterlockedCompareExchange128()` (observed with `clang -fms-extensions` though, I think GCC behaves likewise on `__atomic_*` builtins).

lazka:
> [Arch](https://gitlab.archlinux.org/archlinux/rfcs/-/merge_requests/2)
> 30-08-2021

I tried to replicate the "openssl" benchmark used there (openssl speed rsa) with different march values but there is no difference at all: https://gist.github.com/lazka/722b926e1ffe04ba99da7f33c625ef2b

> The difference between `core2`, `nocona`, `nehalem` etc. is trivial; I don't prefer any of them. But, `x86-64` is different from `nocona` a lot because it doesn't allow `_InterlockedCompareExchange128()` (observed with `clang -fms-extensions` though, I think GCC behaves likewise on `__atomic_*` builtins).

Then I'd vote for `core2` as it matches win8.1/10 requirements.

lhmouse:
Thanks for the information. Updated.

MehdiChinoune:
> I tried to replicate the "openssl" benchmark used there (openssl speed rsa) with different march values but there is no difference at all: https://gist.github.com/lazka/722b926e1ffe04ba99da7f33c625ef2b

Maybe our OpenSSL is already optimized to choose the right CPU architecture at runtime, like OpenBLAS, svt-av1, dav1d, embree ... etc.

lazka:
> Maybe our OpenSSL is already optimized to choose the right CPU architecture at runtime, like OpenBLAS, svt-av1, dav1d, embree ... etc.

yeah, likely that or asm. I was just wondering since they state some improvement here: https://gitlab.archlinux.org/archlinux/rfcs/-/blob/69bda815a157b3593f514d93f961da15ab3c59c1/rfcs/0002-march.rst#L89

lazka:
> Thanks for the information. Updated.

the commit message needs updating too.

lhmouse:
> > Thanks for the information. Updated.
> 
> 
> 
> the commit message needs updating too.

Oops, updated too.

mati865:
> > I tried to replicate the "openssl" benchmark used there (openssl speed rsa) with different march values but there is no difference at all: [gist.github.com/lazka/722b926e1ffe04ba99da7f33c625ef2b](https://gist.github.com/lazka/722b926e1ffe04ba99da7f33c625ef2b)
> 
> Maybe our OpenSSL is already optimized to choose the right CPU architecture at runtime, like OpenBLAS, svt-av1, dav1d, embree ... etc.

OpenSSL doesn't have runtime SIMD features detection.

MehdiChinoune:
C-Ray benchmark:
```
$ wget http://www.phoronix.net/downloads/phoronix-test-suite/benchmark-files/c-ray-1.1.tar.gz
$ tar -xzf c-ray-1.1.tar.gz
$ cd c-ray-1.1
$ export RT_THREADS=$(($(nproc) * 16))
```
x86-64:
```
$ export CFLAGS="-march=x86-64 -mtune=generic"
$ cc -o c-ray-mt c-ray-mt.c -lpthread -O3 $CFLAGS
$ ./c-ray-mt.exe -t $RT_THREADS -s 3840x2160 -r 16 -i sphfract -o output.ppm
c-ray-mt v1.1
Rendering took: 235 seconds (235390 milliseconds)
```
x86-64-v2:
```
$ export CFLAGS="-march=x86-64-v2 -mtune=generic"
$ cc -o c-ray-mt c-ray-mt.c -lpthread -O3 $CFLAGS
$ ./c-ray-mt.exe -t $RT_THREADS -s 3840x2160 -r 16 -i sphfract -o output.ppm
c-ray-mt v1.1
Rendering took: 218 seconds (218813 milliseconds)
```

x86-64-v3:
```
$ export CFLAGS="-march=x86-64-v3 -mtune=generic"
$ cc -o c-ray-mt c-ray-mt.c -lpthread -O3 $CFLAGS
$ ./c-ray-mt.exe -t $RT_THREADS -s 3840x2160 -r 16 -i sphfract -o output.ppm
c-ray-mt v1.1
Rendering took: 159 seconds (159281 milliseconds)
```

jeremyd2019:
> CPU feature |	%
> --------------|----------
> SSE2 |	100.00
> SSE3 |	100.00
> LAHF/SAHF |	100.00
> CMPXCHG16B |	99.99%
> SSSE3 |	99.50

I think that is the cut-off of core2.  The only thing I'm questioning there is SSSE3 just because some AMD did not include that, but 99.5% having it is pretty darn good.

PR #3230, opened 2022-10-15
pacman: Do strip all symbols from EXEs and DLLs

lhmouse:
This has the effect of passing `--strip-all` instead of `--strip-unneeded` to `strip`, which can make installed executables a bit smaller.

For example, stripping all symbols from `/mingw64/bin/gcc.exe` can reduce its size from 2,351,500 to 2,344,448 (-7,052).

Microsoft documentation also suggests that there should be no symbol in PE files; see references about `PointerToSymbolTable` and `NumberOfSymbols`.

Reference: https://learn.microsoft.com/en-us/windows/win32/debug/pe-format
Signed-off-by: LIU Hao <lh_mouse@126.com>

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=pacman
 pkgver=6.0.1
-pkgrel=19
+pkgrel=20
 pkgdesc="A library-based package manager with dependency support (MSYS2 port)"
 arch=('i686' 'x86_64')
 url="https://www.archlinux.org/pacman/"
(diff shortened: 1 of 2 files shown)


mati865:
Is sacrificing some debugging possibilities worth 0.3% size decrease in case of GCC?
```bash
$ nm /clang64/bin/python.exe | wc -l
166

$ cp /clang64/bin/python.exe .

$ strip --strip-all python.exe

$ nm python.exe
python.exe: no symbols
```

lhmouse:
> Is sacrificing some debugging possibilities worth 0.3% size decrease in case of GCC?
> 
> 

Debug symbols are placed into `.debug_*` sections which are already stripped.

`nm` on current executables prints mostly symbols from the import table, which I think contributes too little to debugging. For example, what may `__imp_EnterCriticalSection` help? We can set a breakpoint on `EnterCriticalSection` instead.

mati865:
It can be used to quickly tell whether some features of other libraries (like jit, specific compression formats) are enabled (or at least linked). It's not a huge deal for me since I mostly use LLVM toolset which allows checking import table but IIRC Binutils doesn't have such tools.

lhmouse:
> It can be used to quickly tell whether some features of other libraries (like jit, specific compression formats) are enabled (or at least linked). It's not a huge deal for me since I mostly use LLVM toolset which allows checking import table but IIRC Binutils doesn't have such tools.

`objdump -p` can be used to decode and print the import table. There doesn't seem to be a filter option though, it is only possible to `grep` the result.

lazka:
Not sure if that's worth it. Does anyone know why Arch defaults to "--strip-unneeded" for libraries?

Biswa96:
Found an old discussion https://bugs.archlinux.org/task/13592

lhmouse:
The issue is not about `--strip-unneeded` itself. It's all that, by definition, Windows DLLs (dynamic _link_ libraries) are executables, not shared libraries, and they should match `STRIP_BINARIES`.

lhmouse:
I think I had better propose a change to 'strip.sh' instead.

lazka:
I see, thanks (still wondering why Arch does it).

I guess if we'd have some minimal documentation in either the MSYS2 or mingw-w64 docs for how to list the imported symbols, then this is fine with me.

lhmouse:
Closing this. I will create a PR against msys2-apcman instead.

lhmouse:
https://github.com/msys2/msys2-pacman/pull/20

PR #3232, opened 2022-10-16
pacman: also update -march for msys

lazka:


lazka:
let's wait for some more testing with mingw first..

PR #3234, opened 2022-10-16
makepkg: disable SSSE3 by default

lazka:
There are some old AMD CPUs which run Win10 (somehow) but don't have SSSE3 (not to be confused with SSE3), so explicitely disable it. Both gcc and clang seem to understand this combination.

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=pacman
 pkgver=6.0.1
-pkgrel=19
+pkgrel=20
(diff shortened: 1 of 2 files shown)


jeremyd2019:
SSSE3 was added by intel circa 2007 IIRC, but AMD didn't start supporting it until circa 2012 it looks like.

mati865:
How about setting base to AMD CPU instead?

lazka:
> How about setting base to AMD CPU instead?

That also enables things like `__3dNOW_A__`, `__3dNOW__`, `__SSE4A__`, `__ABM__` which intel CPUs never had from what I see.

PR #3239, opened 2022-10-20
makepkg: don't use "-mno-*" in CFLAGS

lazka:
In case a package has a configure check for say -mssse3 and then enables it if the compiler supports it it would be passed to the compiler before our CFLAGS, which currently has -mno-ssse3 (which seems to also disable sse4 and above while at it). The code then assumes it's available but the compiler fails.

While one could argue that we try to prevent ssse3+ being used in packages, there
are always exceptions like building sse4 code which dynamically gets executed.

So it's better we only enable machine options in CFLAGS and never disable them

Replace "-march=core2 -mno-ssse3" with the equivalent "-march=nocona -msahf". Note that the GCC docs wrongly state that "nocona" doesn't include "CX16", while it does.

See https://github.com/msys2/MINGW-packages/pull/13651

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=pacman
 pkgver=6.0.1
-pkgrel=21
+pkgrel=22
 pkgdesc="A library-based package manager with dependency support (MSYS2 port)"
 arch=('i686' 'x86_64')
 url="https://www.archlinux.org/pacman/"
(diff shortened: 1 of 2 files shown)


mmuetzel:
Could that cause similar issues if a project tries to set `-mno-sahf` and gcc adds `-msahf` after that?

jeremyd2019:
Potentially, but that seems much less likely than somebody trying to turn *on* features in certain files.

MehdiChinoune:
> Could that cause similar issues if a project tries to set `-mno-sahf` and makepkg adds `-msahf` after that?

I read once that Compilers prioritize the latest flag passed, I have verified it personally but I couldn't find docs for it.

Biswa96:
Was adding `-mnocona` set similar requirement for Windows 10? According to MS docs[^1], Windows 10 "Supports PAE, NX and SSE2". But it does not mention SSE3 requirement. Whereas GCC docs[^2] has, nocona supports "MMX, SSE, SSE2, SSE3 and FXSR instruction set". Should the flag be reverted to `pentium4` to remove the SSE3 requirement?

[^1]: https://learn.microsoft.com/en-us/windows-hardware/design/minimum/minimum-hardware-requirements-overview
[^2]: https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

MehdiChinoune:
Windows 10 reuirements:
```
• Compatible with the x86* or x64 instruction set.
• Supports PAE, NX and SSE2.
• Supports CMPXCHG16b, LAHF/SAHF, and PrefetchW for 64-bit OS installation
```

The requirement were made by searching for the earliest CPU from Intel and AMD that support the minimum requirements for *Windows 8.1*, according to this post https://superuser.com/questions/931742/windows-10-64-bit-requirements-does-my-cpu-support-cmpxchg16b-prefetchw-and-la/941175#941175
pentium4 doesn't support CX16 and SAHF.
Looking at https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html
the earliest CPU to support CX16 and SAHF was `core2` from Intel and `bdver1` from AMD.
`core2` should be the minimum without `-mno-ssse3`, because the earliest CPU from AMD that supports both CX16 and SAHF also support SSSE3.

Biswa96:
Understand. Thank you for the explanation.

PR #3259, opened 2022-10-27
filesystem: Set environment variables for pkg-config system paths.

mmuetzel:
The system include path and system library path are hard-coded into the pkg-config executables as UNIX style paths. That leads to those paths not matching literally in the MinGW environments where paths are converted to Windows paths. 

Also set environment variables with the same paths. That works because environment variables are automatically converted for MinGW executables.

I hope it is ok to add those environment variables for everyone.

Fixes https://github.com/msys2/MINGW-packages/issues/13691


lazka:
Interesting, thanks, looks good (minus pkgrel/checksum update). I guess making those defaults relocatable in pkgconf itself would also fix this.

Just checking with @1480c1 who uses pkgconf with a different prefix afaik, does this work for you?

mmuetzel:
> minus pkgrel/checksum update

Oops. Done now.



> I guess making those defaults relocatable in pkgconf itself would also fix this.

That would probably also work. Which functions could be used to convert those hard-coded UNIX paths to the corresponding Windows paths?

lazka:
> That would probably also work. Which functions could be used to convert those hard-coded UNIX paths to the corresponding Windows paths?

for the default pkgconf paths it uses something like: https://github.com/pkgconf/pkgconf/blob/master/libpkgconf/personality.c#L42

mmuetzel:
I'm not exactly sure how to easily apply the same for this case. We could probably query the location of the executable with the same trick. But appending `/../..` + the hard-coded paths would probably not match again.

E.g., the following still contains `-IC:/msys64/mingw64/include` (but it shouldn't):
```
$ PKG_CONFIG_SYSTEM_INCLUDE_PATH="C:/msys64/mingw64/bin/../../mingw64/include" pkgconf --cflags pango
-IC:/msys64/mingw64/include/pango-1.0 -IC:/msys64/mingw64/include/glib-2.0 -IC:/msys64/mingw64/lib/glib-2.0/include -IC:/msys64/mingw64/include -IC:/msys64/mingw64/include/harfbuzz -IC:/msys64/mingw64/include/freetype2 -IC:/msys64/mingw64/include/libpng16 -IC:/msys64/mingw64/include/fribidi -IC:/msys64/mingw64/include/cairo -IC:/msys64/mingw64/include/lzo -IC:/msys64/mingw64/include/pixman-1
```

1480c1:
> Just checking with @1480c1 who uses pkgconf with a different prefix afaik, does this work for you/do you see any problems?

Doesn't seem to affect my usecase, pkgconf returned the same with or without the env var set

lazka:
> I'm not exactly sure how to easily apply the same for this case. We could probably query the location of the executable with the same trick. But appending `/../..` + the hard-coded paths would probably not match again.

I see, you could cut twice by / and append though. Best to ask upstream I guess.

QBos07:
The added `C:/` seems to be problematic for CMake. I get lots of
```
CMake Error in CMakeLists.txt:
  Target "PkgConfig::libusb" contains relative path in its
  INTERFACE_INCLUDE_DIRECTORIES:

    "C:/msys64/ucrt64/include/libusb-1.0"
```

lazka:
That suggest that you are using cygwin cmake and not the native one

PR #3264, opened 2022-10-30
libtool: update to version 2.4.7

mmuetzel:
I built Octave on MINGW64 after upgrading to this version and it still seems to do its job.

### DIFF
--- a/libtool/PKGBUILD
+++ b/libtool/PKGBUILD
@@ -3,7 +3,7 @@
 pkgname=('libtool' 'libltdl')
 pkgver=2.4.7
 _gccver=11.2.0
-pkgrel=1
+pkgrel=2
 pkgdesc="A generic library support script"
 arch=('i686' 'x86_64')
 url="https://www.gnu.org/software/libtool/"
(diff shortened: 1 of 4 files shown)


libtool/PKGBUILD
```diff
@@ -65,7 +76,7 @@ build() {
 
 check() {
   cd ${srcdir}/${pkgbase}-${pkgver}
-  make check
+  # make check
```
Biswa96, line 79: This commenting is not required. One can run check function in their system.

mmuetzel: The check sometimes got stuck for me for some reason. (But that already happened with 2.4.6 iirc.) So, I commented it out. I'll revert that.

Biswa96: You can write it something like this `make check || warning "Tests failed"`. Or you can disable check() while building a package with `makepkg-mingw --nocheck` command.

mmuetzel: The issue wasn't that the checks were failing. But they were stuck for a very long time (maybe indefinitely). Maybe using `timeout 15m make check || warning "Tests failed"` would also have helped...

Biswa96: Opinions may vary. Feel free to change that command as you like. Just I would suggest not to disable/comment out it.

mmuetzel: Done.

mmuetzel:
I also have a patch that would allow linking to the Flang support libraries using libtool. It is kind of orthogonal to the version update.
Would it be ok to include it in this PR? Or better a separate PR?

mmuetzel:
I figured I'll just add that patch for Flang in this PR. Please, let me know if that's not a good idea and I'll revert.

The idea is similar to patch `0013-Allow-statically-linking-compiler-support-libraries-.patch` by @mstorsjo: Add an exception for the support libraries provided by the compilers.
But since the Flang support libraries are added with `-l` flags instead of a full path to the static support library, the exception needs to be added slightly differently.

lazka:
thanks!

Issue #3012, opened 2022-05-22
Move closer to Cygwin


lazka:
Because there hasn't been an issue about this, just chat in various places, a short summary:

We want to move closer to cygwin, in short msys, should no longer be a thing you target, just a slight variant of cygwin.

Done:

* libtool defaults to "msys-" prefixes even for cygwin, so we can build for cygwin #2779
* We are dropping various autotools build system patches by setting a cygwin target triplet, like #3004, #2982 etc.
* Python now pretends to be running on cygwin #2496 
* meson always only reported cygwin (initially patched, now inherited from Python)
* CMake was changed to pretend to be running under cygwin: #2422

To do:

* [ ] Remove all msysize patches
* [ ] Change the default triplet to cygwin
* [ ] Respect the CYGWIN env vars in the runtime as a fallback: https://github.com/msys2/msys2-runtime/issues/70
* [ ] Change the DLL prefix from "msys-" to "cyg"

Open questions:

* What about "uname"? `MSYS_NT-10.0-22000` vs `CYGWIN_NT-10.0-22000`
* `MSYSTEM=CYGWIN` ?

Overall Goals/Strategy:

* Share as much as possible with cygwin
* Make cygwin documentation apply to msys where possible
* In cases where possible and needed allow users to still detect msys in addition to cygwin

MehdiChinoune:
What about GCC and binutils?

mati865:
> What about GCC and binutils?

Less patching because of:
> Change the default triplet to cygwin

angelog0:
What would be the difference between MSYS2 an CYGWIN? Why an user should continue to use MSYS2 and not Cygwin which has many more packages and allows also to run X apps? Why do not rename MSYS2 to CYGWIN2, then? Would the only difference that MSYS2 uses pacman and Cygwin setup.exe? or you are thinking to reintroduce also setup.exe in place of pacman? Please, destroy something that works and is original..

NO, I do not concord with above choices.

mati865:
@angelog0 note this change affects only MSYS environment (`/usr`).
Our goal is still the same: provide users with means to build **native** applications/libraries.
This is sort of opposite of Cygwin where AFAIK everything runs with their POSIX emulation layer.

So for the end user everything related to installation and native compilation (all subsystems other that MSYS) remains the same. For maintainers however it means less effort to keep POSIX emulated things working.

MehdiChinoune:
[...]

PR #3285, opened 2022-11-08
openssh: Update to 9.1p1 / Sync with cygwin

lazka:
Remove some minor msysize patches while at it by faking cygwin Minor cleanup


dscho:
Is the `install-sh` patch really needed? I thought the [`autoreconf -fvi` call](https://github.com/msys2/MSYS2-packages/blob/075fe7f107c4bdda1271bae785506e2fc44a2a9c/openssh/PKGBUILD#L38) would take care of that.

dscho (approved):
Looks good, I particularly like the reduction of MSYS2-specific patches. Makes maintenance easier.

lazka:
> Is the `install-sh` patch really needed? I thought the [`autoreconf -fvi` call](https://github.com/msys2/MSYS2-packages/blob/075fe7f107c4bdda1271bae785506e2fc44a2a9c/openssh/PKGBUILD#L38) would take care of that.

good point, dropped (also checked that there is no difference)

> Looks good, I particularly like the reduction of MSYS2-specific patches. Makes maintenance easier.

part of #3012

PR #3299, opened 2022-11-14
cgdb: update to 0.8.0

RafaelDominiquini:


cgdb/PKGBUILD
```diff
@@ -1,7 +1,7 @@
 # Maintainer: Alexey Pavlov <alexpux@gmail.com>
 
 pkgname=cgdb
-pkgver=0.7.1
+pkgver=0.8.0
 pkgrel=3
```
Biswa96, line 5: pkgrel need to be reset to 1 when pkgver is changed.

PR #3306, opened 2022-11-20
msys2-runtime: apply workaround for missing trailing slashes with path_conv

lazka:
See https://github.com/msys2/msys2-runtime/pull/117


lazka:
fast-tracking this to unbreak the gcc build

PR #3338, opened 2022-12-02
texinfo: Update to 7.0.1; drop msysize patches

lazka:
see commits for details


mmuetzel:
I downloaded and installed the build artifact from this PR.
With it, I see some strange characters in the CLI of Octave for `help plot`.
Those characters are displayed correctly if I execute `system('chcp 65001')` at the Octave prompt first.

I'm using MINGW64 in the Windows Terminal App in case that should matter. It is set-up as described here:
https://www.msys2.org/docs/terminals/

The output of `locale` at the MINGW64 prompt in Windows terminal:
```
$ locale
LANG=de_DE.UTF-8
LC_CTYPE="de_DE.UTF-8"
LC_NUMERIC="de_DE.UTF-8"
LC_TIME="de_DE.UTF-8"
LC_COLLATE="de_DE.UTF-8"
LC_MONETARY="de_DE.UTF-8"
LC_MESSAGES="de_DE.UTF-8"
LC_ALL=
```

Is this an issue with texinfo, with octave, with msys2, with the Windows Terminal, with my local configuration or with something else?

lazka:
Thanks for testing. So I assume this is a regression compared to 6.8? and your Windows is set to German?

mmuetzel:
Yes, this is something that didn't occur with texinfo 6.8. But it might not be an issue with texinfo itself.
In texinfo 6.8, the single and double quotes were the ASCII ones. In texinfo 7.0, they seem to be using the "slanted" quotes (that look slightly different at the start and the end of a quoted string). Those are Unicode characters outside the ASCII range. texinfo 6.8 might have had the same issue with non-ASCII characters. They just didn't show up as often.

To maybe better illustrate what seems to be happening, here a (somewhat shortened) session with Octave after the update to texinfo 7.0 (see the symbols around `1:numel (Y)` shortly before I truncated the output):
```
octave:1> system('locale')
LANG=de_DE.UTF-8
LC_CTYPE="de_DE.UTF-8"
LC_NUMERIC="C"
LC_TIME="C"
LC_COLLATE="de_DE.UTF-8"
LC_MONETARY="de_DE.UTF-8"
LC_MESSAGES="de_DE.UTF-8"
LC_ALL=
ans = 0
octave:2> system('chcp')
Aktive Codepage: 850.
ans = 0
octave:3> help plot
'plot' is a function from the file C:\msys64\mingw64\share\octave\7.3.0\m\plot\draw\plot.m

 -- plot (Y)
 -- plot (X, Y)
 -- plot (X, Y, FMT)
 -- plot (..., PROPERTY, VALUE, ...)
 -- plot (X1, Y1, ..., XN, YN)
 -- plot (HAX, ...)
 -- H = plot (...)
     Produce 2-D plots.

     Many different combinations of arguments are possible.  The
     simplest form is

          plot (Y)

     where the argument is taken as the set of Y coordinates and the X
     coordinates are taken to be the range ÔÇÿ1:numel (Y)ÔÇÖ.


[truncated for shorter output]

octave:4> system('chcp 65001')
Aktive Codepage: 65001.
ans = 0
octave:5> help plot
'plot' is a function from the file C:\msys64\mingw64\share\octave\7.3.0\m\plot\draw\plot.m

 -- plot (Y)
 -- plot (X, Y)
 -- plot (X, Y, FMT)
 -- plot (..., PROPERTY, VALUE, ...)
 -- plot (X1, Y1, ..., XN, YN)
 -- plot (HAX, ...)
 -- H = plot (...)
     Produce 2-D plots.

     Many different combinations of arguments are possible.  The
     simplest form is

          plot (Y)

     where the argument is taken as the set of Y coordinates and the X
     coordinates are taken to be the range ‘1:numel (Y)’.

[truncated for shorter output]

octave:6>
```

Maybe, this is because the output from `locale` and the console codepage don't match (initially).
texinfo might be querying `LC_CTYPE` (or something similar) to determine whether it is running on an UTF-8 locale. That test would tell it that it *should* enable UTF-8 output.
However, the console codepage is still set to the system locale (German with codepage 805 in my case).

Is it the responsibility of the application (Octave in this case) to change the console codepage to match the locale (as returned by e.g. `setlocale(LC_CTYPE, NULL)`)? Or should the msys<->mingw interface be doing that somehow (when a native Windows application uses the Windows Console from inside a bash)?
(I don't know how that "translation layer" between msys vs. native applications is done. So, please, forgive me if this doesn't make much sense.)

lazka:
"native" apps calling cygwin processes isn't very common, so I doubt there is any good way, besides making the tool write to a temporary file (assuming the tool defines a proper encoding for this). The best strategy is probably to only use utf-8 when calling cygwin processes.

At one point there was talk if we could just force 65001 in the shell startup scripts, to avoid such incompatibilities.

mmuetzel:
I compiled Octave locally with the following changes:
``` diff
diff -r 3b7852a822e8 src/main-cli.cc
--- a/src/main-cli.cc	Fri Dec 02 10:12:40 2022 -0500
+++ b/src/main-cli.cc	Fri Dec 02 21:25:55 2022 +0100
@@ -36,6 +36,8 @@
 #  include <vector>
 #  include <locale>
 #  include <codecvt>
+#  include <windows.h>
+#  include <versionhelpers.h>
 #endif
 
 #include "liboctave-build-info.h"
@@ -105,6 +107,13 @@
     argv[i_arg] = &argv_str[i_arg][0];
   argv[argc] = nullptr;
 
+  if (IsWindows7OrGreater ())
+    {
+      // set console input and output codepage to UTF-8
+      SetConsoleCP (65001);
+      SetConsoleOutputCP (65001);
+    }
+
 #else
 int
 main (int argc, char **argv)
diff -r 3b7852a822e8 src/main-gui.cc
--- a/src/main-gui.cc	Fri Dec 02 10:12:40 2022 -0500
+++ b/src/main-gui.cc	Fri Dec 02 21:25:55 2022 +0100
@@ -36,6 +36,8 @@
 #  include <vector>
 #  include <locale>
 #  include <codecvt>
+#  include <windows.h>
+#  include <versionhelpers.h>
 #endif
 
 #include "liboctave-build-info.h"
@@ -118,6 +120,13 @@
     argv[i_arg] = &argv_str[i_arg][0];
   argv[argc] = nullptr;
 
+  if (IsWindows7OrGreater ())
+    {
+      // set console input and output codepage to UTF-8
+      SetConsoleCP (65001);
+      SetConsoleOutputCP (65001);
+    }
+
 #else
 int
 main (int argc, char **argv)
```

That avoids the issue with the garbled characters for me.
We could either apply this as a patch for Octave in MSYS2/MINGW. Or we can wait for Octave 8 for which I'll try to add that upstream.
Anyway, this shouldn't hold back this PR imho.

lazka:
Note that this change will persist after you exit octave, affecting commands run after it. It might still be the easiest fix though :)

lazka:
I just remembered that @Biswa96 also looked into a mingw build for texinfo recently.

mmuetzel:
> Note that this change will persist after you exit octave, affecting commands run after it in the same console.

Thanks for the pointer.
I pushed the following change to the Octave repository which resets to the previous input and output codepages after Octave has finished execution:
https://hg.savannah.gnu.org/hgweb/octave/rev/ea0b06534a37

It would probably be safe to cherry-pick this to Octave 7.3.0. I'll check and open a PR if that seems reasonable.

PR #3364, opened 2022-12-03
libxml2: remove indirect dependencies

lazka:
* libxml2/PKGBUILD:
  - remove the libraries 'ncurses' and 'ncurses-devel' from depends and makedepends, resp., since they are indirect dependencies to libxml2.
  - make autoreconf be verbose
  - remove trailing spaces


lazka:
taken from #1861

PR #3376, opened 2022-12-07
libtool: Allow spaces in program paths

zlaski:
This makes `libtool` more Windows-friendly.  Sometimes, `libtool` will find tools like _C:\Program Files (x86)\Git\bin\ld_ that it subsequently cannot handle.

### DIFF
--- a/libtool/0015-Allow-spaces-in-program-paths.patch
+++ b/libtool/0015-Allow-spaces-in-program-paths.patch
@@ -1,0 +1,60 @@
+diff -u a/bootstrap b/bootstrap
+--- a/bootstrap	2022-07-24 00:19:04.972571900 -0700
++++ b/bootstrap	2022-07-24 00:20:09.875444300 -0700
+@@ -369,8 +369,8 @@
+           _G_path_prog=$_G_dir/$_G_prog_name$_exeext
+           func_executable_p "$_G_path_prog" || continue
+           case `"$_G_path_prog" --version 2>&1` in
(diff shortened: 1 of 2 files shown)


lazka:
thanks

libtool/0016-Allow-spaces-in-program-paths.patch
```diff
@@ -0,0 +1,60 @@
+diff -u a/bootstrap b/bootstrap
+--- a/bootstrap	2022-07-24 00:19:04.972571900 -0700
++++ b/bootstrap	2022-07-24 00:20:09.875444300 -0700
+@@ -369,8 +369,8 @@
+           _G_path_prog=$_G_dir/$_G_prog_name$_exeext
+           func_executable_p "$_G_path_prog" || continue
+           case `"$_G_path_prog" --version 2>&1` in
+-            *GNU*) func_path_progs_result=$_G_path_prog _G_path_prog_found=: ;;
+-            *)     $_G_check_func $_G_path_prog
++            *GNU*) func_path_progs_result="$_G_path_prog" _G_path_prog_found=: ;;
```
jannick0, line 10: Assignments like `a=$b` do not need any quotation, since `bash` does respect any white spaces when expanding the variable in assignments.

This reduces the patch to quoted arguments which is required, of course.

I am wondering how backslashes in a directory are avoided or treated. This additional issue should be dealt with if reported upstream I believe.

PR #3385, opened 2022-12-10
tcl: Update to 8.6.12

lazka:


lazka:
I don't understand this tcl-sc.m4 vs tcl-tea.m4 split, so I did what Arch does and just ship the included one as tcl.m4.

I could be missing something, we'll see..

PR #3392, opened 2022-12-11
filesystem: Include mingw info path in INFOPATH variable

Biswa96:
See https://github.com/msys2/MINGW-packages/issues/631


lazka (approved):
lgtm

PR #3408, opened 2022-12-13
Add openssl-1.1 package

lazka:
This is parallel installable to openssl v3 and can be used for bootstrapping.


lazka:
Everything built fine, so this is no longer needed.

PR #3442, opened 2022-12-19
db: Update to 6.2.32

lazka:
Drop all patches, seems to build fine here, and java/tcl are disabled anyway.


lazka:
rebuils:

```
heimdal
cyrus-sasl
perl
```

and make a pacman rebuild to be sure

PR #3445, opened 2022-12-20
Add opensp

lazka:


lazka:
The goal here was to get openjade to build, but that seems FUBAR, so if this package ever makes a problem in the future and jas no reverse dependencies, then please delete it.

PR #3448, opened 2022-12-22
openssl: disable cleanup in DLL_THREAD_DETACH

lazka:
openssl/cygwin crashes when pthread APIs are called in DllMain:

0  0x0000000000000000 in ?? ()
1  0x00000001801619bd in pthread_rwlock_wrlock (rwlock=0x800275ff0)
   at /c/S/msys2-runtime-3.3/src/msys2-runtime/winsup/cygwin/thread.cc:3228
2  0x000000018019905b in _sigfe () at sigfe.s:35
3  0x00000004358490c9 in CRYPTO_THREAD_write_lock () from /usr/bin/msys-crypto-3.dll 4  0x000000043583c414 in init_thread_remove_handlers () from /usr/bin/msys-crypto-3.dll 5  0x000000043583c843 in OPENSSL_thread_stop () from /usr/bin/msys-crypto-3.dll 6  0x000000043583a953 in DllMain () from /usr/bin/msys-crypto-3.dll

It can be reliably reproduced by calling "meson test", which spawns threads for tests.

Something similar was discussed on the cygwin mailing list years ago, and it seems nothing has changed/improved there, and with openssl v3 it always crashes:

https://cygwin.com/pipermail/cygwin/2011-March/194454.html

Let's do what corinna suggested there, skip OPENSSL_thread_stop() under cygwin. This is a hacky workaround, but better to leak things then to crash.

### DIFF
--- a/openssl/0005-skip-dllmain-detach.patch
+++ b/openssl/0005-skip-dllmain-detach.patch
@@ -1,0 +1,24 @@
+--- openssl-3.0.7/crypto/dllmain.c.orig	2022-12-22 12:52:31.011614800 +0100
++++ openssl-3.0.7/crypto/dllmain.c	2022-12-22 12:52:32.210778100 +0100
+@@ -35,7 +35,9 @@
+     case DLL_THREAD_ATTACH:
+         break;
+     case DLL_THREAD_DETACH:
++# ifndef __CYGWIN__
+         OPENSSL_thread_stop();
++# endif
+         break;
+     case DLL_PROCESS_DETACH:
+         break;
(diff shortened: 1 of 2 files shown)


lazka:
Thanks!

With the stacktrace shown in https://github.com/curl/curl/issues/17262 I wonder if this is the right thing to do anyway, since the pthread exit handlers should get called in cygwin before the thread is cleaned up, which looks equivalent to OPENSSL_thread_stop(): https://github.com/openssl/openssl/blob/a0d1af6574ae6a0e3872d20ff302a78793c05a85/crypto/initthread.c#L204-L205

I haven't tested that though.

PR #3450, opened 2022-12-22
opensp: autoreconf

jeremyd2019:
For some reason, libtool was screwing up when I tried to build this for 32-bit msys2, even though it apparently built fine on 64-bit msys2. autoreconf to get updated (and presumably bugfixed) libtool files.


lazka:
we even get a .dll now, fancy :)

PR #3453, opened 2022-12-23
base-devel: remove groff

lazka:
Not many packages need it, so try to move it to explicit
makedepends instead.


lazka:
The build error is expected since base-devel gets uninstalled midway through.

PR #3459, opened 2022-12-29
sqlite: Update to 3.40.1

lazka:


lazka:
some tcl lib missing, needs some investigation

PR #3466, opened 2022-12-30
pacman: Enable clangarm64 repository

Biswa96:
/cc @driver1998

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=pacman
 pkgver=6.0.1
-pkgrel=27
+pkgrel=28
 pkgdesc="A library-based package manager with dependency support (MSYS2 port)"
 arch=('i686' 'x86_64')
 url="https://www.archlinux.org/pacman/"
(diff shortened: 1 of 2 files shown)


Biswa96:
The autobuild may need some modifications after adding this.

driver1998:
Cool, about time.

Shouldn't hurt on x86 machines if they don't install anything I guess.

Biswa96:
> Shouldn't hurt on x86 machines if they don't install anything I guess.

I install those for cross compilation but it's a rare case in general.

driver1998:
> Yeah, right. I install those for cross compilation but it's a rare case in general.

What is the recommended way to cross compile in MSYS2 btw, install llvm-mingw in MSYS2 or use jeremyd2019's wrapper package?

Biswa96:
> What is the recommended way to cross compile in MSYS2 btw

That would be different topic for this thread. Those two are same. I use the llvm-mingw one.

jeremyd2019:
> The autobuild may need some modifications after adding this.

Should be fine, the current sed used in autobuild and MINGW-packages gha should be idempotent, it would only uncomment the section if it was commented, otherwise it would do nothing.  Now, if it does nothing it could well be removed, but it shouldn't hurt anything in the meantime.

jeremyd2019:
> > What is the recommended way to cross compile in MSYS2 btw
> 
> That would be different topic for this thread. Those two are same. I use the llvm-mingw one.

Maybe MINGW-packages#8736 or a different discussion there.  They are *mostly* the same, but llvm-mingw uses its own build of what aims to be a 'vanilla' (unpatched) llvm/clang, while my package uses some of his wrappers to provide similar functionality using the llvm/clang already packaged by msys2 (which does not have the aim of being as unpatched/close to upstream as possible).  I think this is mostly a theoretical difference rather than there being any practical effects.

PR #3473, opened 2022-12-31
sqlite: Update to 3.40.1

jannick0:


lazka:
already done in #3459

jannick0:
@lazka Thanks - I came up with the same idea just now, but obvioulsy too late. Thanks again!!!

PR #3479, opened 2023-01-02
bisoncpp: Update to 6.04.44

jannick0:
* PKGBUILD:
  - use release tar instead of git tag
  - remove outdated dependency version requirements (not documented in package any more)
  - FIX BUG to handle symlinks when unzipping
  - remove html documentation already covered by man doc in package
  - additional edits

### DIFF
--- a/bisoncpp/PKGBUILD
+++ b/bisoncpp/PKGBUILD
@@ -2,38 +2,44 @@
 
 pkgbase=bisoncpp
 pkgname=bisonc++
-pkgver=6.04.00
+pkgver=6.04.04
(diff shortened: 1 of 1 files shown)


lazka:
Removal of `/usr/share/bisonc++/` and `/usr/share/doc/bisonc++-doc/` is on prupose?

jannick0:
Yes - both the html and the man doc is derived from the same doc source, so deliberate removal of the redundant html doc, since man doc is (still) included as for other packages.

lazka:
(Note that both directories contain demo code mainly)

PR #3484, opened 2023-01-11
filesystem: drop agm2 function

lazka:
This was added in e259adce916745c3934

(1) this isn't executed if bash_profile exists, which is the default and (2) I don't think we should include msys2 development related functions here


yashpalgoyal1304:
does it have anything to do with normal git install?

i have installed a fresh msys2, and [noticed this](https://github.com/search?q=org%3Amsys2+.profile&type=commits) while diff-ing to move over the dot files (the configs, history and such) to new msys2 install.

lazka:
What does your comment have to do with this PR?

PR #3491, opened 2023-01-12
filesystem: more cleanup

lazka:
Continuation of #3488

Goals:

* make startup faster
* remove things that shouldn't be needed anymore

current status: 230ms -> 130ms for `time bash -lc ""`


filesystem/06-windows-files.post
```diff
@@ -1,27 +1,26 @@
...
-    echo "If directory name is garbage you need to update your msys package" >&2
-    exit 1
-  fi
 
   local mketc=
   for mketc in ${FILES}
   do
     if [ ! -e "/etc/${mketc}" -a ! -L "/etc/${mketc}" ]
     then
+      local WINSYS32HOME="$(exec /usr/bin/cygpath -S -w)"
```
jeremyd2019, line 10: This is now inside a loop.  I understand wanting to not call it unless it is needed but maybe the value could be cached so it only has to be done once?

lazka: since it's only called on the first setup it didn't seem worth the extra logic.

lazka:
Some more ideas:

* move ~/.inputrc to /etc/inputrc, so it gets loaded even if there is no home dir and we can get rid of the restart warning (after cleaning up ~/.profile)
* Remote the lowercase tmp/temp hack, which I can't reproduce

filesystem/PKGBUILD
```diff
@@ -5,7 +5,7 @@
 
 pkgname=filesystem
 pkgver=2023.01
```
elieux, line 7: `2023.01.13`? :)

filesystem/01-devices.post
```diff
@@ -1,31 +1,6 @@
...
-  if [ ! -e "${DEVDIR}" ]
-  then
-    echo
-    echo "Creating ${DEVDIR} directory failed."
-    echo "Please fix that manually, otherwise you WILL get problems."
-    echo
-    exit 1
-  fi
 
   # Check for ${DEVDIR}/shm directory (for POSIX semaphores and POSIX shared mem)
```
elieux, line 5: Do we need the other things at all?

filesystem/08-xml-catalog.post
```diff
@@ -1,12 +0,0 @@
...
-  if [ ! -f "/etc/xml/catalog" ]
-  then
-    /usr/bin/mkdir -p /etc/xml
-    /usr/bin/xmlcatalog --noout --create /etc/xml/catalog
-    
-    MAYBE_FIRST_START=true
-  fi
-}
-
-maybe_create_xml_catalog
```
elieux, line 12: Replaced by a hook?

filesystem/01-devices.post
lazka: yes, named semaphores and shared memory need those directories from what I see. I was planing to add some integration tests for that.

Not sure what pacman would do if we add those empty dirs to "filesystem". I'll try.

filesystem/08-xml-catalog.post
lazka: https://github.com/msys2/MSYS2-packages/blob/master/libxml2/libxml2.install

filesystem/PKGBUILD
lazka: ok, sure

filesystem/01-devices.post
lazka: I've added two integration tests which fail if the directories are missing:

* https://github.com/msys2/msys2-tests/pull/29
* https://github.com/msys2/msys2-tests/pull/30

checking if they exist is fast since there are no forks involved, so I'd like to defer that to a future cleanup.

lazka:
@dscho In case this affects git-for-windows, see the separate commits for more context. There is no hurry.

filesystem/profile.000-msys2.sh
```diff
@@ -11,8 +11,11 @@ _warn_deprecated_winver()
     if [ "$__MSYS2_WINDOWS_VERSION_WARNING_DONE" = "true" ]; then
         return;
     fi
-    local winver;
-    winver="$(uname -s | sed -ne 's/\([^-]*\)-\([^-]*\).*/\2/p')"
+
+    local uname_s
+    uname_s="$(uname -s)"
+    local parts=(${uname_s//-/ })
+    local winver="${parts[1]}"
```
dscho, line 18: Wouldn't it be slightly easier to read as:

```sh
local winver
winver=$(uname -s)   # looks like `MINGW64_NT-10.0-22621`
winver=${winver#*-}  # strip off `<prefix>-`
winver=${winver%%-*} # strip off `-<suffix>`, if any
```

As a bonus, this would be pure POSIX shell (apart from the `local` construct).

dscho (approved):
Very nice!

lazka:
thanks for the reviews

PR #3492, opened 2023-01-12
ca-certificates: just call update-ca-trust in the install script

lazka:
currently the install script is more or less a copy of update-ca-trust, so to simplify things just call update-ca-trust in the install script

### DIFF
--- a/ca-certificates/PKGBUILD
+++ b/ca-certificates/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=ca-certificates
 pkgver=20211016
-pkgrel=2
+pkgrel=3
 pkgdesc='Common CA certificates'
 arch=('any')
 url='https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/'
--- a/ca-certificates/ca-certificates.install
+++ b/ca-certificates/ca-certificates.install
@@ -1,25 +1,7 @@
-export LC_ALL=C
-
 post_install() {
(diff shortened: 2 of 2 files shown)


jeremyd2019:
wonder why it was done that way to begin with... maybe something to do with when it's invoked as an install script?

lazka:
It was changed in https://github.com/msys2/MSYS2-packages/commit/9841d13dd8b019151e6ab1f6abb7dbfd6ce1b9f7

I don't know why though

Biswa96:
> It was changed in [9841d13](https://github.com/msys2/MSYS2-packages/commit/9841d13dd8b019151e6ab1f6abb7dbfd6ce1b9f7)

The rest of that commit suggests that something happen with /bin -> /usr/bin like changes.

lazka:
The only reason I can think of is pacman installing to a different root, but it still shouldn't make a difference due to chroot, see https://bbs.archlinux.org/viewtopic.php?pid=913891#p913891

jeremyd2019:
OK, the 'different root' install for making the installer was what I was most concerned about, since I remember this install script breaking in the past and causing issues.  Probably would cause more issues now than before, with more of the extraneous first run actions being removed from filesystem.

lazka:
I've added an optional input flag to the https://github.com/msys2/msys2-installer/actions/workflows/build.yml workflow, so we can test creating the installer from staging.

Issue #3187, opened 2022-09-12
ansible incompatible to python-jinja 3.1


olafmandel:
### Description

In `python-jinja` 3.1, the `environmentfilter` decorator got renamed to `pass_environment`, which is not taken into account in the two files `core.py` and `mathstuff.py` in the `/usr/lib/python3.10/site-packages/ansible/plugins/filter` directory. Fixes I can think of:

* either search&replace all instances of `environmentfilter` with `pass_environment`
* or change only the import lines in those files to something like `from jinja2.filters import pass_environment as environmentfilter`

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

Microsoft Windows [Version 10.0.19044.1889]

### Expected behavior

```
PLAY [Demo] ******************************************************************************************************

TASK [Saying hello] **********************************************************************************************
ok: [localhost] => {
    "msg": "Hello World"
}

PLAY RECAP *******************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0


### Actual behavior

```
PLAY [Demo] ******************************************************************************************************

TASK [Saying hello] **********************************************************************************************
[WARNING]: Skipping plugin (/usr/lib/python3.10/site-packages/ansible/plugins/filter/core.py) as it seems to be invalid: cannot import name 'environmentfilter' from 'jinja2.filters' (/usr/lib/python3.10/site-packages/jinja2/filters.py)
[WARNING]: Skipping plugin (/usr/lib/python3.10/site-packages/ansible/plugins/filter/mathstuff.py) as it seems to be invalid: cannot import name 'environmentfilter' from 'jinja2.filters' (/usr/lib/python3.10/site-packages/jinja2/filters.py)
ok: [localhost] => {
    "msg": "Hello World"
}

PLAY RECAP *******************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0

### Repro steps

```sh
$ cat <<EOF >demo.yml
- name: Demo
  hosts: localhost
  connection: local
  gather_facts: false
  vars:
    greetee: World
  tasks:
  - name: Saying hello
    debug:
      msg: 'Hello {{ greetee }}'
EOF
$ ansible-playbook -i localhost, test.yml
```

olafmandel:
Just checked:

[...]

PR #3494, opened 2023-01-13
update ansible to 2.14.1, add ansible community collections

alexjorgef:
With this, package `ansible-core` was added, which contains only the core logic. `ansible` package now contains the community collection library that depends on `ansible-core`.

When Ansible updated to the 3.0.0 version as they mention, the known ansible-base package moved to ansible-core (see reference: [Ansible 3.0.0 Q&A](https://www.ansible.com/blog/ansible-3.0.0-qa)).

> The former became known as ansible-base (soon to be [ansible-core](https://github.com/ansible/ansible/blob/devel/docs/docsite/rst/roadmap/ROADMAP_2_11.rst)). The latter became additions on top of the core, available either ad-hoc or as part of the Ansible community package, which includes a set of curated and maintained Collections.

Solves:

* https://github.com/msys2/MSYS2-packages/issues/3187

### DIFF
--- a/ansible-core/PKGBUILD
+++ b/ansible-core/PKGBUILD
@@ -20,10 +20,7 @@ sha512sums=('d3710aff1dbb04746b663b7c987c3597005dd8633df82f645312e20ddb9c679e731
 noextract=("${pkgname}-${pkgver}.tar.gz")
 
 prepare() {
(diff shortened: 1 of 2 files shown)


lazka:
thanks!

alexjorgef:
I'm glad I can help, I'll be more alert to new updates

PR #3514, opened 2023-01-24
update vim to v9

francescocorte9001:
Update vim to version 9.0.1172

I have installed this locally and tested it for a couple of days

I've also added the pkgdesc because why not


Biswa96:
Thank you for the contribution. Please reset `pkgrel` variable to 1.

PR #3516, opened 2023-01-25
less: Update to v621

dscho:
There are many commits in the main branch of `less` since the last version that was released "for general use" (v608).

Hoping that some of those commits fix issues reported in the Git for Windows bug tracker, let's update to the latest tagged version, v621.

An alternative would be to backport individual patches, but due to some sweeping changes (e.g. to convert from Kernighan & Ritchie style to C89 style function definitions), I've failed to backport them separately.

Switch to a GitHub archive since this release isn't available on the website yet.

### DIFF
--- a/less/PKGBUILD
+++ b/less/PKGBUILD
@@ -1,22 +1,28 @@
 # Maintainer: Alexey Pavlov <alexpux@gmail.com>
 
 pkgname=less
-pkgver=608
-pkgrel=2
+pkgver=621
+pkgrel=1
 pkgdesc="A terminal based program for viewing text files"
 license=('GPL3')
 arch=('i686' 'x86_64')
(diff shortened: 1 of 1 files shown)


Biswa96:
> ...fix issues reported in the Git for Windows bug tracker

could you share which bugs were fix with this update?

dscho:
> > ...fix issues reported in the Git for Windows bug tracker
> 
> could you share which bugs were fix with this update?

I hope that at least _part_ of https://github.com/git-for-windows/git/issues?q=is%3Aissue+is%3Aopen+pager are helped by this update. Unfortunately, it seems that I cannot quite rely on the bug reporters to verify fixes, so I am flying by night a little. In any case, this here PR should not hurt, but I hope that it helps.

PR #3517, opened 2023-01-28
pkgconf: Update to 1.9.4

lazka:


lazka:
For future reference, this update made https://github.com/msys2/MSYS2-packages/pull/3518 built, 1.8.0 fails. I haven't looked into why.

PR #3524, opened 2023-01-29
Update autotools-wrappers to 20221207

lazka:
Gentoo has now moved the wrappers into their own git repo and versions them in one.

To keep things simple, follow their lead and build both wrapper packages in one PKGBUILD.


autotools-wrappers/PKGBUILD
```diff
@@ -0,0 +1,54 @@
+# Maintainer: Christoph Reiter <reiter.christoph@gmail.com>
+
+pkgbase=autotools-wrappers
+pkgname=('automake-wrapper' 'autoconf-wrapper' )
+pkgver=20221207
+pkgrel=1
+pkgdesc="Wrapper scripts for autoconf & automake commands"
```
Biswa96, line 7: Redundant pkgdesc?

lazka: hm, true, thanks. I always forget that pkgbase doesn't really exist in the repo

PR #3528, opened 2023-01-31
pacman: sync -march flags from mingw to msys

lazka:
With win7/8 support dropped now we can bump our CPU requirements for msys packages.

This makes them match what we already use for mingw packages.


lazka:
@jeremyd2019 any objections? (I also bumped 32bit to pentium4, just to keep it in sync)

jeremyd2019:
no objections

PR #3538, opened 2023-02-08
Heimdal: Rebuild

HazardyKnusperkeks:
To link with OpenSSL 3.

Without my ssh, curl, and thus git didn't work anymore.

Took me literally hours to figure out which DLL still was looking for OpenSSL 1.1, because these seem to be loaded at runtime, not statically.

And rebuilding is also not quite easy when curl doesn't work...

### DIFF
--- a/heimdal/PKGBUILD
+++ b/heimdal/PKGBUILD
@@ -3,7 +3,7 @@
 pkgbase=heimdal
 pkgname=('heimdal' 'heimdal-libs' 'heimdal-devel')
 pkgver=7.8.0
-pkgrel=3
+pkgrel=4
 pkgdesc="Implementation of Kerberos V5 libraries"
 arch=('i686' 'x86_64')
 url="https://www.h5l.org/"


lazka:
It should already link against opensslv3.

Are you by any chance mixing git-for-windows with msys2?

HazardyKnusperkeks:
> It should already link against opensslv3.
> 
> Are you by any chance mixing git-for-windows with msys2? g4f is on openssl1.1, so mixing no longer works.

I was, I explicitly installed all (could it be that I missed heimdal?) packages from msys and then removed git for windows from the pacman.conf.

lazka:
Not sure. Maybe clear the cache `pacman -Scc` and then reinstall the packages `pacman -S heimdal-libs heimdal` just to be sure.

fauxpark:
> so mixing no longer works.

This is not true, I managed to get it working again after the OpenSSL upgrade by replacing most of the GfW packages (mainly curl, heimdal, SSL/SSH) with the MSYS2 ones, except for the runtime and some PCRE packages (not bothering to see what happens there). Git seems not to mind this. Only problem of course is now pacman complains that some of my local packages are newer than the repo.

Biswa96:
heimdal was rebuilt after openssl v3 update in this commit https://github.com/msys2/MSYS2-packages/commit/75bb29d4ff88fbcc64babcb74fbcac626c01b426

HazardyKnusperkeks:
So it seems I've missed to reinstall that and git for windows had the same version, hm? Sorry for the noise.

lazka:
No problem.

PR #3560, opened 2023-02-13
python: Update to 3.11

lazka:


lazka:
The errors don't make much sense, maybe the it's building in the wrong order

PR #3589, opened 2023-02-27
less: Update to v629

lazka:
helps with https://github.com/gwsw/less/issues/337


lazka:
This version is now also marked as beta upstream, which makes me more confident that we can update to it.

PR #3594, opened 2023-03-02
pacman: Update to 6.0.2

lazka:
See https://github.com/msys2/msys2-pacman/pull/28


Biswa96:
Q. In pacman/0010-Set-fixed-paths-for-various-tools.patch, some paths are /bin/gpg. Should those be /usr/bin/gpg?

lazka:
> Q. In pacman/0010-Set-fixed-paths-for-various-tools.patch, some paths are /bin/gpg. Should those be /usr/bin/gpg?

It shouldn't matter, but for consistency probably. There is also the question if it is needed at all, since we redirect all calls to /usr/bin anyway for gpg (see utils_fixed_path.sh.in)

lazka:
I've braindumped some potential cleanups on the patches here: https://github.com/msys2/msys2-pacman/issues/29

I'd like this update to focus just on the v6.0.2 changes.

PR #3674, opened 2023-03-20
gpgme: Update to 1.19.0

lazka:


lazka:
python is broken, needs some work

PR #3695, opened 2023-03-25
joe: Joe's own editor

dlmiles:
My first MSYS2 package contribution.

### DIFF
--- a/joe/0001-fix-build.patch
+++ b/joe/0001-fix-build.patch
@@ -1,0 +1,11 @@
+--- joe-4.6/joe/tty.c.orig	2018-01-10 17:28:34.000000000 +0100
++++ joe-4.6/joe/tty.c	2025-08-07 07:54:23.698861300 +0200
+@@ -952,7 +952,7 @@
+ 
+ /* Strange streams way */
+ 
+-extern char *ptsname();
++extern char *ptsname(int fd);
+ 
+ static char *getpty(int *ptyfd, int *ttyfd)
+ {
--- a/joe/PKGBUILD
+++ b/joe/PKGBUILD
@@ -1,0 +1,41 @@
+# Maintainer: Darryl L. Miles <darryl.miles@darrylmiles.org>
+
+pkgname=joe
+pkgver=4.6
+pkgrel=1
+pkgdesc="Joe's own editor"
+arch=('any')
+url="https://joe-editor.sourceforge.io/"
+license=('spdx:GPL-2.0-or-later')
+depends=('ncurses')
+makedepends=('make' 'gcc' 'autotools' 'ncurses-devel')
+backup=('etc/joe/ftyperc' 'etc/joe/jicerc.ru' 'etc/joe/jmacsrc'
+        'etc/joe/joerc' 'etc/joe/jpicorc' 'etc/joe/jstarrc' 'etc/joe/rjoerc')
+source=("https://downloads.sourceforge.net/joe-editor/${pkgname}-${pkgver}.tar.gz")
+sha256sums=('495a0a61f26404070fe8a719d80406dc7f337623788e445b92a9f6de512ab9de')
+
+build() {
+  cd "${srcdir}/${pkgname}-${pkgver}"
+
+  # Moved joe's configure data into /usr/share/joe otherwise it pollutes the space
+  ./configure --prefix=/usr --sysconfdir=/etc --datarootdir=/usr/share/joe --mandir=/usr/share/man
+
+  # This symlink substitute is problematic, disable it.
+  # LN_S = cp -pR
+  find -type f -name Makefile -exec sed -i "s/LN_S = .*/LN_S = echo/" {} \;
+
+  make
+}
+
+package() {
+  cd "${srcdir}/${pkgname}-${pkgver}"
+
+  make DESTDIR="${pkgdir}" install
+
+  # Replace LN_S function, it is aliasing alternative binary names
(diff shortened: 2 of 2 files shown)


dlmiles:
Just to report I have been using this for over a month and works just like native Linux console / xterm / SSH.

It is a tried and tested text editor that has been around longer than most with low maintenance overhead over the years.

Does the PKGBUILD file need a formal reviewed to promote it to having generally available status ?  It is not clear what the process is at this point.

There is not necessarily any rush I am just confirming there isn't an action from me that is being waited on in this process.

Once I have understood the process and I can then work on the next one (PKGBUILD for another package) to enhance an existing MSYS2 package with the availability of a performance improvement library.

dlmiles:
Tried the github sync button.  Not sure why it doesn't work out there is zero conflict to fast-forward and rebase on top.  All fixed now.  No wonder github projects are in a mess.

This is a force-push is a rebase on top of current master.

What is the limiting factor with progress on with this ?

Thanks

lazka:
> What is the limiting factor with progress on with this ?

We are trying to keep the number of msys packages low.

Depending on the type of package, the reason why it is needed, the complexity of the packaging/dependencies we might consider adding new packages.

Just curious, why do you need this package?

sskras:
@dlmiles commented [9 hours ago](https://github.com/msys2/MSYS2-packages/pull/3695#issuecomment-1562986097):

> Maybe the project FAQ instructions on msys2 website for 'contributing your first package' should have a process to establish which repository should be used.

I agree, saying that in the FAQ might help.  

Actually there are instructions about creating a new package, and these refers to the MINGW repository / space by offering the example / starting point: https://www.msys2.org/dev/new-package/

> It is not clear to me at this time what package space that would be. Do I just replicate and test it for all the combinations that exist ?

I got hit by that too.  Every time I was considering to use MSYS2 ago.  

It would be a lot more easier for a user if the FAQ or even the front page would contain some similar intro or at least the link to this page: https://www.msys2.org/wiki/How-does-MSYS2-differ-from-Cygwin/

To be short (even if subjective): 

- MSYS2 exists to make building MINGW applications easier.
- MINGW is an open source alternative to MSVC \
  (even if based on *nix-like shell and GCC).
- It allows to build native Win32 (Win64) apps.
- There is no place to add POSIX-based apps here.
- The exception is POSIX-based apps that are tools required for porting the app to Windows. \
  (eg. Bash, Libtool)

I guess the `joe` isn't a building dep for any of *nix-like apps.  And so it should belong to the MINGW (native Windows) world, not to the MSYS2.

The problem might be that `joe` [hasn't been ported to Windows](https://joe-editor.sourceforge.io/#:~:text=its%20only%20dependency%20is%20libc&text=use%20either%20the%20termcap%20or%20terminfo%20terminal%20capabilities%20databases) (WinAPI or msvcrt/ucrt).  So to put it into MINGW repo might require significant rewriting / porting effort from you (esp. the terminal I/O handling).

I am *nix-guy working on Windows, and this dichotomy sometimes makes me mad at MSYS2.  While package management on MSYS2 makes me happy.  Combined these polarizing effects puts quite a strain on me.

At the moment, I see only two easy (or at least additional) ways to solve that:

1. you start hosting your own Pacman repository with the MSYS2 builds of your software.
2. you start using some other system where POSIX-based apps are the first class citizens.

I know only of two active projects that goes the latter way:

  - Cygwin \
    (especially now that [MSYS2 is going to become its' plugin](https://www.msys2.org/wiki/Devtopics/#:~:text=games%2C%20image%20editors)-,Merge%20with%20Cygwin,-The%20MSYS2%20runtime) instead of being a separate fork)
  - [Midipix](https://midipix.org/#sec-midipix) (pre-alpha, this summer it's expected to get its' first release)

> This project is not expected to be (or become) a dependency for another project.

I guess, being a no dep makes the project even less needed for software building.
Although one might argue that the porter can only work with the given editor. :)

> I do have a library `libjudy` (C library API, no external dependencies) [...]. libjudy is expected to be a dependency for another (existing) project in the mingw-w64-x86_64 space. It is not clear if libjudy should be submitted here or over there.

IMO this clearly says to put it into MINGW repo.  Especially when it's supports [building with MSVC](https://github.com/netdata/libjudy/blob/777c9f4a8faf3f524d0afa39fb4577876b6b646d/src/build.bat#L4).

Just my €.02

lazka:
Yeah, I agree that the current approach for cygwin packages isn't communicated properly.

> I do have a library libjudy (C library API, no external dependencies) which has not been submitted to the process yet (as I am using this package joe as a process learning exercise). libjudy is expected to be a dependency for another (existing) project in the mingw-w64-x86_64 space. It is not clear if libjudy should be submitted here or over there.

I see, thanks.

libjudy seems to have Windows support, so it can/should go to MINGW-packages, see https://www.msys2.org/dev/new-package/ for how to get started there. libjudy seems abandoned though and from a quick peek at the source seems to assume MSVC on Windows, so some patching/work might be needed get it working.

dlmiles:
`libjudy` is feature complete.  Not really abandoned, a bit like `joe` LOL :)

To be abandoned it needs a lists of bug are that not being fixed.  However my tree currently has all the patches since the original maintainer/creator did any major work on it, specifically I think it is the only tree with a port to native Win64 (and  hopefully GCC/LLVM Win64).

I have a little more testing, with it directly and with the target application that uses it with an autoconf --with-judy option to improve memory usage and performance.

PR #3700, opened 2023-03-28
bison: enable relocatability

lb90:
See https://github.com/akimd/bison#relocatability

With that change I was able to create a small bundle with just bison, flex and bash and successfully compiled MESA w/ MSVC.

Bison tried to open paths in the form of `/usr/share/bison/...` and thus could not find the supporting files. After enabling that option everything worked. Maybe it can also be helpful for cases where MSYS2 is installed in paths other than `C:\msys64`

Feel free to merge or just close this PR :)

### DIFF
--- a/bison/PKGBUILD
+++ b/bison/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=bison
 pkgver=3.8.2
-pkgrel=4
+pkgrel=5
 pkgdesc="The GNU general-purpose parser generator"
 arch=('i686' 'x86_64')
 license=('spdx:GPL-3.0-or-later')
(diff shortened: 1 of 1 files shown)


lazka:
hm, I'm not sure if we want to support half broken cygwin installations.

> Maybe it can also be helpful for cases where MSYS2 is installed in paths other than `C:\msys64`

cygwin relocates paths by default, so that's not needed.

lb90:
> cygwin relocates paths by default, so that works already

Interesting! Out of curiosity, how does that work?

lazka:
> Interesting! Out of curiosity, how does that work?

Like everywhere else afaik, relative to the cygwin DLL. `cygpath -m /usr/share/bison/`

PR #3723, opened 2023-04-11
nsinstall: Add package

Biswa96:
Required for mingw nspr and nss package


lazka:
fine with me.

I got curious, and dug some more and found that Firefox has a Python version that it uses on Windows: https://hg.mozilla.org/mozilla-central/file/tip/config/nsinstall.py

It only has one reference to https://firefox-source-docs.mozilla.org/mozbase/mozfile.html#mozfile.remove which could be replaced with shutil.rmtree.

But feel free to ignore :)

Biswa96:
I am not familiar with python that much. If that allows to compile nspr and nss without makepkg, sure go for it.

PR #3728, opened 2023-04-15
tcl: split development files to separate package

Biswa96:
This is to remove any potential conflict with mingw tcl package

### DIFF
--- a/expect/PKGBUILD
+++ b/expect/PKGBUILD
@@ -4,13 +4,13 @@
 
 pkgname=expect
 pkgver=5.45.4
-pkgrel=4
+pkgrel=5
 pkgdesc="A tool for automating interactive applications"
 arch=('i686' 'x86_64')
 url="https://www.nist.gov/el/msid/expect.cfm"
 license=('custom')
 depends=('tcl')
-makedepends=('autotools' 'gcc')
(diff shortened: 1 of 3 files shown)


Biswa96:
Should dependent packages be rebuilt with new tcl-devel makedepends?

lazka:
> Should dependent packages be rebuilt with new tcl-devel makedepends?

expect and sqlite link against it, so they need "tcl-devel" in makedepends and a rebuild for good measure.

btw `/usr/share/man/man3/Tcl_*` also seems development related, but we could also add a tcl-doc in the future like debian does: https://packages.debian.org/sid/all/tcl8.6-doc/filelist Feel free to ignore though.

Biswa96:
> we could also add a tcl-doc in the future like debian does

I have added the tcl-doc package with all man3 files.

Another problem arises. expect can not be rebuilt, even with existing tcl package.

```
checking for Tcl public headers... /usr/include
checking for Tcl private include files... configure: error: Cannot find private header tclInt.h in /usr/include
```

lazka:
I'll have a look

Biswa96:
Thanks, now sqlite and expect are built without any issue.

PR #3747, opened 2023-04-28
mksh: Fix server address and ignore downloaded file

orgads:


lazka:
I tried to send an email there once because of the broken TLS setup.. but the email bounced because of broken TLS :)

Issue #3764, opened 2023-05-11
glib-compile-schemas failed on directory name with non Latin symbols


dantld:
**Observed**:
`glib-compile-schemas` failed on MS Windows platform under the MINGW64 environment when source directory, passed through the command line, contains the non Latin symbols.

Example below show the problem: calling `glib-compile-schemas` with non-Latin symbols in source directory.
```sh
$ glib-compile-schemas.exe /c/Shared/vista-client-service/Юзер

(glib-compile-schemas.exe:7476): GLib-CRITICAL **: 19:05:16.746: g_dir_open_with_errno: assertion 'wpath != NULL' failed
Error opening directory в??(NULL)в??: No error
```

Biswa96:
Please explain the issue. Also shouldn't you use the mingw glib2 package?

dantld:
> 



> Did you report the issue in upstream and suggest your changes there?

Yes I'm raise the issue and suggest the fixing.

Biswa96:
In that upstream repository, the fix should be provided as a merge request (same as pull request here).

dantld:
I've made mistake: this issue must be opened in the MINGW-packages.

PR #3765, opened 2023-05-11
DONE: #3764 Fixed glib-compile-schemas.

dantld:
Fixed: glib-compile-schemas failed when the source directory has non Latin symbols, under MNGW64 on MS Windows platform.

Root cause: using g_option_context_parse is incorrect under the MS Windows platform due the it expects system codepage instead of UTF-8.

Resolution: using g_option_context_parse_strv when G_OS_WIN32 is defined with g_win32_get_command_line.


dantld:
> Is there an upstream issue for this?

https://gitlab.gnome.org/GNOME/glib/-/issues/3003

dantld:
I've made mistake I need not MSYS but MINGW package so I'd close this pull request and reopen it in the MINGW-packages project.

PR #3770, opened 2023-05-18
curl: Update to 8.1.0

lazka:


lazka:
waiting on https://curl.se/mail/lib-2023-05/0029.html

PR #3809, opened 2023-05-26
sqlite 3.42.0: remove unnecessary dependencies from split packages

jannick0:
NB: The github action run on my fork hangs in step `dll check` for whatever reason. Let's see what happens here.

### DIFF
--- a/sqlite/PKGBUILD
+++ b/sqlite/PKGBUILD
@@ -9,7 +9,7 @@ pkgname=('sqlite' 'libsqlite' 'libsqlite-devel' 'sqlite-doc'
 _sqlite_year=2023
 _amalgamationver=3420000
 pkgver=3.42.0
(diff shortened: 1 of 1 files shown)


jannick0:
@lazka Any idea why the job hangs at [`dll check`](https://github.com/msys2/MSYS2-packages/actions/runs/5092158884/jobs/9153140650?pr=3809#step:10:927)? Anything within the .ci process? ... just wondering.

lazka:
hm, no idea

jannick0:
I'll try to check with other packages with the same minor issue tomorrow. Let's see ...

lazka:
I could be wrong, but I think `ldd` sometimes hangs with 100% CPU.

re the changes, lgtm. This makes us drop "tcl" from base, which is nice too :)

PR #3832, opened 2023-06-08
Add libev

lazka:
required by nghttp2 https://github.com/msys2/MSYS2-packages/pull/3831


lazka:
turns out it isn't needed

PR #3869, opened 2023-07-01
libuv: Update to 1.46.0

lazka:


lazka:
I noticed that cygwin now requires C99 since the last release. I've asked on IRC if that is on purpose.

Issue #3505, opened 2023-01-20
Replace libcrypt with libxcrypt


lazka:
We currently have https://packages.msys2.org/base/libcrypt

But it looks like Linux distros also had to move to a separate library because libcrypt was removed from glibc: https://wiki.gentoo.org/wiki/Project:Toolchain/libcrypt_implementation

We should look into moving to https://github.com/besser82/libxcrypt/ as well.

lazka:
package-grokker lists these users:

```
guile
cvs
heimdal
perl
openssh
ruby
apr-util
cyrus-sasl
```

PR #3875, opened 2023-07-05
Add libxcrypt

lazka:
See #3505

needs dev split and rebuild tests; static libs; is this needed for pacman indirectly?


lazka:
since only the "-devel" package conflicts it should be possible to merge this and migrate things over in smaller groups.

PR #3917, opened 2023-07-20
[CI] DLL Dependency Check - prohibit data chases for LDD

jannick0:
* .ci/ci-build.sh: run ldd separately for each binary


jannick0:
Just for the record: this PR amends the uncomfortable situation that `ldd` makes `CI` hang for, e.g., the package `sqlite`, in particular the split package `sqlite-extensions` with many DLLs with similar dependencies. `ldd` seems not to be able to cope with data races when `dlopen`ing a DLL `ldd` already opened for another DLL dependency check.

PR #3920, opened 2023-07-20
wiggle: add split package (version 1.3)

jannick0:


Biswa96:
Is it possible to build this as mingw package?

jannick0:
> Is it possible to build this as mingw package?

It seems to be a pure Unix package, so probably no, unless you want to cope with Windows porting issues.

PR #3923, opened 2023-07-21
bison 3.8.2: prefer configure argument over patch

jannick0:
* PKGBUILD:
  - remove references to gnulib-weak.patch
  - build: add 'gl_cv_have_weak=no' as argument to configure overriding the result of the relevant gnulib test
* gnulib-weak.patch: removed


lazka:
Just wondering, what's the purpose of enabling the configure cache?

jannick0:
> Just wondering, what's the purpose of enabling the configure cache?

The cache file contains the values of `configure`'s variables with `_cv_` in the variable names (`cv`= cached variable).  Without any time penalty to the `configure` run, this provides a nice immediate overview of most what is happening under `configure`'s hood, such that extra  `configure` runs might be time consuming and are not required any more if something goes wrong;  without the cache file it is rather tedious to get the value of one of `configure`'s variables without running `autoconf` and `configure` again. 

Just some tricks with the cache file without any significant harm to the packaging process of most, if not all, packages.

lazka:
I see, good to know, thanks!

PR #3940, opened 2023-08-04
binutils: Update to 2.41

lazka:


lazka:
Reverted for now since cygwin detected some regressions and reverted as well:

* https://sourceware.org/bugzilla/show_bug.cgi?id=30725
* https://sourceware.org/bugzilla/show_bug.cgi?id=30724

PR #3949, opened 2023-08-09
binutils: Update to 2.41 (v2)

lazka:
v1 was #3940 - this time with a backport from master


lazka:
confirming that things are as fast/slow as before

PR #3965, opened 2023-08-18
msys2-runtime: Update to 3.4.8

lazka:
See https://github.com/msys2/msys2-runtime/pull/168


lazka:
Needs to wait for https://github.com/msys2/msys2-runtime/pull/170

Issue #4011, opened 2023-09-05
Add Libstemmer package to Msys2


Zacharymk1213:
### Description / Steps to reproduce the issue

Libstemmer is needed to build libadwaita. This is needed for Gnome Text Editor

https://archlinux.org/packages/extra/x86_64/libstemmer/

### Expected behavior

Libstemmer would be in the repository.

### Actual behavior

Libstemmer isn't in the repository.

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

MehdiChinoune:
As explained in the PR, such packages are not allowed to be added to MSYS packages.

PR #4012, opened 2023-09-05
Create PKGBUILD

Zacharymk1213:
This addition relates to issue #4011. PKGBUILD is copied from the official Arch repositories.


MehdiChinoune:
1- Use a descriptive message for your commit and pull-request
2- Only tools required to build MinGW packages are allowed to be added to MSYS packages
3- snowball/libstemmer is already available as a MinGW package.

PR #4036, opened 2023-09-08
lcov: update to 2.0

MehdiChinoune:


lazka:
This broke lcov as all the new dependencies are missing

PR #4043, opened 2023-09-12
gcc: update to 13.2.0

MehdiChinoune:


lazka:
This breaks updates:

```
resolving dependencies...
:: installing gcc (13.2.0-1) breaks dependency 'gcc=11.3.0-4' required by gcc-fortran
```

I guess we need a `conflicts=('gcc-fortran<=11.3.0-4')` ?

lazka:
pushed to master

lazka:
Did some basic tests:

* pacman and binutils build fine
* msys2-runtime doesn't because of lots of new warnings errors which requires `-Wno-error=use-after-free` but our cross compiler is still at v10, so `'-Wno-error=use-after-free': no option '-Wuse-after-free'`

MehdiChinoune:
> * msys2-runtime doesn't because of lots of new warnings errors which requires `-Wno-error=use-after-free` but our cross compiler is still at v10, so `'-Wno-error=use-after-free': no option '-Wuse-after-free'`

I don't know autotools, but Isn't possible to check if a compiler supports a flag before using it? I think It's supported by autotools.

lazka:
> I don't know autotools, but Isn't possible to check if a compiler supports a flag before using it? I think It's supported by autotools.

I'll try to solve it with a small pragma patch, and then focus on getting our cross-gcc up-to-date (unless you want to?)

PR #4044, opened 2023-09-13
remove gcc-fortran

MehdiChinoune:


lazka:
What's the motivation here?

Arch has it in checkdepends.

MehdiChinoune:
> What's the motivation here?
> 
> Arch has it in checkdepends.

I have removed it from gcc in #4043 because gfortran is not needed on MSYS environment.

lazka:
Did it not build?

MehdiChinoune:
> Ah, I see, did it not build?

I said I do disabled Fortran because is not required.

lazka:
Alright, fair enough.

PR #4046, opened 2023-09-13
glib2: Update to 2.78.0

lazka:


lazka:
Not sure what's wrong, will have to bisect.

I've filed https://gitlab.gnome.org/GNOME/glib/-/issues/3111 for a different build error we also hit

lazka:
The second build issue is tracked here: https://gitlab.gnome.org/GNOME/glib/-/issues/3114

lazka:
Blocked on https://github.com/mesonbuild/meson/issues/12330 now

lazka:
This was fixed in #4420

PR #4051, opened 2023-09-15
cmake: Update to 3.27.5

lazka:


mmuetzel:
Fails with errors like the following:
```
  [ 72%] Linking CXX executable ../bin/cmake.exe
  [ 72%] Building CXX object Source/CMakeFiles/CTestLib.dir/CTest/cmCTestBinPacker.cxx.o
  /usr/lib/gcc/x86_64-pc-msys/13.2.0/../../../../x86_64-pc-msys/bin/ld: libCMakeLib.a(cmMakefile.cxx.o):cmMakefile.cxx:(.text+0x1e546): undefined reference to `Json::StreamWriterBuilder::operator[](std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
```
Maybe jsoncpp needs to be rebuilt with the updated GCC?

lazka:
yeah, that was also my guess

lazka:
I guess we are in a dep cycle now... meh

mmuetzel:
Is the existing cmake already broken? Or does it still work well enough to rebuild cppdab?

lazka:
once jsoncpp or cppdab is rebuild, cmake fails to run to build the other. I can either push a cmake build that vendors one of them for now, or bootstrap locally and manually upload.

lazka:
Here is a list of packages linking against libstdc++, which might cover all? https://github.com/lazka/package-grokker/actions/runs/6173574032/job/16756247566

mmuetzel:
Oof. That's a rather long list.
Let's hope the number of dependency cycles like this one is limited...

lazka:
Maybe it might make sense to introduce a cmake-bootstrap package in the future, with vendored deps, which is used to build cmake dependencies.

PR #4058, opened 2023-09-18
vim: only link to the perl DLL _dynamically_

dscho:
The major Perl version is encoded in the file name of the Perl DLL, which makes it much more desirable to link _dynamically_ to the DLL, i.e. on demand. Otherwise `vim.exe` would not even start unless Perl is installed, and the exact expected version at that.

However, due to a bug, even `--enable-perlinterop=dynamic` would let `vim.exe` link to the Perl DLL, defeating the purpose of the `dynamic` part.

This was fixed in the VIM project, via https://github.com/vim/vim/commit/55460da26c27, and here is a backport to get this fix into MSYS2.

This PR addresses https://github.com/git-for-windows/MSYS2-packages/pull/115#issuecomment-1722506060.

### DIFF
--- a/vim/0001-BACKPORT-dynamically-linking-perl-is-broken.patch
+++ b/vim/0001-BACKPORT-dynamically-linking-perl-is-broken.patch
@@ -1,0 +1,334 @@
+From 08391275e79154d723fa9daadc60298e7cee1c39 Mon Sep 17 00:00:00 2001
+From: Christian Brabandt <cb@256bit.org>
+Date: Tue, 29 Aug 2023 21:31:28 +0200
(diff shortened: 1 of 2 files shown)


dscho:
> our vim is a bit outdated right now since newer versions fail to build for a few months

I noticed, and will have a quick look at it now.

In the meantime, maybe we can go with this PR, as a band-aid until we get current VIM versions to build again?

dscho:
@lazka FWIW my current progress getting current VIM to build can be viewed in this here peanut gallery: https://github.com/msys2/MSYS2-packages/pull/4059

PR #4060, opened 2023-09-19
coreutils: update to 9.4

MehdiChinoune:
patches were copied from CygWin.


lazka:
Some fallout: https://github.com/msys2/MINGW-packages/pull/18565

I'm not sure yet what/why, but I can reproduce locally.

MehdiChinoune:
Feel free to revert it, I don't think I have enough knowledge and time to investigate and fix it.

lazka:
Fair enough.

Atm it's still the only package that has issues. We should try building the toolchains and the runtime though as well before releasing it from staging.

MehdiChinoune:
Packaging newlib for arm and riscv targets also fails.

lazka:
I think the problem is that `cygwin_spelling()` is called with destination relative paths in some places, but cygwin_spelling() doesn't deal with them.

I think we should (1) revert this for now (2) update to 9.0 instead which is actively used in cygwin, and is before coreutils switched to destination relative paths.

I've emailed Brian Inglis about my findings.

PR #4080, opened 2023-09-25
tar: disable check for year 2038 support on 32-bit.

### LINKED ISSUE #4078: [tar] 1.35 fails to build for 32-bit msys2
I know 32-bit msys2 is no longer supported...

```
checking for gcc option for timestamps after 2038... support not detected
configure: error: in `/t/r/tar/src/tar-1.35':
configure: error: could not enable timestamps after mid-January 2038.
This package recommends support for these later
timestamps. However, to proceed with signed 32-bit
time_t even though it will fail then, configure with
'--disable-year2038'.
```

Just wanted to record this somewhere.  I will look into whether there is some define to get 64-bit `time_t` on 32-bit msys2 (somehow I doubt there is if configure didn't find it), or if that configure option needs to be added for i686.

jeremyd2019:
32-bit cygwin (and thus msys2) only supports 32-bit time_t, so this configure check was failing there.  Yet another reason not to use 32-bit anymore.  This might also be relevant to @dscho in git-for-windows

Fixes #4078

### DIFF
--- a/tar/PKGBUILD
+++ b/tar/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=tar
 pkgver=1.35
-pkgrel=1
+pkgrel=2
 pkgdesc="Utility used to store, backup, and transport files"
 arch=('i686' 'x86_64')
 url="https://www.gnu.org/software/tar/tar.html"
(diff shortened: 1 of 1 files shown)


dscho:
Thank you for pinging me! I don't think we'll build a new `tar` package for git-sdk-32, though, so we're all good over here.

PR #4095, opened 2023-10-06
irssi: Update to 1.4.5

lazka:


sskras:
How come `irssi` is need for building software? =)

PR #4141, opened 2023-11-06
kicad7.0-library: Add package

stahta01:
Move kicad-library group of split packages except for meta to MSys2 repo.
Reason, these are data packages no reason for multiple copies of them.

Tim S.


stahta01:
Decided that this folder needs to be named kicad7.x-library; closing PR to rename it.

ognevnydemon:
> Decided that this folder needs to be named kicad7.x-library; closing PR to rename it.

you could commit file renaming

PR #4165, opened 2023-11-11
kicad7.x-library: Add package folder

stahta01:
This package folder provides common data libraries for a MINGW kicad package I am working on. Currently trying to get UCRT64 kicad to run, before doing the MINGW package PR.

### DIFF
--- a/kicad7.x-library/PKGBUILD
+++ b/kicad7.x-library/PKGBUILD
@@ -1,0 +1,129 @@
+# Contributor: Tim Stahlhut <stahta01@gmail.com>
+# These packages are data only and do not contain machine code.
+
+_enable_packages3D=yes
+_enable_templates=yes
+
+#_generator="Unix Makefiles"
+_generator="Ninja"
+
+_basename=kicad
+_majorpkgver=7
+_basepkgver=${_majorpkgver}.0
+_basepkgname=${_basename}${_basepkgver}
+_sharefoldername=${_basename}${_basepkgver}
+pkgbase=${_majorpkgver}.x-library
+pkgname=(
+  "${_basepkgname}-footprints"
+  "${_basepkgname}-symbols"
+  $( [[ ${_enable_templates} == "yes" ]] && echo "${_basepkgname}-templates")
+  $( [[ ${_enable_packages3D} == "yes" ]] && echo "${_basepkgname}-packages3D")
+)
+pkgdesc="Default description for libraries needed by mingw-w64 kicad application"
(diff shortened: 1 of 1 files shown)


mmuetzel:
Is there a reason why this package will be called `kicad7.x-library` instead of (an unversioned) `kicad-library`?

mmuetzel:
More general: I see your point in trying to reduce the installed data. But I don't know if the MSYS2 environment should be used as a data dump for the other (native) environments.

@lazka: What's your take on this?

stahta01:
> Is there a reason why this package will be called `kicad7.x-library` instead of (an unversioned) `kicad-library`?

My long term plan is to allow 2 or 3 version of kicad packages at once.
And, to use the largest of each major version.
As if the kicad people release an kicad 7.1 kicad7.x-library becomes 7.1 instead of the current 7.0.

And, I plan to add an kicad6.x-library that will be 6.0 or maybe 6.1.

Tim S.

stahta01:
> More general: I see your point in trying to reduce the installed data. But I don't know if the MSYS2 environment should be used as a data dump for the other (native) environments.
> 
> @lazka: What's your take on this?

I mostly agree; but, had no idea where else to put it. I could put it in a share folder under one of the MINGW location like MINGW64 or UCRT64.
Or, I could make an new location to put it; if a new location I would prefer someone else chose the name.

Tim S.

mmuetzel:
> I mostly agree; but, had no idea where else to put it. I could put it in a share folder under one of the MINGW location like MINGW64 or UCRT64.

I don't think the native environments should cross-reference like this either.
MSYS2 is sometimes also used as a base to create "re-distributable packages" for one software. (Think of "Git for Windows" or others.) It would be strange if they'd need to "mix" installations in different environments to get to a working software.

> Or, I could make an new location to put it; if a new location I would prefer someone else chose the name.

What is the main goal of this idea? Reduce the size of the packages on the MSYS2 mirrors? Or reduce the size of the installation on the user side?
If it is the latter: How many users actually install the same software in multiple environments (apart from testers or packagers)? (To be honest: I do that. But I'd count myself to the testers and packagers group.) A "normal" user of the software probably only installs it once in one single environment.
If it is the former: Do we need to worry about that?

lazka:
This package here at least is 385MB per env. Not sure that's enough to warrant anything special

stahta01:
> This package here at least is 385MB per env. Not sure that's enough to warrant anything special

Okay I will close this PR; I thought it was much bigger.

PR #4168, opened 2023-11-13
cross-gcc: Update to 13.2.0

lazka:
* sync in patches from cygwin gcc
* sync in important patches from mingw gcc not related to the build
* enabling stripping (not sure why it was disabled)
* reduce diff for build options to the mingw gcc build a bit


jeremyd2019:
> * enabling stripping (not sure why it was disabled)

building this for i686 msys2 results in a failure to build, eg, mingw-w64-cross-zlib for x86_64 with
```
/opt/lib/gcc/x86_64-w64-mingw32/13.2.0/../../../../x86_64-w64-mingw32/bin/ld: /opt/lib/gcc/x86_64-w64-mingw32/13.2.0/libgcc.a: error adding symbols: archive has no index; run ranlib to add one
```

I am guessing trying to use i686-pc-msys strip against x86_64 .a files somehow breaks the index.  I don't know why the other way (x86_64-pc-msys strip against i686 .a files) doesn't break anything.

lazka:
Oh, that might be why strip was disabled

lazka:
Btw, I was planning to add a ucrt variant too. Now that I think of msys32 I wonder how I can do that in a non hacky way.. Hm..

jeremyd2019:
Well, i686 is not supported, so there's no guarantee that every package (especially a new one) needs to be able to build there.  But I don't see why a ucrt-targetting cross-compiler would be any more of an issue on i686 than on x86_64.  Unless by "a non hacky way that works in CI" you just mean bootstrapping... That would need to be done on x86_64 also, but maybe you were just going to do that manually.

lazka:
Yeah, I meant bootstrapping onlt, I was building some things locally with msvcrt to break the cycles.

PR #4177, opened 2023-11-16
mingw-w64-cross-gcc: disable stripping again

jeremyd2019:
This was causing x86_64 libgcc.a to fail to link on an i686 msys host, apparently due to strip messing up the non-native libraries.  See https://github.com/msys2/MSYS2-packages/pull/4168#issuecomment-1813274876

### DIFF
--- a/mingw-w64-cross-gcc/PKGBUILD
+++ b/mingw-w64-cross-gcc/PKGBUILD
@@ -4,7 +4,7 @@ _realname=gcc
 _mingw_suff=mingw-w64-cross
 pkgname=("${_mingw_suff}-${_realname}")
 pkgver=13.2.0
-pkgrel=1
+pkgrel=2
 pkgdesc="Cross GCC for the MinGW-w64"
 arch=('i686' 'x86_64')
 url="https://gcc.gnu.org"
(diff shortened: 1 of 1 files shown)


jeremyd2019:
It seems kind of dangerous to run strip against 'foreign' binaries, cross-gcc is special among cross packages because it contains both host and target binaries.

lazka:
My thinking there is that we have integration tests for this: https://github.com/msys2/msys2-tests/blob/32bf984a018838d404ae3edc1183562de8a9f0dc/toolchain/test.sh#L11-L23

So if something is broken we hopefully notice and if not we should add a test.

lazka:
If that's "officially" not supported, then I'm fine with disabling.

jeremyd2019:
Quick check of mingw-w64-cross-* packages:
* mingw-w64-cross-binutils: no strip in options
* mingw-w64-cross-clang: !strip
* mingw-w64-cross-clang-crt: !strip
* mingw-w64-cross-clang-headers: !strip
* mingw-w64-cross-compiler-rt: !strip
* mingw-w64-cross-crt: !strip
* mingw-w64-cross-gcc: !strip only for i686
* mingw-w64-cross-headers: !strip
* mingw-w64-cross-tools: !strip
* mingw-w64-cross-ucrt-crt: !strip
* mingw-w64-cross-ucrt-headers: !strip
* mingw-w64-cross-windows-default-manifest: no options
* mingw-w64-cross-winpthreads: !strip
* mingw-w64-cross-winstorecompat: !strip
* mingw-w64-cross-zlib: no strip in options

I think binutils is only native binaries, so can strip or not.  Perhaps windows-default-manifest and zlib should also have !strip.  windows-default-manifest only has .o files, but zlib has both .a and .dll.a files.  I don't know if they are messed up on i686, but I will probably find out soon after I get cross-gcc and cross-zlib rebuilt.

lazka:
I've removed the !strip from zlib recently. hm, ok, if you really think this is tricky, those packages are probably not used that much to worth the trouble.

PR #4186, opened 2023-11-16
mingw-w64-cross: add x86_64-w64-mingw32ucrt

lazka:
This is a cross compiler targeting for 64bit ucrt. The target triple is the some Fedora decided to use: https://fedoraproject.org/wiki/Changes/F37MingwUCRT

### DIFF
--- a/mingw-w64-cross-binutils/PKGBUILD
+++ b/mingw-w64-cross-binutils/PKGBUILD
@@ -5,7 +5,7 @@ _realname=binutils
 _mingw_suff=mingw-w64-cross
 pkgname=("${_mingw_suff}-${_realname}")
 pkgver=2.41
-pkgrel=1
+pkgrel=2
 pkgdesc="A set of programs to assemble and manipulate binary and object files"
 arch=('i686' 'x86_64')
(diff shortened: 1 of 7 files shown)


lazka:
@jeremyd2019 There is a cycle between gcc and crt/winpthreads/default-manifest. I solved it by building the later ones with msvcrt locally, then building gcc, and then rebuilding the other packages again, which happened to work. Not sure what the expected path is there (??)

jeremyd2019:
I think that the way one is expected to start from scratch is to build some subset of gcc, then use that to build crt/winpthreads/windows-default-manifest, then build the rest of gcc.

I can build the bootstrapping packages if necessary, or even just use the x86_64 ones since they should just be target binaries (they could probably be `'all'` packages).

lazka:
> I think that the way one is expected to start from scratch is to build some subset of gcc, then use that to build crt/winpthreads/windows-default-manifest, then build the rest of gcc.

ok, that's probably not worth it at it point

> I can build the bootstrapping packages if necessary, or even just use the x86_64 ones since they should just be target binaries (they could probably be `'all'` packages).

true, let's see then

PR #4193, opened 2023-11-18
c-ares: Update to 1.22.1

lazka:


lazka:
some cmake configure checks fail. Not clear where the problem is yet.

PR #4199, opened 2023-11-22
pkgconf: Update to 2.0.3

lazka:


dg0yt:
We need 2.1.0 now!

dg0yt:
(All 2.0 releases are broken due to https://github.com/pkgconf/pkgconf/issues/322.)

MehdiChinoune:
> We need 2.1.0 now!

Are you using pkgconf or mingw-w64-pkgconf, I think vcpkg uses the latter

dg0yt:
> Are you using pkgconf or mingw-w64-pkgconf, I think vcpkg uses the latter

Good question. 
It uses the mingw versions for general PKGCONFIG purposes (e.g. via CMake). 
But it always carried msys pkg-config, now pkgconf, as a default package for the autotools build environment. And it works.

Issue #2282, opened 2020-12-27
Bad address on some path conversions


lazka:
```
#!/bin/bash
# this isn't supposed to work, just some parameters and calling a random mingw executable
/mingw64/bin/gcc.exe --foreign-types=C:/qt5/pkg/mingw-w64-x86_64-qt5,C:/qt5/pkg/mingw-w64-x86_64-qt5
```

```
$ bash a.sh
a.sh: line 2: /mingw64/bin/gcc.exe: Bad address
```

Looks like the path conversion code errors out, but it's erratic, so likely a memory corruption.

Adding `export MSYS2_ARG_CONV_EXCL='*'` makes it work

Possibly related: https://github.com/m-ab-s/media-autobuild_suite/issues/1584

dg0yt:
Still reproducable...
The error goes away if another command is called first. Even `:` is enough. The output is surprising:
~~~
gcc.exe: error: unrecognized command-line option '--foreign-types=C;D:\msys64\qt5\pkg\mingw-w64-x86_
64-qt5,C;D:\msys64\qt5\pkg\mingw-w64-x86_64-qt5'
gcc.exe: fatal error: no input files
compilation terminated.
~~~

dg0yt:
IIUC in `C:/qt5/pkg/mingw-w64-x86_64-qt5,C:/qt5/pkg/mingw-w64-x86_64-qt5`
we assume a mixed (`cypath -m`) path list, separated by `;`, but  
`bash`/MSYS2 assumes a Unix (`cypath -u`) path list, separated by `:`, transforming

`C` -> `C`
`/qt5/pkg/mingw-w64-x86_64-qt5,C` -> `D:\msys64\qt5\pkg\mingw-w64-x86_
64-qt5,C`
`/qt5/pkg/mingw-w64-x86_64-qt5` -> `D:\msys64\qt5\pkg\mingw-w64-x86_
64-qt5`

(given my MSYS2 installation in `D:\msys64`)

That is just the unexpected part of the successful conversion when calling MINGW64 from MSYS2.
It doesn't explain the "Bad address".

dg0yt:
> Adding `export MSYS2_ARG_CONV_EXCL='*'` makes it work

Any non-empty value makes it work...

dg0yt:
I have a working fix: Change
[...]

PR #4200, opened 2023-11-22
msys2-runtime: avoid "Bad address" errors in the MSYS2 path conversion

dscho:
This commit corresponds to https://github.com/msys2/msys2-runtime/pull/179 and should fix https://github.com/msys2/MSYS2-packages/issues/2282 and maybe even https://github.com/microsoft/vcpkg/pull/34274.

### DIFF
--- a/msys2-runtime/0039-fixup-Add-functionality-for-converting-UNIX-paths-in.patch
+++ b/msys2-runtime/0039-fixup-Add-functionality-for-converting-UNIX-paths-in.patch
@@ -1,0 +1,30 @@
+From 0e7d0755a911149d0c517c6b93887b3155a9cfe0 Mon Sep 17 00:00:00 2001
+From: Kai Pastor <dg0yt@darc.de>
+Date: Tue, 21 Nov 2023 09:24:03 +0100
+Subject: [PATCH 39/N] fixup! Add functionality for converting UNIX paths in
+ arguments and environment variables to Windows form for native Win32
+ applications.
+
+Don't memchr behind end, it+1
+---
+ winsup/cygwin/msys2_path_conv.cc | 7 +++++++
+ 1 file changed, 7 insertions(+)
+
+diff --git a/winsup/cygwin/msys2_path_conv.cc b/winsup/cygwin/msys2_path_conv.cc
+index 2bdf0ae..d4f0192 100644
+--- a/winsup/cygwin/msys2_path_conv.cc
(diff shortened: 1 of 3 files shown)


dg0yt:
Two more fixup patches here? Shouldn't the commits have better been squashed when merged to msys2-runtime? 
I'm planning to submit more changes to reduce the number of allocations... and I often keep individual changes as individual commits in one PR, for PR review purposes.

lazka:
The PKGBUILD is just a straight mirror of the active branch of msys2/msys2-runtime. In msys2/msys2-runtime we only squash when we rebase for a new version to not break the existing history.

dscho:
> Two more fixup patches here?

Yes. To allow for easier collaboration, we never force-push `msys2-<version>` branches in https://github.com/msys2/msys2-runtime.

And as @lazka pointed out, we only every update the patches in this repository's `msys2-runtime/` subdirectory via the `update-patches.sh` script (I always meant to add some automation that I could easily kick off from a merged msys2/msys2-runtime PR to create the corresponding msys2/MSYS2-packages PR, but never got around to doing that). The reason is that it is much easier to build confidence in the correctness that way.

> Shouldn't the commits have better been squashed when merged to msys2-runtime?

Yes! And we will do that when the next Cygwin version is released: Then, we will create a new `msys2-<new-version>` branch that points at the Cygwin revision, then rebase the commits from the previous `msys2-<version>` branch, push them as `tentative/msys2-<new-version>` and open a PR in msys2/msys2-runtime for cross-validation by other team members.

That rebase will be performed with `--auto-squash` and everything will get neat and tidy again.

> I'm planning to submit more changes to reduce the number of allocations...

Looking forward to them!

> and I often keep individual changes as individual commits in one PR, for PR review purposes.

Excellent. I like well-prepared PRs like that.

dscho:
I guess there is room for explanation in [msys2-runtime's `README`](https://github.com/msys2/msys2-runtime#readme)...

lazka:
It's documented here, somewhat: https://github.com/msys2/msys2-runtime/wiki

dg0yt:
Well, I really had to pull together pieces from different sources until I was able to arrive at testable packages.

lazka:
yeah, lots of room for improvements there

lazka:
It's in the pacman repo now

dg0yt:
vcpkg CI is happy :tada:

PR #4245, opened 2023-12-08
Update pacman

lazka:
Adds https://github.com/msys2/msys2-pacman/pull/33


lazka:
This also pulls in new curl and openssl afaics

PR #4297, opened 2023-12-23
autoconf: Add autoconf 2.72

lazka:
All the removed patches (compared to 2.71) are included in this new release

### DIFF
--- a/autoconf2.72/PKGBUILD
+++ b/autoconf2.72/PKGBUILD
@@ -1,0 +1,64 @@
+# Maintainer: Christoph Reiter <reiter.christoph@gmail.com>
+
+_realname=autoconf
+pkgname=${_realname}2.72
+pkgver=2.72
+pkgrel=1
+pkgdesc="A GNU tool for automatically configuring source code"
+arch=('any')
+license=('spdx:GPL-2.0-or-later' 'spdx:GPL-3.0-or-later' 'spdx:Autoconf-exception-3.0')
+url="https://www.gnu.org/software/autoconf"
+depends=('awk' 'm4' 'diffutils' 'bash' 'perl' 'sed')
+makedepends=('make')
+source=(https://ftp.gnu.org/pub/gnu/${_realname}/${_realname}-${pkgver}.tar.xz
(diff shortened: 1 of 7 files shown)


autoconf2.72/PKGBUILD
```diff
@@ -0,0 +1,66 @@
...
+  # license exception
+  install -Dm644 ${_realname}-${pkgver}/COPYING.EXCEPTION \
+    $pkgdir/usr/share/licenses/autoconf${pkgver}/COPYING.EXCEPTION
+
+  find ${pkgdir}/usr/share
+
+  # conflict with bintuils
+  rm -f ${pkgdir}/usr/share/info/standards.info
+  # conflicts with autoconf
+  rm -rf ${pkgdir}/usr/share/{emacs,info}
```
MehdiChinoune, line 65: Those lines are not needed
binutils doesn't have that file. and there is no package named autoconf.
Beside that they are suffixed there is no way they conflict with any other package.

lazka: we have three other autoconf packages. Do you mean we should have this files only in the newest version?

MehdiChinoune: Sorry, I was talking about man files which were not removed.
We could move those files for the latest autoconf. Yes that's my idea. No other package have them so why remove them.

lazka: fair enough, done

jannick0:
... and function CHECK runs through smoothly.

MehdiChinoune:
> ... and function CHECK runs through cleanly (without GO and FORTRAN being installed).

Are you asking or reporting?

jannick0:
> > ... and function CHECK runs through cleanly (without GO and FORTRAN being installed).
> 
> Are you asking or reporting?

That was an assertion, of course. :) 

A more general remark: What would be great is to make sure - for fundamental packages like `autoconf` as a minimum - that the package test suite runs through cleanly or with well-known failures not important on the platform.  

Updating packages is important, of course, which might be automatized somehow.  However a clean build is only the first step for generating a working package, but does not reduce the risk of bugs or failures as running the package test suite if available and monitoring the test results. 

See for a package not working for a couple of versions now: #4311.

PR #4321, opened 2024-01-01
add picotool package

thewh1teagle:
Add picotool package from 
https://github.com/raspberrypi/picotool
Inspired by picotool arch Linux package.

Statically linked libraries so users can use it outside msys too.
Final binary size is ~1.32MB

### DIFF
--- a/picotool/PKGBUILD
+++ b/picotool/PKGBUILD
@@ -1,0 +1,54 @@
+# Maintainer: Thewh1teagle
+
+_realname=picotool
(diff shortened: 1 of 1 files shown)


Biswa96:
I apologize for any inconvenience, but it appears that the pull request has been created in the incorrect repository. Additionally, the package recipe seems to be totally incorrect as well. Would it be possible for me to suggest adding the package to the mingw-packages repository?

thewh1teagle:
Sorry, you can delete this PR. thanks for help!

PR #4342, opened 2024-01-12
gsl: new package

orgads:


orgads:
Not everything can be built for MinGW. Specifically, sipp is not Windows compatible, and it has an optional dependency on gsl.

MehdiChinoune:
> Not everything can be built for MinGW. Specifically, sipp is not Windows compatible, and it has an optional dependency on gsl.

If you want sipp package then append this with it. I mean sipp with all of its dependencies in one PR.
Maintainers will take a look and see if It deserves to be accepted.

orgads:
ping? If you dislike it, I'll close it. Just don't leave it hanging here.

PR #4383, opened 2024-01-30
binutils: Update to 2.42

lazka:
* Remove backport which is included in the new release
* 0100-binutils-2.37-msys2.patch: simple refresh


piotrvv:
Binutils causes compilation error in package "msys2-runtime". DO NOT INSTALL

ognevny:
> Binutils causes compilation error in package "msys2-runtime". DO NOT INSTALL

open an issue about it

PR #4399, opened 2024-02-06
msys2-w32api-headers: work around `__buildreadseg` errors

dscho:
In msys2/msys2-runtime@ae5462a53b (CI: fix the build with gcc 13, 2023-11-10), we worked around build errors that look like this:

```
  In member function ‘virtual ssize_t fhandler_console::write(const void*, size_t)’:
  cc1plus: note: source object is likely at address zero
  In function ‘long long unsigned int __readgsqword(unsigned int)’,
      inlined from ‘_TEB* NtCurrentTeb()’ at /usr/include/w32api/winnt.h:10013:86,
      inlined from ‘tmp_pathbuf::~tmp_pathbuf()’ at winsup/cygwin/local_includes/tls_pbuf.h:21:5,
      inlined from ‘virtual ssize_t fhandler_console::write(const void*, size_t)’ at winsup/cygwin/fhandler/console.cc:4047:1:
  /usr/include/w32api/psdk_inc/intrin-impl.h:838:1: warning: array subscript 0 is outside array bounds of ‘long long unsig
  ned int [0]’ [-Warray-bounds=]
    838 | __buildreadseg(__readgsqword, unsigned __int64, "gs", "q")
        | ^~~~~~~~~~~~~~
```

This happens since the recent upgrade in MSYS2 from GCC 11.3.0-4 -> 13.2.0-2.

The work-around was to add the big hammer `-Wno-error`, which was later toned down to `-Wno-error=array-bounds`, among other more surgical options, in msys2/msys2-runtime@76fd7eae5a (fixup! CI: fix the build with gcc 13, 2023-12-22).

Helpfully, a work-around for the actual compile error was proposed in https://sourceforge.net/p/mingw-w64/mailman/message/37674508/ on the mingw-w64-public mailing list. Let's use that for now.

### DIFF
--- a/msys2-w32api-headers/0001-Work-around-Warray-bounds-warning-for-__readfsdword-.patch
+++ b/msys2-w32api-headers/0001-Work-around-Warray-bounds-warning-for-__readfsdword-.patch
@@ -1,0 +1,70 @@
+From 006dd2f19262c89ebbd0f4f2b920825783e7c107 Mon Sep 17 00:00:00 2001
+From: LIU Hao <lh_mouse@126.com>
+Date: Wed, 29 Jun 2022 07:38:24 -0700
+Subject: [PATCH] Work around -Warray-bounds warning for __readfsdword and
+ __readgsqword
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+在 2022-06-29 19:51, Ozkan Sezer 写道:
+
+    Noticed this in github CI runs of vkQuake. Is the warning bogus?
+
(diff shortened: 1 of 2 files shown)


lazka:
For future reference:

```
05:23 <msys2-on-discord> <biswa96> Would it be possible to add this patch in mingw-w64 https://github.com/msys2/MSYS2-packages/pull/4399 ?
06:08 <lh_ideapad> to be honest I don't think it's an 'appropriate' workaround....
06:09 <lh_ideapad> and, as the message suggests, it causes errors when compiled with Clang.
```

So we shouldn't backport this to the other mingw-w64 packages.

PR #4412, opened 2024-02-12
filesystem: Use working directory only if it is applicable

orgads:
Check for both msys-2.0.dll and bash.exe. Otherwise, use the script directory. Some applications install msys2 runtime as a dependency, while it is not a full msys environment.

### DIFF
--- a/filesystem/msys2_shell.cmd
+++ b/filesystem/msys2_shell.cmd
@@ -2,8 +2,10 @@
 setlocal EnableDelayedExpansion
 
 set "WD=%__CD__%"
-if NOT EXIST "%WD%msys-2.0.dll" set "WD=%~dp0usr\bin\"
 set "LOGINSHELL=bash"
+if EXIST "%WD%msys-2.0.dll" if EXIST "%WD%%LOGINSHELL%.exe" goto use_cwd
+set "WD=%~dp0usr\bin\"
+:use_cwd
 set /a msys2_shiftCounter=0
 
 rem To activate windows native symlinks uncomment next line


Biswa96:
What issue does this fix?

orgads:
I tried to state it in the commit message. I guess it was not enough.

When I run `C:/msys64/msys2_shell.cmd -mingw64 -defterm -no-start` from a directory that has `msys-2.0.dll`, but doesn't have `bash.exe`, it fails with this error:
```
'"F:\Utils\\bash"' is not recognized as an internal or external command,
operable program or batch file.
```

Biswa96:
What does the directory have? What do you want to do with the msys2_shell.cmd if that directory does not have any shell?

orgads:
I just want to start a shell, and I happened to run from this directory.

It has a utility that is built with msys, so I need the msys runtime there.

orgads:
What's the verdict?

orgads:
@Biswa96?

Biswa96:
I can not understand the situation clearly. Please wait for other to comment :pray:

orgads:
I'll try to explain with an example:

* Build sipp with msys
* Copy sipp.exe and all its DLL dependencies (including msys-2.0.dll) to C:\sipp
* cd /d C:\sipp
* Run C:\msys64\msys2_shell.cmd -defterm -no-start

```
C:\sipp>C:\msys64\msys2_shell.cmd -defterm -no-start
'"C:\sipp\\bash"' is not recognized as an internal or external command,
operable program or batch file.
```

orgads:
ping? @Biswa96 / @lazka

PR #4419, opened 2024-02-15
tmux: add --enable-sixel

Kreijstal:
This commit adds the --enable-sixel flag of tmux.

### DIFF
--- a/tmux/PKGBUILD
+++ b/tmux/PKGBUILD
@@ -39,6 +39,7 @@ prepare() {
 build() {
   cd "${srcdir}/${pkgname}-${_base_ver}${_bugfix/./}"
   ./configure \
+    --enable-sixel \
     --prefix=/usr \
     --sysconfdir=/etc \
     --localstatedir=/var \


lazka:
increase pkgrel.

and why is this needed/wanted?

lazka:
to be clear, I don't oppose this, I'd just like to have some context

Kreijstal:
> to be clear, I don't oppose this, I'd just like to have some context

Well, it simply adds sixel capabilities to tmux, so while mintty is sixel capable when you are using tmux, it your sixels won't be seen. Which is a shame, but now that I think of it, the libsixel library might be a build-dependency on tmux when this flag is enabled. I'd have to test this.

Biswa96:
> but now that I think of it, the libsixel library might be a build-dependency on tmux when this flag is enabled.

According to https://gitlab.archlinux.org/archlinux/packaging/packages/tmux/-/commit/3f25ffee1121a4eb837a292a19d49751a70dd49d

> as it turns out that tmux comes with its own implementation, and does not depend in libsixel, at all.

Biswa96:
I have tested the CI artifact but sixel is not shown in tmux with and without conpty. This repository may help to test sixel https://codeberg.org/coffee/sixel-experiments e.g. `cat text-test.sixel`

lazka:
> I have tested the CI artifact but sixel is not shown in tmux with and without conpty. This repository may help to test sixel https://codeberg.org/coffee/sixel-experiments e.g. `cat text-test.sixel`

I can confirm that running `./sixel.sh` from there works with this change, and doesn't work without it.

PR #4421, opened 2024-02-15
msys2-runtime-3.3: add -Wno-error for additional warning on i686 build

jeremyd2019:
On i686, an additional warning showed up that failed the build.  This adds this warning to the CFLAGS/CXXFLAGS exclusions.

```
/t/r/msys2-runtime-3.3/src/msys2-runtime/winsup/cygwin/fhandler_proc.cc: In function 'off_t format_proc_swaps(void*, char*&)': /t/r/msys2-runtime-3.3/src/msys2-runtime/winsup/cygwin/fhandler_proc.cc:1964:43: error: '%s' directive writing 4 bytes into a region of size between 1 and 2147483606 [-Werror=format-overflow=]
 1964 |           bufptr += sprintf (bufptr, "%s%s%s\t\t%llu%s\t%llu%s\t%d\n",
      |                                           ^~
......
 1967 |                                         "file",
      |                                         ~~~~~~
/t/r/msys2-runtime-3.3/src/msys2-runtime/winsup/cygwin/fhandler_proc.cc:1964:38: note: directive argument in the range [0, 18014398501093376]
 1964 |           bufptr += sprintf (bufptr, "%s%s%s\t\t%llu%s\t%llu%s\t%d\n",
      |                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/t/r/msys2-runtime-3.3/src/msys2-runtime/winsup/cygwin/fhandler_proc.cc:1964:38: note: directive argument in the range [0, 18014398501093376]
/t/r/msys2-runtime-3.3/src/msys2-runtime/winsup/cygwin/fhandler_proc.cc:1964:29: note: 'sprintf' output between 54 and 2147483693 bytes into a destination of size 2147483647
 1964 |           bufptr += sprintf (bufptr, "%s%s%s\t\t%llu%s\t%llu%s\t%d\n",
      |                     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1965 |                                     filename,
      |                                     ~~~~~~~~~
 1966 |                                     tabo < 5 ? "\t\t\t\t\t" + tabo : " ",
      |                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1967 |                                         "file",
      |                                         ~~~~~~~
 1968 |                                             total >> 10,
      |                                             ~~~~~~~~~~~~
 1969 |                                             total < 10000000000 ? "\t" : "",
      |                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1970 |                                                 used  >> 10,
      |                                                 ~~~~~~~~~~~~
 1971 |                                                 used  < 10000000000 ? "\t" : "",
      |                                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1972 |                                                                 0);
      |                                                                 ~~
```

### DIFF
--- a/msys2-runtime-3.3/0057-Move-_cygheap_start-into-.cygheap-section.patch
+++ b/msys2-runtime-3.3/0057-Move-_cygheap_start-into-.cygheap-section.patch
@@ -1,0 +1,34 @@
+From 1f3be1e79303711927bf7462625a7522ef57a3d9 Mon Sep 17 00:00:00 2001
+From: Jeremy Drake <github@jdrake.com>
+Date: Tue, 20 Feb 2024 10:18:56 -0800
+Subject: [PATCH 57/N] Move _cygheap_start into .cygheap section.
+
+Binutils >= 2.41 started making .rsrc section read-only, which caused
+memset(_cygheap_start, ...) to segfault.  Instead, put _cygheap_start in
+the .cygheap section, which makes more sense anyway, but may result in
+reducing the difference between _cygheap_start and _cygheap_end by the
+fraction of a page not used in the .rsrc section.
+
+Fixes #200
+See-also: https://cygwin.com/pipermail/cygwin/2024-February/255472.html
+Signed-off-by: Jeremy Drake <github@jdrake.com>
+---
+ winsup/cygwin/cygwin.sc.in | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/winsup/cygwin/cygwin.sc.in b/winsup/cygwin/cygwin.sc.in
+index 264e240..dec8f87 100644
+--- a/winsup/cygwin/cygwin.sc.in
++++ b/winsup/cygwin/cygwin.sc.in
+@@ -142,10 +142,10 @@ SECTIONS
+   {
+     *(.rsrc)
+     *(SORT(.rsrc$*))
+-    _SYM (_cygheap_start) = .;
+   }
+   .cygheap ALIGN(__section_alignment__) :
+   {
++    _SYM (_cygheap_start) = .;
+ #ifdef __x86_64__
+     . = . + (3072 * 1024);
+ #else
--- a/msys2-runtime-3.3/0058-Cygwin-dumper-also-link-with-libzstd-as-libbfd-may-r.patch
+++ b/msys2-runtime-3.3/0058-Cygwin-dumper-also-link-with-libzstd-as-libbfd-may-r.patch
@@ -1,0 +1,61 @@
+From 69e4fbea86b07a7960c1fd98b0e17bd5bdb4d2b8 Mon Sep 17 00:00:00 2001
+From: Jon Turney <jon.turney@dronecode.org.uk>
+Date: Tue, 14 Feb 2023 13:52:39 +0000
+Subject: [PATCH 58/N] Cygwin: dumper: also link with libzstd, as libbfd may
+ require it
+
+Also allow that linkage to be dynamic, as libzstd-devel doesn't
+currently provide a static library.
+
+(cherry picked from commit 1387ea9f984d5a7aa096a66b67d61dc2cc565d21)
+---
+ winsup/configure.ac      | 10 ++++++----
+ winsup/utils/Makefile.am |  7 ++-----
+ 2 files changed, 8 insertions(+), 9 deletions(-)
+
+diff --git a/winsup/configure.ac b/winsup/configure.ac
+index ca2b8c0..977a5ce 100644
+--- a/winsup/configure.ac
++++ b/winsup/configure.ac
+@@ -110,10 +110,12 @@ AC_CHECK_LIB([bfd], [bfd_init], [true],
+ 
+ AM_CONDITIONAL(BUILD_DUMPER, [test "x$ac_cv_lib_bfd_bfd_init" = "xyes"])
+ 
+-AC_CHECK_LIB([sframe], [sframe_decode],
+-	     AC_MSG_NOTICE([Detected libsframe; Assuming that libbfd depends on it]), [true])
+-
+-AM_CONDITIONAL(HAVE_LIBSFRAME, [test "x$ac_cv_lib_sframe_sframe_decode" = "xyes"])
++# libbfd.a doesn't have a pkgconfig file, so we guess what it's dependencies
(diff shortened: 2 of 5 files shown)


jeremyd2019:
Even with this, the resulting package seems to be broken on i686.  Might have some debugging to do...

jeremyd2019:
@dscho did you happen to run into issues building an updated msys2-runtime-3.3 for i686?

jeremyd2019:
I don't know why this would make any sense:
```
Starting program: C:\msys32\usr\bin\bash.exe -l
[New Thread 4912.0x1398]
[New Thread 4912.0x1504]
[New Thread 4912.0x34d0]
warning: cYgFFFFFFFF 6122D4D8 0

Thread 1 received signal SIGSEGV, Segmentation fault.
memset () at ../../../../.././newlib/libc/machine/i386/memset.S:94
```

```
(gdb) bt
#0  memset () at ../../../../.././newlib/libc/machine/i386/memset.S:94
#1  0x61005eec in cygheap_init()@0 ()
    at ../../.././winsup/cygwin/cygheap.cc:258
#2  0x00000000 in ?? ()
(gdb) down
#1  0x61005eec in cygheap_init()@0 ()
    at ../../.././winsup/cygwin/cygheap.cc:258
258           cygheap = (init_cygheap *) memset (_cygheap_start, 0,
(gdb) down
#0  memset () at ../../../../.././newlib/libc/machine/i386/memset.S:94
94              rep
(gdb) p _cygheap_start
$2 = 0x612eb408 ""
(gdb) i regis
eax            0x0                 0
ecx            0x1f44              8004
edx            0x0                 0
ebx            0x61005eec          1627414252
esp            0x85c45c            0x85c45c
ebp            0x85f6dc            0x85f6dc
esi            0xffffffff          -1
edi            0x85d95c            8771932
eip            0x0                 0x0
eflags         0x10246             [ PF ZF IF RF ]
cs             0x23                35
ss             0x2b                43
ds             0x2b                43
es             0x2b                43
fs             0x53                83
gs             0x2b                43
```
```
   0x61005ed0 <+112>:   movl   $0x7d10,0x8(%esp)
   0x61005ed8 <+120>:   movl   $0x0,0x4(%esp)
   0x61005ee0 <+128>:   movl   $0x612eb408,(%esp)
   0x61005ee7 <+135>:   call   0x61192d30 <memset>
```

So memset is screwed up somehow?!?

jeremyd2019:
No, I stepped through it and it's trying to write to 0x612eb408 and getting a SIGSEGV there.  So maybe the linker stuff is messed up and it isn't turning out to be a writable section?

jeremyd2019:
I downgraded gcc to the version that was in place the last time I built a working msys2-runtime-3.3 (11.3.0-4), and the problem persisted.  I then upgraded gcc back to the most recent, and downgraded binutils to the last known working (2.40-1) and the problem went away.  Next step will be to find the most recent binutils that still worked

jeremyd2019:
2.40-1 was the last version that worked.

dscho:
> @dscho did you happen to run into issues building an updated msys2-runtime-3.3 for i686?

Not recently, but keep in mind that Git for Windows i686 MSYS builds are essentially "stuck" with the toolchain version that is able to build the MSYS2 runtime v3.3 without any additional `-W` flags.

lazka:
msys2-runtime-3.3 seems to be broken for me too (64bit), so maybe it's a 3.3+binutils problem?

jeremyd2019:
> msys2-runtime-3.3 seems to be broken for me too (64bit), so maybe it's a 3.3 related problem?

"good", in that I don't have to try to track down a 32-bit specific issue :grin:

jeremyd2019:
Did the previous revision of msys2-runtime-3.3 still work on 64-bit?  Seems like it would have, because I was actively using it on 32-bit... If so, maybe the update should be reverted if it is broken?

jeremyd2019:
Yes, on x86_64 msys2-runtime-3.3-3.3.6-6 worked while -7 did not.

lazka:
I'll revert later

lazka:
I've reverted the package in the repo, and reverted the pkgrel bump (only).

If all fails, and this really is binutils specific, we can introduce a separate 2.40 binutils package.

jeremyd2019:
Apparently not too important, I think I was the only one who reported an issue, and then on i686 simply because this is the only runtime version packaged that still builds there.  I assume you only checked on x86_64 because of my report, rather than some actual user of msys2-runtime-3.3 on x86_64 reporting an issue.

lazka:
It was actually motivated by the upcoming msys2-runtime-3.4 for win7 users, and I wanted to check how things work with 3.3 for a comparison.

jeremyd2019:
I've opened msys2/msys2-runtime#203.  This PR currently contains the patches from that, for testing purposes.  Once that is merged, I'll update this PR with the properly formatted patches (I just used `git format-patch` and manually moved them in for now, but should use the update-patches.sh script)

lazka:
I can confirm that things start again with this.

I also noticed though that things like python3 fail to start due to missing symbols, like __cpuset_free, __cpuset_alloc.
Maybe backporting https://github.com/cygwin/cygwin/commit/d27e9fb29a92fc5984de51470d68d859783ab986 would help?

jeremyd2019:
I was a little confused, that commit only showed as being in the 3.5 release.  Turns out it was already backported to 3.4 in msys2/msys2-runtime@797d1bbd2e115872df10cf1c15ccc6cfc99648d6.  Maybe I'm overthinking it (since it was already backported to a 3.4 point release) but maybe we could just backport the exported functions and leave the headers alone for 3.3?  (or at least mostly alone, might need to prototype the exported functions).

lazka:
The motivation stated in the commit is coreutils, which we still haven't updated (it's on my todo list..), so I'd guess backporting it as a whole might help there in the future(?)

jeremyd2019:
The commit was part of a whole series of commits tweaking cpuset.h, and appeared to start with making it compile with clang.  I'm kind of guessing here, that the version in 3.3 would probably work with coreutils, just not with clang (but we already know the clang we have in MSYS2-packages is not to be trusted to compile for the cygwin/msys2 target anyway).

In any event, this seems like something for a future PR against msys2-runtime and we can test some different things out.

PR #4422, opened 2024-02-15
msys2-runtime: introduce `msys2-runtime-3.5`

dscho:
With this PR, the `msys2-runtime` package is still tracking Cygwin v3.4.x, and a new `msys2-runtime-3.5` package is introduced that wants to track Cygwin v3.5.x for the time being.

The reason why `msys2-runtime` is not upgraded to v3.5.x immediately is that this upgrade would have major implications: Cygwin v3.5.x (and hence MSYS2 runtime v3.5.x) drops support for Windows 7 and Windows 8.

To accommodate for that complication, let's fork off the v3.5.* release train into a separate `msys2-runtime-3.5` package, [much like we did with the v3.3.* release train](https://github.com/msys2/MSYS2-packages/pull/3379).

Having this in a separate package (that conflicts with `msys2-runtime`, forcing that package to be uninstalled when switching to `-3.5`) lets interested users kick the tires in a safe and easily-reversible way.

Note: Originally, my plan was to fork off an `msys2-runtime-3.4` package and then switch `msys2-runtime` to v3.5.x, but @lazka helpfully suggested that a safer course of action would be to introduce `msys2-runtime-3.5` first and leave `msys2-runtime` alone, for the time being.


lazka:
Do you happen know if things built against 3.5 will still work with 3.4? I'll ask in IRC.

I tried our 3.3 package yesterday and that no longer seems to work. I haven't looked into it though.

lazka:
> Do you happen know if things built against 3.5 will still work with 3.4? I'll ask in IRC.

> \<jturney\> generally, yes
>\<jturney\> as discussed on #cygwin-developers, unless the thing built against 3.5 uses a newly added API, or uses cpu flags above the level required for the OSes which 3.4 can run on

dscho:
> Do you happen know if things built against 3.5 will still work with 3.4?

I know for a fact that it works the other way round.

> I tried our 3.3 package yesterday and that no longer seems to work. I haven't looked into it though.

That bug seems to be tracked in https://github.com/msys2/msys2-runtime/issues/200, right?

lazka:
> That bug seems to be tracked in [msys2/msys2-runtime#200](https://github.com/msys2/msys2-runtime/issues/200), right?

yes

lazka:
(we could also do a msys2-runtime-3.5 package, and dog food it for some time. not saying that I'd prefer that, just another option)

dscho:
I like that idea!

dscho:
> I like that idea!

And at long last, I had time to implement it...

lazka:
thanks!

lhmouse:
This seems to work perfectly. Thanks.

PR #4475, opened 2024-03-29
xz: Switch from upstream tarball to git tarball

lazka:
Run autogen to get the documentation built instead of just autoreconf Also needs extra build depends for things already included in the tarball.

And remove the signing key from the bad actor.

### DIFF
--- a/xz/PKGBUILD
+++ b/xz/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=('xz' 'liblzma' 'liblzma-devel')
 pkgver=5.6.1
-pkgrel=1
+pkgrel=2
 pkgdesc='Library and command line tools for XZ and LZMA compressed files'
 arch=('i686' 'x86_64')
 url='https://tukaani.org/xz/'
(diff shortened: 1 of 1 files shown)


jeremyd2019:
I'm (already) getting a checksum failure on that new tar.

jeremyd2019:
oh, access to the repository has been disabled due to terms of service violation.  That's messing up the tar link too.

jeremyd2019:
For now, I'm reverting to 5.6.0-1 in my i686 msys repo.

lazka:
uh, annoying. It seems contra productive to hinder people from downgrading, or switching sources.

> For now, I'm reverting to 5.6.0-1 in my i686 msys repo. EDIT: oops, should have read first, 5.6.0 was also backdoored. But, luckily, it sounds like only linux x86_64 was targeted, and then only when building rpm or deb packages. Reverted to 5.4.6-2 now.

Yeah, it shouldn't be a problem either way. But using git at least doesn't run any of the found (as of now) problematic code.

Biswa96:
Would it be possible to use Debian repository https://salsa.debian.org/debian/xz-utils ? I am not sure if it includes some changes specific to Debian.

PR #4477, opened 2024-03-31
bash: Fix CR handling

### LINKED ISSUE #1839: bash: '\n' after command substitution in $PS1 causes error
When I use `\n` after command substitution in `$PS1`, it causes syntax error:

```
$ PS1='$(date)\n\$ '
-bash: command substitution: line 1: syntax error near unexpected token `)'
-bash: command substitution: line 1: `date)'
```

It should not cause the error.
If I remove the `\n` after `$(date)`, it works fine:

```
$ PS1='$(date)\$ '
Mon Jan 27 19:16:55 JST 2020$ 
```

Using backquotes instead of `$()` also works fine:

```
$ PS1='`date`\n\$ '
Mon Jan 27 19:21:51 JST 2020
$ 
```

Setting `PS1='$(date)\n\$ '` works fine on Cygwin, so this seems an MSYS2 specific issue.

See also: https://stackoverflow.com/questions/21517281/ps1-command-substitution-fails-when-containing-newlines-on-msys-bash

k-takata:
Fixes #1839

`0005-bash-4.3-msys2-fix-lineendings.patch` adds CRLF support. However, `0001-bash-4.4-cygwin.patch` already added `igncr` option to Bash to support CRLF.

I confirmed that the Cygwin version of Bash has also the same issue with #1839 when the `igncr` option is set.

After debugging, I found that there is an issue in `rewind_input_string()` in `parser.y` that it doesn't take the CR into account.

This PR adds the following changes:
* ~~Set the `igncr` option by default.~~
* ~~Remove unnecessary changes in `yy_input_name()` in `parser.y` and use the `igncr` implementation.~~
* Modify `rewind_input_string()` to take the CR into account. (It might be better to apply a similar change to the Cygwin version of Bash.)
* Remove all the changes from `y.tab.c`. This file should be automatically generated from `parser.y`.
* Set `LC_ALL=C.UTF-8` when running `make check` for running on non-English locales.
* Add two tests: `run-ps1lf` and `run-crlf`.

### DIFF
--- a/bash/0005-bash-4.3-msys2-fix-lineendings.patch
+++ b/bash/0005-bash-4.3-msys2-fix-lineendings.patch
@@ -14,6 +14,84 @@ index c1135ec..b388af6 100644
    if (niflp)
      *niflp = invfl;
    if  (vlp)
+diff --git a/parse.y b/parse.y
+index 8fd24a1c..232a33dc 100644
+--- a/parse.y
++++ b/parse.y
+@@ -1684,7 +1690,15 @@ rewind_input_string ()
+      into account, e.g., $(...\n) */
+   xchars = shell_input_line_len - shell_input_line_index;
+   if (bash_input.location.string[-1] == '\n')
+-    xchars++;
++    {
++      xchars++;
++#ifdef __MSYS__
++      {
++	if (bash_input.location.string[-2] == '\r')
++	  xchars++;
++      }
++#endif
++    }
+ 
+   /* XXX - how to reflect bash_input.location.string back to string passed to
+      parse_and_execute or xparse_dolparen? xparse_dolparen needs to know how
+diff --git a/tests/crlf.right b/tests/crlf.right
+new file mode 100644
+index 0000000..d7fd195
+--- /dev/null
++++ b/tests/crlf.right
+@@ -0,0 +1,8 @@
++Line with LF
++Line with CR
++Line with CRLF
++Line with
++ LF
++Line with CR
++Line with
++ CRLF
+diff --git a/tests/crlf.tests b/tests/crlf.tests
+new file mode 100644
+index 0000000..b2c4da7
+--- /dev/null
++++ b/tests/crlf.tests
+@@ -0,0 +1,6 @@
++echo $(echo -e "Line with\n LF")
++echo $(echo -e "Line with\r CR")
++echo $(echo -e "Line with\r\n CRLF")
++echo "$(echo -e "Line with\n LF")"
++echo "$(echo -e "Line with\r CR")"
++echo "$(echo -e "Line with\r\n CRLF")"
+diff --git a/tests/ps1lf.right b/tests/ps1lf.right
+new file mode 100644
+index 0000000..ee83131
+--- /dev/null
++++ b/tests/ps1lf.right
+@@ -0,0 +1,2 @@
++foo
++$ exit
+diff --git a/tests/ps1lf.tests b/tests/ps1lf.tests
+new file mode 100644
+index 0000000..01a2265
+--- /dev/null
++++ b/tests/ps1lf.tests
+@@ -0,0 +1 @@
++PS1='$(echo foo)\n\$ '
+diff --git a/tests/run-crlf b/tests/run-crlf
+new file mode 100644
+index 0000000..3f6037c
+--- /dev/null
++++ b/tests/run-crlf
+@@ -0,0 +1,2 @@
++${THIS_SH} ./crlf.tests > ${BASH_TSTOUT}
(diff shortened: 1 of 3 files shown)


lazka:
Sorry for the delay. I'll try to find some time/motivation for a more in-depth look.

Do you see any potential problems with the default changing for existing users?

k-takata:
`0005-bash-4.3-msys2-fix-lineendings.patch` and the `igncr` option of `0001-bash-4.4-cygwin.patch` had duplicated functionality.
MSYS2's bash has already done almost the same thing as `igncr` is set to 1. So, changing the default of `igncr` (and removing the duplicated part) doesn't have potential issues, I think. (If a user explicitly unset `igncr`, the behavior will be changed, though.)

----
 
In this PR, the most important part is here:
```diff
diff --git a/parse.y b/parse.y
index 8fd24a1c..232a33dc 100644
@@ -1678,7 +1684,16 @@ rewind_input_string ()
      into account, e.g., $(...\n) */
   xchars = shell_input_line_len - shell_input_line_index;
   if (bash_input.location.string[-1] == '\n')
-    xchars++;
+    {
+      xchars++;
+#ifdef __CYGWIN__
+      {
+	extern int igncr;
+	if (igncr && bash_input.location.string[-2] == '\r')
+	  xchars++;
+      }
+#endif
+    }
 
   /* XXX - how to reflect bash_input.location.string back to string passed to
      parse_and_execute or xparse_dolparen? xparse_dolparen needs to know how
```

If you have concerning about changing `igncr`, then this part can be like this:
```diff
diff --git a/parse.y b/parse.y
index 8fd24a1c..232a33dc 100644
@@ -1678,7 +1684,15 @@ rewind_input_string ()
      into account, e.g., $(...\n) */
   xchars = shell_input_line_len - shell_input_line_index;
   if (bash_input.location.string[-1] == '\n')
-    xchars++;
+    {
+      xchars++;
+#ifdef __MSYS__
+      {
+	if (bash_input.location.string[-2] == '\r')
+	  xchars++;
+      }
+#endif
+    }
 
   /* XXX - how to reflect bash_input.location.string back to string passed to
      parse_and_execute or xparse_dolparen? xparse_dolparen needs to know how
```
(Some other hunks need to be reverted in this case.)

lazka:
I gave it another try. One thing `igncr` also changes is it ignores "\r" in text in general, not just at the end:

```
test3() {
    local input="Line with\r Carriage Return"
    local expected="Line with Carriage Return"
    local result=$(echo -e "$input")
    [ "$result" != "$expected" ] && echo "OK" || echo "FAILED"
}

test3
```

that feels a bit risky, so I'd prefer the non-igncr variant, for now at least.

I'll try to write some tests covering the changes we currently have compared to upstream. Do you also have some example code that tests your changes, so we can add it to our integration tests?

After that we can think about enabling igncr again.

k-takata:
OK, I've updated the PR not to use igncr.

> Do you also have some example code that tests your changes, so we can add it to our integration tests?

Unfortunately, I have no ideas how to test the PS1 issue.
I currently set `PS1='$(date)\n\$ '` manually to see if it works.

Another test that might be better to add is to test a script file with CRLF line endings.

k-takata:
(Sorry, closed by mistake. Reopened.)

akinomyoga:
We received another report in oh-my-bash caused by this issue. May I ask what is the current blocker of this fix? Is it blocked because no one knows how to test the change? What is the current progress for the tests?

k-takata:
Added a test named `run-ps1lf`.
If this test is run on the bash without the patch, the following differences are shown:
```
run-ps1lf
1,3c1
< bash: command substitution: line 1: syntax error near unexpected token `)'
< bash: command substitution: line 1: `echo foo)'
<
---
> foo
5,7c3
< bash: command substitution: line 1: syntax error near unexpected token `)'
< bash: command substitution: line 1: `echo foo)'
<
---
> foo
```

On the patched version, only the test name is shown:
```
run-ps1lf
```

bash/0005-bash-4.3-msys2-fix-lineendings.patch
```diff
@@ -156,6 +173,63 @@ index 2001b4e..3ba2029 100644
...
++Line with LF
++Line with
 CR
++Line with
 CRLF
++Line with
++ LF
++Line with
 CR
++Line with
```
k-takata, line 188: Lines 183, 184, and 187 contain CR, and line 188 ends with CRLF.
`bash/.gitattributes` has been added to keep the CRs.

k-takata (commented):
`run-crlf` is added to test CRLF handling in `$(...)`.

k-takata:
@lazka
Rebased on top of the latest master branch.
I think this is ready for merging.

k-takata:
Rebased onto the latest master branch.

PR #4478, opened 2024-03-31
Remove procps, make procps-ng replace it

lazka:
That's already the case in Cygwin and Arch


lazka:
Next step: Sync procps-ng from cygwin

PR #4480, opened 2024-03-31
msys2-runtime-3.5: abort installation on Windows 7 and 8.0

lazka:
Cygwin 3.5 no longer supports Windows 7 and 8.0 and letting users update would render all cygwin tools broken.

There is no good way to abort an installation from within a package, so we run an install script on pre install/upgrade and just kill pacman from there, after printing a message for how to fix things. Since killing pacman would leave the DB locked, remove the lock file before, so the suggested way forward works.

While we don't officially support Win 7/8.0 for quite a while now, this should help those remaining users keep using the parts that still work for a bit longer.

### DIFF
--- a/msys2-runtime-3.5/PKGBUILD
+++ b/msys2-runtime-3.5/PKGBUILD
@@ -4,7 +4,7 @@
 pkgbase=msys2-runtime-3.5
 pkgname=('msys2-runtime-3.5' 'msys2-runtime-3.5-devel')
 pkgver=3.5.1
-pkgrel=2
+pkgrel=3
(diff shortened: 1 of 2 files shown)


lazka:
image

lazka:
This doesn't seem to work, at least for the 3.5 update case, since that will first uninstall msys2-runtime at which point the install script for 3.5 fails to run.

So I'm going to apply this to msys2-runtime directly now (which is a good idea anyway, since it it fails users still have a working msys2-runtime, since we are still on 3.4 there)

PR #4515, opened 2024-04-07
pacman: remove `-pipe` from default `LDFLAGS` for all environments

mmuetzel:
It probably doesn't do anything and breaks Flang 18 when it is being called as a linker.

See https://github.com/msys2/MINGW-packages/pull/20591.

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=pacman
 pkgver=6.0.2
-pkgrel=13
+pkgrel=14
 pkgdesc="A library-based package manager with dependency support (MSYS2 port)"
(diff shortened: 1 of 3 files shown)


lazka:
I've updated the PKGBUILD and expanded the commit message. I hope that's OK.

mmuetzel:
Oops. Rookie mistake. Thank you for fixing it and expanding on the message.

As for the reason why that starts to fail now: ISTR, that flang accepted all flags that clang would accept in version 17. For LLVM 18, the developers made an effort to only mark flags that are supported for the respective frontend as visible.
I don't know if not adding `-pipe` to Flang was an oversight or if it was never meant to work with flang as the frontend for the driver.

PR #4560, opened 2024-04-27
pacutils: Update to 0.14.0

lazka:


lazka:
Needs newer pacman: https://github.com/msys2/msys2-pacman/issues/38

PR #4570, opened 2024-04-29
ninja: Update to 1.12.1

lazka:


lazka:
huh, the failing code hasn't changed in a while...

PR #4576, opened 2024-05-02
msys2-runtime: abort installation on Windows 7 and 8.0

lazka:
Cygwin 3.5 no longer supports Windows 7 and 8.0 and letting users update would render all cygwin tools broken.

There is no good way to abort an installation from within a package, so we run an install script on pre install/upgrade and just kill pacman from there, after printing a message for how to fix things. Since killing pacman would leave the DB locked, remove the lock file before, so the suggested way forward works.

While we don't officially support Win 7/8.0 for quite a while now, this should help those remaining users keep using the parts that still work for a bit longer.


lazka:
First attempt was #4480, no using the main package

PR #4593, opened 2024-05-10
pactoys: Update to r65.b5fcc0f

lazka:
See https://github.com/msys2/pactoys/pull/6


lazka:
There is a small potential that this might break someones workflow in a way I didn't anticipate. Let us know if that's the case.

PR #4596, opened 2024-05-13
msys2-runtime: attempt to use makepkg's split debuginfo feature

### LINKED ISSUE #4595: msys2-runtime-devel: missing debug info
msys2-runtime-devel-3.5.3-2 is missing msys-2.0.dbg.  It also seems to be missing from msys2-runtime-3.4-devel-3.4.10-1, so I kind of suspect binutils, though I haven't looked into it at all.

jeremyd2019:
To avoid additional changes to msys2-runtime's build (such as changing flags due to 'debug' being set), I didn't set debug at the top level of the PKGBUILD, but rather just in the msys2-runtime package.  This means that makepkg didn't attempt to actually package an msys2-runtime-debug package, so I included the files in msys2-runtime-devel.

It doesn't look like the debug info produced is large enough to be complete, and objdump seems to back this up.  Also, there are some extra files around the .exe files (whatever.exe.debug.exe in addition to some having whatever.exe.debug which is the name I would expect).

I think the makepkg procedure to produce split debuginfo is broken, but I wanted to open this as a draft to start the discussion.

Fixes #4595

### DIFF
--- a/msys2-runtime/PKGBUILD
+++ b/msys2-runtime/PKGBUILD
@@ -237,6 +237,7 @@ package_msys2-runtime-devel() {
   depends=("msys2-runtime=${pkgver}")
   conflicts=('libcatgets-devel' 'msys2-runtime-3.4-devel' 'msys2-runtime-3.5-devel')
   replaces=('libcatgets-devel' 'msys2-runtime-3.5-devel')
+  options=('!strip')
 
   mkdir -p "${pkgdir}"/usr/bin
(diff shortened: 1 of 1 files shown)


jeremyd2019:
An alternative to this approach would be https://github.com/msys2/MSYS2-packages/compare/master...jeremyd2019:MSYS2-packages:msys2-runtime-split-debuginfo-2 which results in the same msys-2.0.dbg file as before.  I don't know which would be preferred.

jeremyd2019:
I'm inclined to go with the msys-2.0.dbg version.  Closing this

PR #4600, opened 2024-05-14
msys2-runtime{,3.4}: restore msys-2.0.dbg

### LINKED ISSUE #4595: msys2-runtime-devel: missing debug info

jeremyd2019:
Replicate the objcopy commands that used to work in msys2-runtime build process.

also, don't strip msys2-runtime-devel.  makepkg's strip mechanism thinks split debug files are binaries that need to be stripped, and debug files with the debug info stripped are pretty useless.  msys2/msys2-pacman#52

Fixes #4595

### DIFF
--- a/msys2-runtime-3.4/PKGBUILD
+++ b/msys2-runtime-3.4/PKGBUILD
@@ -4,7 +4,7 @@
 pkgbase=msys2-runtime-3.4
 pkgname=('msys2-runtime-3.4' 'msys2-runtime-3.4-devel')
 pkgver=3.4.10
-pkgrel=1
+pkgrel=2
 pkgdesc="Cygwin POSIX emulation engine"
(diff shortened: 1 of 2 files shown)


jeremyd2019:
If this is acceptable, I can apply the same changes to msys2-runtime-3.4

jeremyd2019:
I applied these changes to msys2-runtime-3.4, and added `-Wno-error=maybe-uninitialized` to both.  msys2-runtime-3.3 didn't appear to need either.

lazka:
nice, thanks

PR #4602, opened 2024-05-17
msys2-w32api: Update to 11.0.0.r750.g05598db99

lazka:


Biswa96:
Isn't this a version downgrade?

PR #4605, opened 2024-05-20
gpgme: cache the engine version

lazka:
See https://github.com/msys2/msys2-pacman/issues/26
See https://dev.gnupg.org/T6369

This saves a gpg spawn for every gpgme_get_key() call, for example around 100ms when starting pacman.


lazka:
Updated to the patch that was applied upstream.

PR #4614, opened 2024-05-22
CI: fix dll dependency check

### LINKED ISSUE #4611: CI: dll check doesn't work
It seems every package has the dll check fail with:
```
find: missing argument to `-exec'
```

I believe this is due to this line
https://github.com/msys2/MSYS2-packages/blob/ec5ca31736cbe33c62230cf2f8141b93b9d2f22e/.ci/ci-build.sh#L68

and find's `-exec` option not running its argument in a shell.  I get the same error with:
```
find ... -exec "echo '{}:' && ldd '{}';"
```
but get correct behavior with, for instance:
```
find ... -exec dash -c "echo '{}:' && ldd '{}'" \;
```

Or, avoiding a shell,
```
find ... -exec echo '{}:' \;  -exec ldd '{}' \;
```

Generally, I'm not sure that #3917 actually worked as intended.

jeremyd2019:
The find expression was broken for a couple of reasons: it does not run in a subshell (so cannot use && like that), and it needs split arguments not one quoted argument.  Avoid having to call echo by using -printf argument.

Fixes #4611

### DIFF
--- a/.ci/ci-build.sh
+++ b/.ci/ci-build.sh
@@ -65,7 +65,7 @@ install_packages() {
 list_dll_deps(){
     local target="${1}"
     echo "$(tput setaf 2)MSYS2 DLL dependencies:$(tput sgr0)"
-    find "$target" -regex ".*\.\(exe\|dll\)" -exec "echo '{}:' && ldd '{}';" | GREP_COLOR="1;35" grep --color=always "msys-.*\|" \
+    find "$target" -regex ".*\.\(exe\|dll\)" -printf '%p:\n' -exec ldd '{}' \; | GREP_COLOR="1;35" grep --color=always "msys-.*\|" \
     || echo "        None"
 }
 
--- a/.ci/ci-build.sh
+++ b/.ci/ci-build.sh
@@ -61,21 +61,6 @@ install_packages() {
     pacman --noprogressbar --upgrade --noconfirm *.pkg.tar.*
 }
 
-# List DLL dependencies
-list_dll_deps(){
-    local target="${1}"
-    echo "$(tput setaf 2)MSYS2 DLL dependencies:$(tput sgr0)"
-    find "$target" -regex ".*\.\(exe\|dll\)" -printf '%p:\n' -exec ldd '{}' \; | GREP_COLOR="1;35" grep --color=always "msys-.*\|" \
-    || echo "        None"
-}
-
-list_dll_bases(){
-    local target="${1}"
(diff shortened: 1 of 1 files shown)


jeremyd2019:
I did a run to test this: https://github.com/jeremyd2019/MSYS2-packages/actions/runs/9195655593/job/25291960933#step:10:609

Interestingly, the dll check seems to happen while dependencies are uninstalled (they are removed by makepkg, but have not yet been reinstalled while installing the package).  This is probably a tougher problem to solve, though, because this check is done for the whole `pkg` directory tree, but split packages are now installed/uninstalled individually.  Perhaps the dll dependency check should be moved into the per-package loop (and then might use `pacman -Qlq "$pkgname"` instead of `find`).

Also, why does the file check pipe pacman output through sed to remove extraneous output instead of using the `-q` option?

jeremyd2019:
> > Perhaps the dll dependency check should be moved into the per-package loop (and then might use `pacman -Qlq "$pkgname"` instead of `find`).
> 
> Sounds good, afaics this is also done that way in the MINGW repo already

Hmm, I thought the CI scripts were synchronized at some point.  must have diverged again since then

Biswa96:
Moving the scripts to a common repository may help.

lazka:
Should we sync it here or in a separate PR?

jeremyd2019:
I was thinking about doing it here, but the few differences are mostly specific to MINGW-packages.  I will try moving the dll checks though.

jeremyd2019:
Sorry, I had other things come up and I didn't finish testing this yesterday.  I think it's ready now.

lazka:
It got stuck on my first try: https://github.com/msys2/MSYS2-packages/actions/runs/9222109999/attempts/1

maybe just unlucky, we'll see.

lazka:
it also got stuck the second time

jeremyd2019:
https://cygwin.com/pipermail/cygwin/2024-May/256002.html

> I think the fix for this would be to add -Wl,--disable-high-entropy-va to
> ldh_LDFLAGS, as was done for strace and cygcheck at least.

jeremyd2019:
https://cygwin.com/pipermail/cygwin-patches/2024q2/012696.html

jeremyd2019:
https://github.com/msys2/msys2-runtime/commit/10b921681108b333d17d8438a45afe6fbc1169dd (also backported to cygwin 3.5 branch)

lazka:
Thanks, see https://github.com/msys2/MSYS2-packages/pull/4656

PR #4641, opened 2024-06-01
Fixes #4640 openssl legacy provider missing

### LINKED ISSUE #4640: openssl legacy provider missing
### Description / Steps to reproduce the issue

When attempting to use the legacy provider with mys2 the legacy provider does not seem to be supported.   It looks like the build script does not copy the legacy provider

openssl version
OpenSSL 3.3.0 9 Apr 2024 (Library: OpenSSL 3.3.0 9 Apr 2024)

$ openssl ciphers -provider legacy

341C0000:error:12800067:DSO support routines:win32_load:could not load the shared library:../openssl-3.3.0/crypto/dso/dso_win32.c:108:filename(ZLIB1.dll)
341C0000:error:12800067:DSO support routines:DSO_load:could not load the shared library:../openssl-3.3.0/crypto/dso/dso_lib.c:147:
341C0000:error:0A0000A1:SSL routines:SSL_CTX_new_ex:library has no ciphers:../openssl-3.3.0/ssl/ssl_lib.c:3955:

### Expected behavior

The legacy provider should be available

### Actual behavior

it is not

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

CYGWIN_NT-10.0-19042

### Are you willing to submit a PR?

yes

timlegge:
Should probably be in the other targets as well

I don't have the build tools installed so I am not certain of the change but it should be close

### DIFF
--- a/openssl/PKGBUILD
+++ b/openssl/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=('openssl' 'libopenssl' 'openssl-devel' 'openssl-docs')
 pkgver=3.3.0
-pkgrel=1
+pkgrel=2
 pkgdesc='The Open Source toolkit for Secure Sockets Layer and Transport Layer Security'
 arch=('i686' 'x86_64')
 url='https://www.openssl.org'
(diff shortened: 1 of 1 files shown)


jeremyd2019:
> Should probably be in the other targets as well

If by that you mean mingw-w64-openssl in MINGW-packages, it looks to me like `${MINGW_PREFIX}/lib/ossl-modules/legacy.dll` is already present there. (example: file list for https://packages.msys2.org/package/mingw-w64-x86_64-openssl?repo=mingw64)

timlegge:
> > Should probably be in the other targets as well
> 
> If by that you mean mingw-w64-openssl in MINGW-packages, it looks to me like `${MINGW_PREFIX}/lib/ossl-modules/legacy.dll` is already present there. (example: file list for https://packages.msys2.org/package/mingw-w64-x86_64-openssl?repo=mingw64)

Yes, that's what I mean and I think you're right. I noticed that afterwards

lazka:
thanks!

It's really annoying that it's so easy to miss new files in split packages.

timlegge:
> thanks!
> 
> It's really annoying that it's so easy to miss new files in split packages.

Thanks.  I have confirmed that the updated version fixed the issue I was having.

PR #4665, opened 2024-06-06
hooks: add some .exe to some invokations

1480c1:


1480c1:
fixes the hooks failing to run on msys2 under docker or archlinux

1480c1:
forgot to bump pkgrel

Biswa96:
> fixes the hooks failing to run on msys2 under docker or archlinux

Could you share the error output without this change?

1480c1:
> > fixes the hooks failing to run on msys2 under docker or archlinux
> 
> Could you share the error output without this change?

```bash
(90/90) installing base                                                                                                    [--------------------------------------------------------------------------] 100%
:: Running post-transaction hooks...
(1/1) Updating the info directory file...
call to execv failed (No such file or directory)
error: command failed to execute correctly
```
This is attempting to use archlinux docker image to use pacman to bootstrap msys2's pacman to further install packages without using the premade tar.xz.

1480c1:
this uses archlinux's pacman and bash rather than msys2's (since at this point, it doesn't have them installed), so that's why it's not able to do the magic .exe suffix lookup that cygwin does.

lazka:
if it helps, sure..

Biswa96:
> For `.install` files, what's the general guidelines for them inside PKGBUILD files?

According to `man 5 PKGBUILD`,

> It does not need to be included in the source array (e.g., install=$pkgname.install).

I guess, both are allowed.

lazka:
So, with native pacman, what runs/executes the install scripts and the .exe files?

1480c1:
> So, with native pacman, what runs/executes the install scripts and the .exe files?

For archlinux "native," it runs it using arch's /usr/bin/bash, which in turn will call the executables using wine that is set up similar to msys2-docker from the wine fork.

For msys2 "native," it runs using msys2's bash on top of our msys2-runtime base on top of wine, so it can handle calling programs without the .exe.

lazka:
> For archlinux "native," it runs it using arch's /usr/bin/bash, which in turn will call the executables using wine that is set up similar to msys2-docker from the wine fork.

oh, fancy, thanks

jeremyd2019:
> For archlinux "native," it runs it using arch's /usr/bin/bash, which in turn will call the executables using wine that is set up similar to msys2-docker from the wine fork.
> 
> For msys2 "native," it runs using msys2's bash on top of our msys2-runtime base on top of wine, so it can handle calling programs without the .exe.

This is interesting.  Can you provide a link/links to these (docker?)?  Does msys2 actually run on top of wine now?

lazka:
For the full wine variant we have: https://github.com/msys2/MSYS2-docker - The other I don't know.

jeremyd2019:
I blame github... they always seem to give me notifications on things I don't care about, but not notifications on what I do... 😁  I'd like to be notified for new repositories under msys2 by default, but I don't think it does that.  I've just "followed" the org, maybe that will do it...

lazka:
We also have a NEWS page ;) https://www.msys2.org/news/

PR #4684, opened 2024-06-10
icu: Update to 75.1

lazka:


jeremyd2019:
this failed to build on i686, because cygwin/msys2-runtime 3.3 doesn't provide uchar.h.  I don't know if I want to bother trying to work around this, or just freeze icu on 32-bit msys for the remainder of the time I keep building packages for it...

jeremyd2019:
https://github.com/unicode-org/icu/commit/e246d3b712c317d6c5e2063cc11db89da7de5094

jeremyd2019:
unicode-org/icu#3038

PR #4692, opened 2024-06-11
parallel: Update to 20240522

Suyun114:


parallel/PKGBUILD
```diff
@@ -13,16 +13,13 @@ msys2_references=(
...
-            '20b526c7ffc8632e5eccb1171e30a08650d550de9e2f6e56fe3bb86165e2d3e7')
+sha256sums=('67ed9fad31bf3e25f09d500e7e8ca7df9e3ac380fe4ebd16c6f014448a346928'
+            'SKIP')
 
 prepare() {
   cd "${srcdir}/${pkgname}-${pkgver}"
-  patch -p1 -i "${srcdir}/0001-support-MSys2.patch"
 
   # ACLOCAL_PATH is needed to pick up local m4 macro
   ACLOCAL_PATH="${srcdir}/${pkgname}-${pkgver}" autoreconf -vif
```
lazka, line 25: With the patch removed, should this ACLOCAL_PATH stuff also be removed?

lazka:
While I haven't tested it, it looks like the patch makes it work if mingw perl is in PATH, so I'm wondering why it was removed?

PR #4693, opened 2024-06-11
libbobcat: Update to 5.11.01

lazka:


lazka:
I don't want to deal with crappy custom made build systems, so I'm giving up for now.

PR #4695, opened 2024-06-13
icu: fix build with cygwin < 3.5.0

jeremyd2019:
This probably only matters for i686, which is stuck on 3.3.6.

I don't know if the upstream pull request (unicode-org/icu#3038) will go anywhere or not, feels like the kind of org that makes you jump through a lot of hoops to contribute anything.


lazka:
Thanks!

PR #4705, opened 2024-06-16
ctags: Update to 6.1.0

Suyun114:


Biswa96:
I am not sure if this should be update to universal ctags project. Hence, there are two ctags package in mingw packages repository.

lazka:
> I am not sure if this should be update to universal ctags project. Hence, there are two ctags package in mingw packages repository.

Didn't we switch there too? (Arch did also)

Biswa96:
Oh, right.

piotrvv:
Better be careful 
ctags: /usr/bin/etags.exe exists in filesystem (owned by emacs)

lazka:
Meh, thanks

PR #4713, opened 2024-06-18
pacman: Add Japanese translation

k-takata:
Add the Japanese translation back which was removed in #4591.


k-takata:
Should I add this patch to the msys2/msys2-packman repository?

Biswa96:
It should be added to upstream https://gitlab.archlinux.org/pacman/pacman

k-takata:
No, these messages are MSYS2 specific messages and they are not included in the upstream source code.
This is the commit: https://github.com/msys2/msys2-pacman/commit/9d91ab906d427f87c5115211b11f0ebef364515c

See also the previous discussion: https://github.com/msys2/MSYS2-packages/pull/1786

Biswa96:
So, the patches should be msys2 pacman.

k-takata:
Created a PR there: https://github.com/msys2/msys2-pacman/pull/53

lazka:
Fixed via #4905

PR #4723, opened 2024-06-27
mintty: update to 3.7.2

Biswa96:


Biswa96:
The build issue is reported to upstream https://github.com/mintty/mintty/issues/1275

PR #4724, opened 2024-07-01
mintty: Update to 3.7.3; add default minttyrc

lazka:
Since 3.7.3 mintty will no longer write user changes to /etc/minttyrc, so we can start shipping defaults there without having to worry about everyone getting config backups when changing things in the GUI.

See https://github.com/msys2/MSYS2-packages/pull/3087 for a previous discussion.


lazka:
@mintty

PR #4743, opened 2024-07-04
Python 3.12.4

lazka:
* drop all distutils patches, distutils is gone
* drop all setup.py patches, that's gone too
* refresh the 32bit rebase (things changed, but I added
  it to the same make target, hopefully still works)
* adjust the new module building code to depend on the import
  lib and also add the build dir to the library path so it can
  be found.
* ctypes dropped vendored things, so remove code deleting it
* the code printing a warning if a module build fails no longer
  exists, so drop the "tee" stuff and the error check there


jeremyd2019:
> * refresh the 32bit rebase (things changed, but I added
>   it to the same make target, hopefully still works)

So far, python has been *the worst* package to have to build for 32-bit, due to imagebase conflicts with all the dll modules, plus the fact that the compileall script uses a forking backend for multiprocessing.  For some reason, I was never able to get it to work right in GHA, but building locally with that hack has worked.

I thought this was a 32-bit thing, but it seems to have happened here too.... any idea what's going on with the compileall script runs in `make install` (copied from CI output from #4761):
```
PYTHONPATH=/C/_/python/pkg/python/usr/lib/python3.12  \
  	./python.exe -E -Wi /C/_/python/pkg/python/usr/lib/python3.12/compileall.py \
  	-o 0 -o 1 -o 2 -j0 -d /usr/lib/python3.12 -f \
  	-x 'bad_coding|badsyntax|site-packages|test/test_lib2to3/data' \
  	/C/_/python/pkg/python/usr/lib/python3.12
  Listing '/C/_/python/pkg/python/usr/lib/python3.12'...
  Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "/C/_/python/src/Python-3.12.4/Lib/multiprocessing/forkserver.py", line 258, in main
  Traceback (most recent call last):
    File "/C/_/python/pkg/python/usr/lib/python3.12/compileall.py", line 468, in <module>
      fds = reduction.recvfds(s, MAXFDS_TO_SEND + 1)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/C/_/python/src/Python-3.12.4/Lib/multiprocessing/reduction.py", line 164, in recvfds
      raise RuntimeError('received %d items of ancdata' %
  RuntimeError: received 0 items of ancdata
      exit_status = int(not main())
                            ^^^^^^
    File "/C/_/python/pkg/python/usr/lib/python3.12/compileall.py", line 445, in main
      if not compile_dir(dest, maxlevels, args.ddir,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/C/_/python/pkg/python/usr/lib/python3.12/compileall.py", line 109, in compile_dir
      results = executor.map(partial(compile_file,
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/C/_/python/pkg/python/usr/lib/python3.12/concurrent/futures/process.py", line 859, in map
      results = super().map(partial(_process_chunk, fn),
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/C/_/python/pkg/python/usr/lib/python3.12/concurrent/futures/_base.py", line 608, in map
      fs = [self.submit(fn, *args) for args in zip(*iterables)]
            ^^^^^^^^^^^^^^^^^^^^^^
    File "/C/_/python/pkg/python/usr/lib/python3.12/concurrent/futures/process.py", line 830, in submit
      self._adjust_process_count()
    File "/C/_/python/pkg/python/usr/lib/python3.12/concurrent/futures/process.py", line 789, in _adjust_process_count
      self._spawn_process()
    File "/C/_/python/pkg/python/usr/lib/python3.12/concurrent/futures/process.py", line 807, in _spawn_process
      p.start()
    File "/C/_/python/pkg/python/usr/lib/python3.12/multiprocessing/process.py", line 121, in start
      self._popen = self._Popen(self)
                    ^^^^^^^^^^^^^^^^^
    File "/C/_/python/pkg/python/usr/lib/python3.12/multiprocessing/context.py", line 301, in _Popen
      return Popen(process_obj)
             ^^^^^^^^^^^^^^^^^^
    File "/C/_/python/pkg/python/usr/lib/python3.12/multiprocessing/popen_forkserver.py", line 35, in __init__
      super().__init__(process_obj)
    File "/C/_/python/pkg/python/usr/lib/python3.12/multiprocessing/popen_fork.py", line 19, in __init__
      self._launch(process_obj)
    File "/C/_/python/pkg/python/usr/lib/python3.12/multiprocessing/popen_forkserver.py", line 57, in _launch
      with open(w, 'wb', closefd=True) as f:
  BrokenPipeError: [Errno 32] Broken pipe
  make: [Makefile:2422: libinstall] Error 1 (ignored)
  PYTHONPATH=/C/_/python/pkg/python/usr/lib/python3.12  \
  	./python.exe -E -Wi /C/_/python/pkg/python/usr/lib/python3.12/compileall.py \
  	-o 0 -o 1 -o 2 -j0 -d /usr/lib/python3.12/site-packages -f \
  	-x badsyntax /C/_/python/pkg/python/usr/lib/python3.12/site-packages
  Listing '/C/_/python/pkg/python/usr/lib/python3.12/site-packages'...
  Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "/C/_/python/src/Python-3.12.4/Lib/multiprocessing/forkserver.py", line 258, in main
  Traceback (most recent call last):
    File "/C/_/python/pkg/python/usr/lib/python3.12/compileall.py", line 468, in <module>
      fds = reduction.recvfds(s, MAXFDS_TO_SEND + 1)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/C/_/python/src/Python-3.12.4/Lib/multiprocessing/reduction.py", line 164, in recvfds
      exit_status = int(not main())
      raise RuntimeError('received %d items of ancdata' %
  RuntimeError: received 0 items of ancdata
                            ^^^^^^
    File "/C/_/python/pkg/python/usr/lib/python3.12/compileall.py", line 445, in main
      if not compile_dir(dest, maxlevels, args.ddir,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/C/_/python/pkg/python/usr/lib/python3.12/compileall.py", line 109, in compile_dir
      results = executor.map(partial(compile_file,
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/C/_/python/pkg/python/usr/lib/python3.12/concurrent/futures/process.py", line 859, in map
      results = super().map(partial(_process_chunk, fn),
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/C/_/python/pkg/python/usr/lib/python3.12/concurrent/futures/_base.py", line 608, in map
      fs = [self.submit(fn, *args) for args in zip(*iterables)]
            ^^^^^^^^^^^^^^^^^^^^^^
    File "/C/_/python/pkg/python/usr/lib/python3.12/concurrent/futures/process.py", line 830, in submit
      self._adjust_process_count()
    File "/C/_/python/pkg/python/usr/lib/python3.12/concurrent/futures/process.py", line 789, in _adjust_process_count
      self._spawn_process()
    File "/C/_/python/pkg/python/usr/lib/python3.12/concurrent/futures/process.py", line 807, in _spawn_process
      p.start()
    File "/C/_/python/pkg/python/usr/lib/python3.12/multiprocessing/process.py", line 121, in start
      self._popen = self._Popen(self)
                    ^^^^^^^^^^^^^^^^^
    File "/C/_/python/pkg/python/usr/lib/python3.12/multiprocessing/context.py", line 301, in _Popen
      return Popen(process_obj)
             ^^^^^^^^^^^^^^^^^^
    File "/C/_/python/pkg/python/usr/lib/python3.12/multiprocessing/popen_forkserver.py", line 35, in __init__
      super().__init__(process_obj)
    File "/C/_/python/pkg/python/usr/lib/python3.12/multiprocessing/popen_fork.py", line 19, in __init__
      self._launch(process_obj)
    File "/C/_/python/pkg/python/usr/lib/python3.12/multiprocessing/popen_forkserver.py", line 57, in _launch
      with open(w, 'wb', closefd=True) as f:
  BrokenPipeError: [Errno 32] Broken pipe
  make: [Makefile:2423: libinstall] Error 1 (ignored)
```

jeremyd2019:
I did a hack to change `-j0` to `-j1` in COMPILEALL_FLAGS in Makefile.pre.in, to disable using multiprocessing, and it seems to have successfully run those compileall scripts in that case, so there may be some issue with multiprocessing.

lazka:
ah, I didn't notice the error, thanks. Seems like the receiving end crashed (?). A simple compileall (with the same options) seems to work locally but I haven't tested the python build.

yeah, it's really annoying, there was a conflict with cython too: https://github.com/msys2/MSYS2-packages/commit/2f9352502b362c598ac8a1bbe3342ea6009d1e37

jeremyd2019:
I was thinking maybe the RuntimeError was from the other end of the pipe/the child process? `RuntimeError: received 0 items of ancdata`

jeremyd2019:
In retrospect, I probably should have opened a new issue for this...

I originally thought it was a 32-bit imagebase conflict thing, because that would generally manifest as a failure during compileall, and would result in missing .pyc files in the package, which could cause file conflicts on future package updates.  But this doesn't seem to be an imagebase issue.

However, the 64-bit package is missing many .pyc files (compared to a 32-bit package built with `COMPILEALL_FLAGS=-j1`) so it will probably also have to be on the lookout for future file conflicts.

jeremyd2019:
I don't really know what I'm looking at, but https://github.com/msys2/msys2-runtime/blob/28d69fba269dd4a9f4281f8af7c2775292241e8b/winsup/cygwin/fhandler/socket_unix.cc#L1933C14-L1933C26 and https://github.com/msys2/msys2-runtime/blob/28d69fba269dd4a9f4281f8af7c2775292241e8b/winsup/cygwin/fhandler/socket_local.cc#L1364.  I *think* we're talking about socket_unix (the python code is using AF_UNIX), and it does try to use sendmsg.  No idea how that would have worked for you once installed 🤷

lazka:
Ah, I was passing the arguments wrong which made it ignore the path and use one worker..

jeremyd2019:
strace is showing fhandler_socket_local not unix, but stilll it seems descriptor passing is not implemented 😦

lazka:
yeah, I'll try to switch things to "spawn" by default

jeremyd2019:
Kind of unfortunate that `HAVE_SEND_HANDLE` is evaluating true (https://github.com/python/cpython/blob/da7f6745b81b229d0f01dc07a1ebc0374c95bf94/Lib/multiprocessing/reduction.py#L24-L27) probably because Cygwin defines all the necessary defines but just doesn't implement sending handles.

lazka:
See #4765

lazka:
I'll write a NEWS entry for the file conflicts

lazka:
I rushed this a bit, to avoid more fallout. Feedback still welcome.

jeremyd2019:
Turns out in the python rebuilds for i686, only python itself needed to be built locally (compileall still wound up leaving out pyc files in gha).  All the other packages built successfully in GHA.  I seem to remember one or two python-using packages that used to fail to build in gha, something scm related like mercurial or python-dulwich or subversion, but they all seemed to build ok this time. 🎉

PR #4744, opened 2024-07-04
Update gdbinit

jinger7281:
fix wrong path of `libstdcxx` python script
If the version number of `gcc` was updated, but this script was not update the variable of `%GCC_NAME%` which will lead an error of missing `libstdcxx` module
This update use python script to search the real path of the `gcc` folder and insert it into the `sys.path` variable

### DIFF
--- a/gdb/gdbinit
+++ b/gdb/gdbinit
@@ -1,6 +1,15 @@
 python
 import sys
-sys.path.insert(0, sys.path[0] + '/../../%GCC_NAME%/python')
+import os
+from os import path
+dir_of_gcc = list(filter(lambda x: x.startswith('gcc'), os.listdir('/usr/share')))
+if len(dir_of_gcc) > 0:
+  dir_of_gcc = path.join('/usr/share', dir_of_gcc[0], 'python')
+  if not path.exists(dir_of_gcc):
+    dir_of_gcc = sys.path[0]
+else:
+  dir_of_gcc = sys.path[0]
+sys.path.insert(0, dir_of_gcc)
 from libstdcxx.v6.printers import register_libstdcxx_printers
 register_libstdcxx_printers (None)
 end


lazka:
Sounds good. One thing I noticed is that have multiple dirs left because of python cache files (we should try fix this with the next gcc update). A hacky workaround is to sort the dirs, which is a good idea anyway since listdir order isn't stable.

What about:

```python
import sys
import glob
python_dirs = sorted(glob.glob("/usr/share/gcc*/python"), reverse=True)
sys.path.insert(0, (python_dirs or [sys.path[0])[0])
```

lazka:
Or we add a warning to the gcc build file that gdb needs to be rebuilt on version changes.

lazka:
I've created https://github.com/msys2/MSYS2-packages/pull/4748 as an alternative

jinger7281:
> Sounds good. One thing I noticed is that have multiple dirs left because of python cache files (we should try fix this with the next gcc update).
> 
> ```
> $ echo /usr/share/gcc*
> /usr/share/gcc-11.3.0 /usr/share/gcc-13.2.0 /usr/share/gcc-13.3.0
> ```
> 
> A hacky workaround is to sort the dirs, which is a good idea anyway since listdir order isn't stable.
> 
> What about:
> 
> ```python
> import sys
> import glob
> python_dirs = sorted(glob.glob("/usr/share/gcc*/python"), reverse=True)
> sys.path.insert(0, (python_dirs or [sys.path[0])[0])
> ```

This may be a good idea. Some newbies like me are always confused about the prompt, and don't know how to deal with this, we added this method to provide a way for newbies to easily use it.
```python
> import sys
> import glob
> python_dirs = sorted(glob.glob("/usr/share/gcc*/python"), reverse=True)
> sys.path.insert(0, (python_dirs or sys.path)[0])
```

PR #4745, opened 2024-07-04
mintty: default to BoldAsFont=no

lazka:
This is of course subjective, but at least on my low-dpi screen the default rendering of the bold font makes things harder to read.

On linux it looks nicer (not sure if rendering or font is the reason), and the Windows terminal doesn't do bold.

Screenshot 2024-07-04 234328

Screenshot 2024-07-04 234351

### DIFF
--- a/mintty/PKGBUILD
+++ b/mintty/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=mintty
 pkgver=3.7.3
-pkgrel=1
+pkgrel=2
 epoch=1
 pkgdesc="Terminal emulator with native Windows look and feel"
 arch=('i686' 'x86_64')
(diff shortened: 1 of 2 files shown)


Biswa96:
> This is of course subjective

Indeed, this is subjective. Probably it depends on the font and its size. Bold font works fine in my setup. I enabled that since the first time I used msys/cygwin/mintty.

<details><summary>Click here to see the screenshot of mintty window in my setup</summary>

mintty

</details>

raedrizqie:
it also depends on your screen monitor size

lazka:
thanks for the feedback.

yeah, I'm probably on the smaller font size/screen size side of things

lazka:
Let's leave this as is then.

PR #4758, opened 2024-07-06
glib2: Update to 2.80.3

lazka:
* drop the docs package: it moved from gtk-doc to gi-docgen which requires introspection, but we don't have the tooling for that yet
* small patch to fix the build
* it moved from distutils to packaging for version parsing
* rst2man from python-docutils is now used for man pages
* move gobject-query to devel to match Debian
* move gresource from devel, to match Debian
* install man pages
* xslt seems to no longer be used
* install some missing binaries (gio, gio-launch-desktop)
* install missing gettext files
* add conflicts since files have moved

### DIFF
--- a/glib2/002-fix-build.patch
+++ b/glib2/002-fix-build.patch
@@ -1,0 +1,13 @@
+diff --git a/glib/glib-private.h b/glib/glib-private.h
+index 7ab2579b0..3318e4c91 100644
+--- a/glib/glib-private.h
++++ b/glib/glib-private.h
+@@ -64,7 +64,7 @@
+  * as we'd like: https://stackoverflow.com/a/11529277/210151 and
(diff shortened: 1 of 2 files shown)


ognevny:
this adds file to `libexec`: `+/usr/libexec/gio-launch-desktop.exe`

lazka:
Is that a problem?

ognevny:
> Is that a problem?

it's better to get a rid of it IMO

lazka:
Why?

ognevny:
> Why?

lib is usually preferred over libexec

lazka:
So you want to get rid of it, or move it? If it's a problem we need to file a bug with glib.

ognevny:
> So you want to get rid of it, or move it? If it's a problem we need to file a bug with glib.

I meant do libexec -> lib move (or patch it to install to lib)

PR #4762, opened 2024-07-06
python: PEP 668 - mark as externally managed

lazka:
See https://peps.python.org/pep-0668/

Similar to what Arch does.

```console
$ pip install pip
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try 'pacman -S
    python-xyz', where xyz is the package you are trying to
    install.

    If you wish to install a non-MSYS2-packaged Python package,
    create a virtual environment using 'python -m venv path/to/venv'.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

### DIFF
--- a/python/EXTERNALLY-MANAGED
+++ b/python/EXTERNALLY-MANAGED
@@ -1,0 +1,8 @@
+[externally-managed]
+Error=To install Python packages system-wide, try 'pacman -S
+ python-xyz', where xyz is the package you are trying to
+ install.
+
+ If you wish to install a non-MSYS2-packaged Python package,
+ create a virtual environment using 'python -m venv path/to/venv'.
+ Then use path/to/venv/bin/python and path/to/venv/bin/pip.
--- a/python/PKGBUILD
+++ b/python/PKGBUILD
@@ -4,7 +4,7 @@
 pkgbase=python
 pkgname=('python' 'python-devel')
 pkgver=3.12.4
-pkgrel=4
+pkgrel=5
 _pybasever=${pkgver%.*}
 pkgdesc="Next generation of the python high-level scripting language"
 arch=('i686' 'x86_64')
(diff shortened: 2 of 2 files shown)


lazka:
Let's see if anyone complains :)

eabase:
I complain. 
> To install Python packages system-wide, try 'pacman -S ...`

What does this do, exactly?

System Wide for me means usable on windows PC from anywhere. basically same as installing python for Windows, and it doesn't seem appropriate behavior. MSYS2 should stay in it's box.

lazka:
system wide in the context of Python means installing into system-site-packages, which is `/usr/lib/python3.12/site-packages`

eabase:
> system-site-packages, which is /usr/lib/python3.12/site-packages

I'm on a Windows PC, so what is the path of that? 
Because MSYS2 already has it's own path like that in:  
`C:\msys64\usr\lib\python3.12\site-packages`

But now you're saying it is installed in:  
`C:\Python312\Lib\site-packages`

^ **THAT* is for me the "`system-site-packages`".

lazka:
To get a native Windows path that would be:

```console
$ cygpath -m  /usr/lib/python3.12/site-packages
C:/msys64/usr/lib/python3.12/site-packages
```

eabase:
Yes, but this is not exposed to my OS, so this path is not a (windows) "system-site-packages" path, only within the MSYS2 environment as exposed under `C:/msys64/`.

jeremyd2019:
Maybe the message could be improved somehow, perhaps "system-wide" is too wide on Windows, where there is no single system-wide python installation per se.  I'm not sure what would be better, maybe something about "globally for this python installation" instead of "system-wide"?

PR #4764, opened 2024-07-08
msys2-runtime-3.4: synchronize with the `msys2-runtime-3.4.10` branch in the `msys2-runtime` repository

dscho:
Seems that we haven't updated this package in a while...

NOTE: There seems to be [a compile error](https://github.com/msys2/msys2-runtime/actions/runs/9835525686/job/27149328693#step:4:1495) that I still need to look into, hence the draft status of this PR.

### DIFF
--- a/msys2-runtime-3.4/update-patches.sh
+++ b/msys2-runtime-3.4/update-patches.sh
@@ -75,6 +75,7 @@ then
 	exit 0
 fi
 
+GIT_CONFIG_PARAMETERS="${GIT_CONFIG_PARAMETERS+$GIT_CONFIG_PARAMETERS }'core.autocrlf=false'" \
 updpkgsums ||
 die "Could not update the patch set checksums in PKGBUILD"
 
--- a/msys2-runtime-3.4/update-patches.sh
+++ b/msys2-runtime-3.4/update-patches.sh
@@ -21,8 +21,10 @@ base_tag=refs/tags/cygwin-"$(sed -ne 's/^pkgver=//p' <PKGBUILD)"
 msys2_branch=refs/heads/msys2-${base_tag#refs/tags/cygwin-}
 url=https://github.com/msys2/msys2-runtime
 
-test -d msys2-runtime ||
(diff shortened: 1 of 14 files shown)


dscho:
> NOTE: There seems to be [a compile error](https://github.com/msys2/msys2-runtime/actions/runs/9835525686/job/27149328693#step:4:1495) that I still need to look into, hence the draft status of this PR.

There are two problems at hand:
1. As of Pacman v6.1.0, [the checksum of the Git revision is checked and not `SKIP`ped](https://github.com/msys2/MSYS2-packages/actions/runs/9836026425/job/27150879968?pr=4764#step:10:424). I have a patch in hand, but want to push only when both problems are solved.
2. The build failure was addressed in the `msys2-3.5.3` branch via https://github.com/msys2/msys2-runtime/pull/215, and in Cygwin via https://github.com/msys2/msys2-runtime/commit/7e3c833592b282355a57dd34459b152e4e078d19. I opened https://github.com/msys2/msys2-runtime/pull/220 to backport that patch, and once that latter PR is merged, I will update this here PR and that should let the PR build pass.

msys2-runtime-3.4/update-patches.sh
```diff
@@ -21,8 +21,10 @@ base_tag=refs/tags/cygwin-"$(sed -ne 's/^pkgver=//p' <PKGBUILD)"
 msys2_branch=refs/heads/msys2-${base_tag#refs/tags/cygwin-}
 url=https://github.com/msys2/msys2-runtime
 
-test -d msys2-runtime ||
-git clone --bare $url msys2-runtime ||
+test -d msys2-runtime || {
```
lazka, line 24: I wonder if it wouldn't be easier to just name the clone differently to avoid any conflicts. But maybe I'm missing something.

dscho: I do want to reuse the clone, as it is not exactly a small repository.

lazka: ok, fair enough

PR #4765, opened 2024-07-08
python: switch default mp start method to "spawn" and disable "forkserver"

lazka:
"forkserver" doesn't work, and "fork" is deprecated, so give "spawn" a try See https://github.com/msys2/MSYS2-packages/pull/4743#issuecomment-2213043559


lazka:
See https://www.msys2.org/news/#2024-07-08-file-conflicts-when-updating-python

piotrvv:
May I know why make it so difficult to build python?

lazka:
What do you mean?

piotrvv:
I'm asking this because msys-*.dll files are larger in the sources like in the pacman package.

jeremyd2019:
> May I know why make it so difficult to build python?

I think both cygwin and mingw are unsupported targets upstream, so we have to figure out how to patch things ourselves.  Also, it seems like many times we are ahead of cygwin's versions of packages, so we are generally on the leading edge of figuring this stuff out (I think cygwin is still on python 3.9).

PR #4773, opened 2024-07-09
Prevent different msys2 runtime versions from sharing cygheaps (targeting msys2-runtime v3.5)

dscho:
This corresponds to https://github.com/msys2/msys2-runtime/pull/219.


lazka:
thanks!

PR #4787, opened 2024-07-20
libassuan: Update to 3.0.2

lazka:


lazka:
I wonder if the ABI break can cause a problem during upgrades

lazka:
See https://github.com/msys2/MSYS2-packages/pull/5244#issuecomment-2713029125

apparently ABI didn't break as much to keep a rename working.

lazka:
went for building both, just to be sure

PR #4797, opened 2024-07-26
tmux: fix startup in local session

rkitover:
Add a patch to fix tmux failing to start in a local terminal session when using Microsoft Terminal and other Windows terminals.

See:

https://github.com/microsoft/terminal/issues/5132

### DIFF
--- a/tmux/PKGBUILD
+++ b/tmux/PKGBUILD
@@ -4,7 +4,7 @@ pkgname=tmux
 _base_ver=3.4
 _bugfix=
 pkgver=${_base_ver}${_bugfix}
-pkgrel=2
+pkgrel=3
 pkgdesc='A terminal multiplexer'
 url='https://tmux.github.io/'
 msys2_repository_url="https://github.com/tmux/tmux"
@@ -22,18 +22,21 @@ makedepends=('ncurses-devel' 'libevent-devel' 'autotools' 'gcc')
 source=(${pkgname}-${pkgver}.tar.gz::https://github.com/tmux/tmux/archive/${_base_ver}${_bugfix/./}.tar.gz
         'msys-platform.patch'
         'no-check-dirs-permission.patch'
-        'inherit-msystem-var.patch')
+        'inherit-msystem-var.patch'
+        'pass-terminal-data-on-socket-on-Cygwin-platforms.patch')
 
 sha256sums=('ec7ddf021a0a1d3778862feb845fd0c02759dcdb37ba5380ba4e0038164f7187'
             'cf7e0f642d43fc97f7c55be41c693cceab5c891f81fcee687e918075e55cd533'
             '31f3c9f1b98a97788edca46fb7a0ba2b5328758f014d445096561c8466540c5d'
-            'aeaea034eeefe9598a56ebfabdf21cb991b2ac39d62dca9f857d24e5bc375366')
+            'aeaea034eeefe9598a56ebfabdf21cb991b2ac39d62dca9f857d24e5bc375366'
+            '46727f30ad9fa8671348077f8025875bd8a82f7c72ea590f2f3b38c7d2432845')
 
 prepare() {
   cd "${srcdir}/${pkgname}-${_base_ver}${_bugfix/./}"
   patch -p1 -i "${srcdir}"/msys-platform.patch
   patch -p0 -i "${srcdir}"/no-check-dirs-permission.patch
   patch -p0 -i "${srcdir}"/inherit-msystem-var.patch
+  patch -p1 -i "${srcdir}"/pass-terminal-data-on-socket-on-Cygwin-platforms.patch
 
   cp osdep-cygwin.c osdep-msys.c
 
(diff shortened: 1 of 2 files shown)


jeremyd2019:
It looks like there have been some fixes to cygwin for tmux since the last release (eg https://github.com/cygwin/cygwin/commit/186ee9b67607deeb19993385bc2c07ba20050237).  Perhaps they will make this patch unnecessary, once they make a new release.

rkitover:
Not possible, tmux has the client send the terminal file descriptor over the client <-> server UNIX domain socket, which is possible on UNIX, but not possible on Cygwin because it emulates UNIX sockets using a localhost TCP socket.

jeremyd2019:
OK I guess I just don't understand then.  I thought that tmux mostly worked, except when run from a Windows console (as opposed to a Cygwin pty from the likes of mintty), and that recent fixes to cygwin would help in that scenario.  However, I do know that Cygwin does not support passing file descriptors over unix-domain sockets

rkitover:
I will explain.

See [here](https://github.com/tmux/tmux/blob/master/server-client.c#L3043) in the upstream sources.

What this does, is set the server client handler terminal file descriptor to the client terminal device path, since the code above that to receive the terminal file descriptor from the client does nothing.

In mintty or ssh, the client uses a Cygwin pty device, so it is possible for the server process to access it directly.

In a local native Windows terminal session, such as in Microsoft Terminal, the client terminal device is a special file that is not accessible from another process, such as the tmux server process, by any means. Windows does not allow that.

This patch is very simple (it was however, a lot of research and work for me to make.) What it does is set up a pty device for the server client handler, and pass data between it and the actual client terminal on the socket using libevent handlers.

jeremyd2019:
> In a local native Windows terminal session, such as in Microsoft Terminal, the client terminal device is a special file that is not accessible from another process, such as the tmux server process, by any means. Windows does not allow that.

It seemed like @tyan0 had something in place now in Cygwin to allow that.

rkitover:
> It seemed like @tyan0 had something in place now in Cygwin to allow that.

Do you have any more details? Even if that was the case, tmux would still have to be patched for this, as you can see from my explanation above.

And why would this be better than my solution? I think it's clean and simple. I have a PR pending in tmux upstream for it.

tyan0:
At least for cygwin, tmux works on cygwin 3.5.0, 3.5.1 and 3.5.4 (which will be released soon) in windows terminal. This is because https://github.com/cygwin/cygwin/commit/3721a756b0d8067cf070e2aedcf3ea5e32e47270 makes the console accessible from other terminals like ptys.

Unfortunately, 3.5.3 has a regression due to the commit https://github.com/cygwin/cygwin/commit/cf121e088f928c68a8e3c799db97478c89b6dd96 which prevents tmux from starting in a console. This is already fixed by https://github.com/cygwin/cygwin/commit/186ee9b67607deeb19993385bc2c07ba20050237, so tmux will work again in 3.5.4.

I am not sure what problem this PR fixes, but could you please check if tmux works on 3.5.1 with your scenario?

tyan0:
msys2-runtime 3.5.1 can be downloaded from:
https://repo.msys2.org/msys/x86_64/msys2-runtime-3.5-3.5.1-2-x86_64.pkg.tar.zst
https://repo.msys2.org/msys/x86_64/msys2-runtime-3.5-devel-3.5.1-2-x86_64.pkg.tar.zst

and can be installed by:
```pacman -U msys2-runtime-3.5-3.5.1-2-x86_64.pkg.tar.zst msys2-runtime-3.5-devel-3.5.1-2-x86_64.pkg.tar.zst```

rkitover:
@tyan0 question for you, is this new terminal device visible outside of the Cygwin virtual filesystem, like in the scenario where you have a `tmux.exe` bundled with the Cygwin dll and other dependent dlls.

tyan0:
Is your question whether the console device is accessible from another installation of cygwin? If so, the answer is no.

rkitover:
Not exactly.

My question is whether the console device is accessible **WITHOUT** an installation of Cygwin.

You may be aware of some Windows software that are not full native ports and use an msys or cygwin dll for a part of their functionality, sshfs for Windows is one such example. This software typically bundles the msys/cygwin dll in the same directory as the exetuable with any other dependent dlls.

In this scenario, there is no full virtual filesystem, like when running a program from an msys/cygwin installation.

My question is whether the console device is accessible in this scenario, because I am working on a "more" native package for tmux made in this way.

tyan0:
Yes. The console devices (/dev/cons*) and their functionalities are provided by cygwin1.dll (msys-2.0.dll), so the program loading cygwin1.dll (msys-2.0.dll) can access to other consoles if exist (i.e. some program using that dll has been opened in the console device, such as Windows Terminal, cmd.exe).

rkitover:
Don't know what happened here or if this is the actual patch that I sent, anyway I'm closing this.

@tyan0 thank you for your work on the Cygwin runtime for this and thank you for explaining how this works to me.

PR #4810, opened 2024-08-05
mingw-w64-cross-winpthreads: convert to split package

lazka:


mingw-w64-cross-winpthreads/PKGBUILD
```diff
@@ -43,23 +39,60 @@ prepare() {
   autoreconf -vfi
 }
 
+_build() {
+  _target=$1
+  mkdir -p ${srcdir}/winpthreads-build-${_target} && cd ${srcdir}/winpthreads-build-${_target}
+  unset CC
+  ${srcdir}/mingw-w64/mingw-w64-libraries/winpthreads/configure \
```
Blackhex, line 46: I am missing some sort of CRT selection. Wouldn't it pick the first found? Wouldn't the content of the `ucrt64`, `mingw32`, `mingw64` package variants be identical?

mingw-w64-cross-winpthreads/PKGBUILD
```diff
@@ -43,23 +39,60 @@ prepare() {
...
+  conflicts=("${_mingw_suff}-${_realname}-git")
+  replaces=("${_mingw_suff}-${_realname}-git")
 }
 
-package() {
-  for _target in ${_targets}; do
-    cd ${srcdir}/winpthreads-build-${_target}
-    make DESTDIR=${pkgdir} install
-  done
+package_mingw-w64-cross-ucrt64-winpthreads() {
```
Blackhex, line 79: Is there a possibility to build only one variant of the package? If all variants need to be build at once, it will require porting/bootstrapping all those variants for aarch64...

mingw-w64-cross-winpthreads/PKGBUILD
lazka: From what I understand the `--host=${_target}` below makes it pick the right gcc/toolchain, which implies the crt etc.

Blackhex: OK, it's probable it works like that.

mingw-w64-cross-winpthreads/PKGBUILD
lazka: I could check add a check in _build() for which package is defined in the pkgname array. Then it would only build/package what is defined in pkgname. Would that help?

Blackhex: Yes, it would make the necessary changes minimal.

lazka: You could also have your own PKGBUILD building just aarch64

lazka: btw, after this is all done I would look into integrating https://github.com/Windows-on-ARM-Experiments/MSYS2-packages/commits/woarm64/

Blackhex: Thank you. Any help would be appreciated.

PR #4812, opened 2024-08-06
fish PKGBUILD: remove ancient dependencies

zanchey:
There are a number of very old dependencies listed here, which should be able to be removed without any change in the function of the final binary.

### DIFF
--- a/fish/PKGBUILD
+++ b/fish/PKGBUILD
@@ -11,7 +11,7 @@ msys2_references=(
 )
 license=('GPL2')
 groups=()
-depends=('gcc-libs' 'gettext' 'libiconv' 'libpcre2_16' 'man-db' 'ncurses')
+depends=('gcc-libs' 'gettext' 'libpcre2_16' 'man-db' 'ncurses')
 makedepends=('gcc' 'gettext-devel' 'intltool' 'ncurses-devel' 'pcre2-devel' 'cmake')
 optdepends=('python: for manual page completion parser and web configuration tool')
 install=fish.install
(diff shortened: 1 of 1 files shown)


fish/PKGBUILD
```diff
@@ -11,8 +11,8 @@ msys2_references=(
 )
 license=('GPL2')
 groups=()
-depends=('bc' 'gcc-libs' 'gettext' 'libiconv' 'libpcre2_16' 'man-db' 'ncurses')
-makedepends=('doxygen' 'gcc' 'gettext-devel' 'intltool' 'libiconv-devel' 'ncurses-devel' 'pcre2-devel' 'cmake')
```
lazka, line 15: Why was libiconv-devel removed?

zanchey: fish used to depend on iconv for Unicode support, but replaced it with a modified version of https://www.bsdua.org/libbsdua.html which is vendored internally in 2014. I'll drop it from depends as well, thanks for the pickup!

lazka: It still links against libiconv: https://github.com/msys2/MSYS2-packages/pull/4812/checks#step:10:950

lazka: hm, it seems ntldd and Dependencies seem to now show libiconv being linked, unlike ldd. So maybe ldd is buggy somehow there.

lazka: I got confused by ntldd only showing direct deps that I forgot about ldd showing all of them, sorry for the confusion.

PR #4826, opened 2024-08-11
libtool: update to 2.5.1

ikspress:
The propose of this PR is as follows.

1. Rewrite PKGBUILD to keep it same as mingw-w64-libtool/PKGBUILD
2. Run `autoconf -f` to regenerate `configure` with MSYS2 support
3. Install license files


ikspress:
autoconf complained
```text
aclocal.m4:17: warning: this file was generated for autoconf 2.72e.
You have another version of autoconf.  It may work, but is not guaranteed to.
If you have problems, you may need to regenerate the build system entirely.
To do so, use the procedure documented by the package, typically 'autoreconf'.
```
However, I think we can just ignore it.

PR #4850, opened 2024-08-20
makepkg-mingw.conf: add RUSTFLAGS

ognevny:
frame-pointers are already enabled for rust package, so let's put it for all rust-based packages

push repo checksum while at it

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=pacman
 pkgver=6.1.0
-pkgrel=4
+pkgrel=5
 pkgdesc="A library-based package manager with dependency support (MSYS2 port)"
 arch=('i686' 'x86_64')
 url="https://www.archlinux.org/pacman/"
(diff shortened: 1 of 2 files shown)


Biswa96:
Did you try to compile and use any rust packages with this change?

ognevny:
> Did you try to compile and use any rust packages with this change?

rust compiler works, you mean _other_ programs? I can test Zed

Biswa96:
I meant other programs, just normal use cases.

ognevny:
Zed works

lazka:
Arch does it too, so fine I guess: https://gitlab.archlinux.org/archlinux/packaging/packages/pacman/-/blob/4950709a4fd95172c6ffa04abd78b0fce8ff4591/rust.conf

ognevny:
> Arch does it too, so fine I guess: https://gitlab.archlinux.org/archlinux/packaging/packages/pacman/-/blob/4950709a4fd95172c6ffa04abd78b0fce8ff4591/rust.conf

Zed is built with this flag for more than a week, and for now the only new issues about are known (not related to such change)

PR #4857, opened 2024-08-28
Fix spelling error in dot.bash_profile

emuslor:


lazka:
sha256sums in the PKGBUILD needs to be updated

emuslor:
> sha256sums in the PKGBUILD needs to be updated

Oops!  Got 'er - thanks.

PR #4872, opened 2024-09-01
rhash: update to 1.4.4

emuslor:
Draft to test action...

Update to 1.4.4:
- Remove patch for obsolete timing lib.
- Add proper `ln -s` definition for lib build.
- Correct flags definition.


lazka:
> @lazka I noticed rhash supports gettext if the libs are present at build time. What's the decider on adding (or not adding) intl/iconv as package dependencies for MSYS2 when optional?

I can't think of any reason. If you think it's useful feel free to add.

emuslor:
> > @lazka I noticed rhash supports gettext if the libs are present at build time. What's the decider on adding (or not adding) intl/iconv as package dependencies for MSYS2 when optional?
> 
> I can't think of any reason. If you think it's useful feel free to add.

This is ready to go, unless you have concerns, or some of it needs to be submitted upstream, etc.

PR #4885, opened 2024-09-06
ncurses: enable term-driver for TERM='#win32con'

rkitover:
Add a patch to enable the term-driver support for the Win32 console with TERM='#win32con', which does not require a terminfo database.

Does not change the regular terminfo-based functionality of ncurses otherwise.

This patch has been submitted upstream and this is the rebased version for 6.5.

You can verify that this works by running an ncurses program with TERM='#win32con' in a Windows Terminal session or similar. For example this one:

https://github.com/mtoyoda/sl
.

### DIFF
--- a/ncurses/PKGBUILD
+++ b/ncurses/PKGBUILD
@@ -3,7 +3,7 @@
 pkgbase=ncurses
 pkgname=('ncurses' 'ncurses-devel')
 pkgver=6.5
-pkgrel=1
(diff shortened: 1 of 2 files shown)


rkitover:
Never mind, I did not see that this has been merged upstream and is in the current ncurses update patch, I will resubmit this with the ncurses update patches instead.

PR #4886, opened 2024-09-06
ncurses: update to 6.5-20240831,enable term-driver

rkitover:
Update to the current dev version, as the last stable release was already two years ago.

This patch set includes a patch to enable the term-driver support for the Win32 console with TERM='#win32con' on the Cygwin runtime, like it does for MinGW, which does not require a terminfo database.

This does not change the regular terminfo-based functionality of ncurses otherwise.

You can verify that this works by running an ncurses program with TERM='#win32con' in a Windows Terminal session or similar. For example this one:

https://github.com/mtoyoda/sl
.

### DIFF
--- a/ncurses/PKGBUILD
+++ b/ncurses/PKGBUILD
@@ -2,7 +2,9 @@
 
 pkgbase=ncurses
 pkgname=('ncurses' 'ncurses-devel')
-pkgver=6.5
+_base_ver=6.5
+_date_rev=20240831
+pkgver=${_base_ver}.${_date_rev}
 pkgrel=1
 pkgdesc="System V Release 4.0 curses emulation library"
 arch=('i686' 'x86_64')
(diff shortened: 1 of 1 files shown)


ncurses/PKGBUILD
```diff
@@ -71,8 +72,8 @@ build() {
     --with-build-cflags=-D_XOPEN_SOURCE_EXTENDED \
     --with-pkg-config-libdir=/usr/lib/pkgconfig
 
-  make
-  make DESTDIR=${srcdir}/dest install
+  make -j$(nproc)
+  make -j$(nproc) DESTDIR=${srcdir}/dest install
```
lazka, line 76:
```suggestion
  make
  make DESTDIR=${srcdir}/dest install
```

we set that via MAKEFLAGS already

lazka:
Unfortunate that it doesn't just auto-detect the console and depends on TERM for this.

Let's see how it works.. :)

thanks!

lazka:
hm, I get `Error opening terminal: #win32con.`

rkitover:
@lazka I need to do a bit more research to confirm what I think is happening here and I will discuss on the discord, we'll get this sorted out.

lazka:
lol, hold your horses.

So, first off, does the package from the repo work for you?

rkitover:
I'm reasonably certain that it should and it does not.

Here is how you can verify that it has been sabotaged:

- `set  disassemble-next-line on`,
- break on `__imp_new_prescr`,
- use `ni`/`si` to get to the `calloc()` call in the runtime and check the return value in one of the registers, probably `%rax`, it will be zero.

I can't give you more exact instructions right now because these ****s are breaking my GDB.

Biswa96:
You can install a separate msys2 without disturbing the current installation.

rkitover:
@lazka So you said you got an error when you tried the term-driver, if that also happens with the repo package then I believe it has been sabotaged, and you can verify this using the instructions I posted.

rkitover:
@lazka @Biswa96 Here is my full account of what happened to this change:

https://github.com/tmux/tmux/pull/4054#issuecomment-2451971135
.

PR #4928, opened 2024-09-30
lua: add 5.4.7

lazka:
required for tio


lazka:
Compared to what other distros do and patch this is a very basic, so could be something is missing/not working. It's at least enough to get tio to build though.

PR #4946, opened 2024-10-04
ncurses: disable term-driver again to fix regressions

### LINKED ISSUE #4945: [ncurses] Nano editor sets wrong fg/bg colors
### Description / Steps to reproduce the issue

1. Set mintty theme to kohlrausch. This way it's more evident as kohlrausch is black on white theme. Though, with other themes (but not all) it's the same.
2. Run nano.

For the full story please read the thread following the links: [start](https://lists.gnu.org/archive/html/help-nano/2024-09/msg00005.html), [continuation](https://lists.gnu.org/archive/html/help-nano/2024-10/msg00000.html).

What helped is downgrading ncurses package from 6.5.20240831 to 6.5 version.

### Expected behavior

Nano respects theme colors.

### Actual behavior

Nano uses white foreground and black background colors.

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MSYS_NT-10.0-19045

### Are you willing to submit a PR?

_No response_

lazka:
It breaks colors with nano for some reason, see
https://github.com/msys2/MSYS2-packages/issues/4945

Fixes #4945

### DIFF
--- a/ncurses/PKGBUILD
+++ b/ncurses/PKGBUILD
@@ -5,7 +5,7 @@ pkgname=('ncurses' 'ncurses-devel')
 _base_ver=6.5
 _date_rev=20240831
 pkgver=${_base_ver}.${_date_rev}
-pkgrel=1
+pkgrel=2
 pkgdesc="System V Release 4.0 curses emulation library"
 arch=('i686' 'x86_64')
 url="https://www.gnu.org/software/ncurses/"
@@ -56,7 +56,7 @@ build() {
     --enable-sp-funcs \
     --with-wrap-prefix=ncwrap_ \
     --enable-sigwinch \
-    --enable-term-driver \
+    --disable-term-driver \
     --enable-colorfgbg \
     --enable-tcap-names \
     --disable-termcap \


rkitover:
@lazka 

If you look at my patch, you will see that it's impossible for there to be any regressrion:

https://mail.gnu.org/archive/html/bug-ncurses/2024-08/msg00009.html .

The patch does nothing except allow the term-driver code to be enabled, which has no effect otherwise.

Furthermore, I proved that the binary package that resulted from that PR was compromised here:

https://github.com/msys2/MSYS2-packages/pull/4886#issuecomment-2364224587

rkitover:
If the term-driver actually works now, I'd rather fix whatever the problem in Nano is, and if `$TERM` is set, I don't see how there could possibly be a problem in the first place.

lazka:
I can reproduce it breaking nano colors. Ideas how to fix things welcome. I'm happy to test them.

PR #4959, opened 2024-10-16
fix: nanorc.nanorc on nano-syntax-highlighting

luau-project:
## Description

There is a mistake in the package ```nano-syntax-highlighting```, specifically at line 26 in the file ```/usr/share/nano-syntax-highlighting/nanorc.nanorc``` (brightnormal is not allowed as icolor).

This issue has been reported a few times upstream, and it is also cited on [arch wiki](https://wiki.archlinux.org/title/Nano#Syntax_highlighting), but maintainer seems to be unreachable (his last github activity was mid 2023).

## Solution

Even though arch wiki says to replace 'brightnormal' with 'normal', it would redefine icolor normal, which
was defined a few lines above on that file. Instead, I decided to keep the original value defined earlier.

Then, I just commented that offending line.

## Reproduction

1. Install the package ```nano-syntax-highlighting```:

```bash
pacman -S nano-syntax-highlighting
```

2. Open the file ```/usr/share/nano-syntax-highlighting/nanorc.nanorc``` with ```nano```
```bash
nano /usr/share/nano-syntax-highlighting/nanorc.nanorc
```

## Actual Behavior

nanorc

## Expected Behavior

After applying the fix:

expected-nanorc

### DIFF
--- a/nano-syntax-highlighting/001-nanorc.patch
+++ b/nano-syntax-highlighting/001-nanorc.patch
@@ -1,0 +1,12 @@
+diff -Naur a/nanorc.nanorc b/nanorc.nanorc
+--- a/nanorc.nanorc	2020-10-10 16:25:18.000000000 -0300
++++ b/nanorc.nanorc	2024-10-16 13:16:04.816910360 -0300
+@@ -23,7 +23,7 @@
+ icolor brightmagenta " brightmagenta"
+ icolor brightcyan " brightcyan"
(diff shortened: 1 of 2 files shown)


lazka:
Arch moved to a fork, which might also be an option: https://gitlab.archlinux.org/archlinux/packaging/packages/nano-syntax-highlighting/-/commit/126c95619ca35a885e4869830a6347ad409af8b0

PR #4974, opened 2024-10-25
Add genromfs

jhol:


lazka:
just wondering, what do you need this for?

jhol:
[ROMFS](https://docs.kernel.org/filesystems/romfs.html) is a very simple filesystem used for small embedded devices such as routers, and is supported by Linux and a few other operating systems. We are using it as part of an embedded project we are working on.

jhol:
If you mind me asking, is this something that is likely to be accepted? I think the packaging is solid, but is there a question about whether this tool is in scope for MSYS2-packages?

lazka:
We normally avoid adding non-dev cygwin packages, especially code that hasn't been updated in 14 years ;)

But this seems very much self contained at least.. and we still have no official policy there.

jhol:
Thanks! I think it's worth having upstream.

As for updates: I would say the tool is simply done. It's simple and it does does everything it's meant to do.

PR #4980, opened 2024-10-28
autoconf: Treat Flang as GNU-compatible Fortran compiler.

mmuetzel:
This allows building more projects with LLVM Flang as the Fortran compiler without needing additional (obscure) configure flags.

Note of caution: I've never used Classic Flang. I don't know if it also defines `__flang__`. And if it does, I don't know if it can be considered as a "drop-in replacement" of gfortran (or other GNU Fortran compilers).

### DIFF
--- a/autoconf2.71/0004-LLVM-Flang.patch
+++ b/autoconf2.71/0004-LLVM-Flang.patch
@@ -1,0 +1,60 @@
+Treat LLVM Flang as GNU-compatible compiler for GNU targets.
+
+--- autoconf-2.71/lib/autoconf/fortran.m4.orig	2021-01-28 21:46:48.000000000 +0100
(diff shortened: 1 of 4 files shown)


mmuetzel:
I downloaded and installed the build artifacts from this PR.
With them (and the recent libtool update), I can configure Octave in the CLANG64 environment without `--enable-fortran-calling-convention=gfortran ac_cv_f77_compiler_gnu=yes lt_cv_prog_gnu_ld=yes` and it still configures and builds correctly.
It also still configures and builds correctly in the MINGW64 environment (like before).

I'll leave this open for a day or two for feedback before potentially merging it.

mmuetzel:
Come to think of it: LLVM Flang could potentially also target Windows MSVC.
It's probably not possible to check which target the flang compiler will use only with a preprocessor test.
I amended the PR to add a test for the target triplet of Flang.

PR #5003, opened 2024-11-13
base-devel: Remove pkgconf dependency.

MehdiChinoune:
It interferes with mingw-w64-pkgconf.

### DIFF
--- a/base-devel/PKGBUILD
+++ b/base-devel/PKGBUILD
@@ -1,8 +1,8 @@
 # Maintainer: Christoph Reiter <reiter.christoph@gmail.com>
 
 pkgname=base-devel
-pkgver=2022.12
-pkgrel=2
+pkgver=2024.11
+pkgrel=1
 pkgdesc='Minimal package set for building packages with makepkg'
 url='https://www.msys2.org'
 arch=('any')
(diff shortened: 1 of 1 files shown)


lazka:
Any details on the problems it causes? (in case mingw variant is missing?)

I don't mind removing it.

MehdiChinoune:
As an example with qt6-static, CMake picked msys2-pkgconf instead of mingw-w64-pkgconf.
msys2-pkgconf doesn't work with mingw packages the same as mingw-w64-pkgconf. which results in not detecting dbus dependency and fails to configure.

lazka:
huh, I see, weird

MehdiChinoune:
> huh, I see, weird. I wonder how it ended up picking the wrong one.

The issue when I exported `PKG_CONFIG="${MINGW_PREFIX}/bin/pkgconf.exe --static"`. Apparently (not sure, just guessing) msys2 shell didn't transform that path to a Windows path so Cmake saw as a wrong path and picked `/usr/bin/pkgconf`. When passing only the argument `PKG_CONFIG_ARGN="--static"`, CMake pickes the mingw-w64-pkgconf.

I want to make sure In case CMake or meson fail to detect mingw-w64-pkgconf they don't pick the MSYS2 one.

PR #5007, opened 2024-11-13
msys2-runtime: Update

lazka:
See https://github.com/msys2/msys2-runtime/pull/234


jeremyd2019:
Has there been an update to the update-patches script?  When I tried it for a local test build, it blew up on updpkgsums, is this something that happens in non-hacked-up scenarios too?

lazka:
this one worked fine for me, the 3.3 one errored out though and I manually finished it

jeremyd2019:
yeah, it was 3.3 cause my i686 build was easier to use a PKGBUILD for.

PR #5009, opened 2024-11-14
msys2-runtime-3.3: Update (suppress proc_waiter error messages)

jeremyd2019:
See https://github.com/msys2/msys2-runtime/pull/237

Also sync update-patches.sh with msys2-runtime, so it doesn't fail in updpkgsums (and filters out .github changes)


jeremyd2019:
I'm gonna go ahead and merge this.

PR #5015, opened 2024-11-21
msys2-runtime: integrate suspendthread fixup

dscho:
This corresponds to https://github.com/msys2/msys2-runtime/pull/239 and integrates one minor change of behavior where `remove_proc()` now _also_ tries `CancelSynchronousIo()` before using the big `terminate_thread()` hammer.

Probably not a very noticeable change of behavior. But still. Better stay aligned with Cygwin!


jeremyd2019:
This is not merged yet, so at least I know the hang is not the additional CancelSynchronousIo call...

Issue #4340, opened 2024-01-11
[pacman] getting stuck on clang-aarch64


Wormnest:
### Description / Steps to reproduce the issue

Since about a month GIMP's aarch64 CI runner is getting stuck when running `pacman --noconfirm -Suy`.
The last job that succeeded was Dec 11 and another from the same day and any later one is failing.

Most of the time (e.g. [here](https://gitlab.gnome.org/GNOME/gimp/-/jobs/3445501)) it already seems to stop before the databases are updated:
```
$ C:\msys64\usr\bin\bash -lc "bash -x ./build/windows/gitlab-ci/1_build-deps-msys2.sh"
+ set -e
+ [[ aarch64 == \a\a\r\c\h\6\4 ]]
+ export ARTIFACTS_SUFFIX=-a64
+ ARTIFACTS_SUFFIX=-a64
+ [[ CI_NATIVE != \C\I\_\N\A\T\I\V\E ]]
+ pacman --noconfirm -Suy
WARNING: Failed to terminate process: 1 error occurred:
	* failed to attach the runner process to the console of its parent process: The handle is invalid.
WARNING: Timed out waiting for the build to finish
```

[Sometimes](https://gitlab.gnome.org/GNOME/gimp/-/jobs/3441058) it gets a little further:
```
$ C:\msys64\usr\bin\bash -lc "bash -x ./build/windows/gitlab-ci/1_build-deps-msys2.sh"
+ set -e
+ [[ aarch64 == \a\a\r\c\h\6\4 ]]
+ export ARTIFACTS_SUFFIX=-a64
+ ARTIFACTS_SUFFIX=-a64
+ [[ CI_NATIVE != \C\I\_\N\A\T\I\V\E ]]
+ pacman --noconfirm -Suy
:: Synchronizing package databases...
 clangarm64 downloading...
 msys downloading...
:: Starting core system upgrade...
 there is nothing to do
:: Starting full system upgrade...
 there is nothing to do
[...]

PR #5019, opened 2024-11-23
msys2-runtime: revert use of CancelSyncronousIo on wait_thread

dscho:
This change seems to have caused hangs on x86_64, so let's revert it.

Addresses https://github.com/msys2/MSYS2-packages/issues/4340#issuecomment-2491401847 and corresponds to https://github.com/msys2/msys2-runtime/pull/243.


jeremyd2019:
6614da9a71fc5cf24991b47fd62a56749c7f4ccc

PR #5024, opened 2024-11-26
sqlite: Update to 3.47.2

lazka:


MehdiChinoune:
You could disable tcl plugin. https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-sqlite3/0003-do-not-install-tclsqlite3-with-make.patch

jannick0:
@lazka Hi, happy to have a look a that, but that might last a bit, since I am currently running out of time with other stuff.

lazka:
I've created #5024 now for a newer version. I dropped all the extension code because I'm not quite sure how to port that. Ideas welcome.

lazka:
added extensions too now. I merged it already, but feedback always welcome.

PR #5046, opened 2024-12-11
autoconf2.72: Don't strip MinGW libraries from Fortran linker flags

mmuetzel:
Stripping `-lmingwex` from the Fortran standard linker flags can lead to a situation where the first `-lmsvcrt` appears before the first `-lmingwex` for a library that consists of Fortran and C++ sources. That means that the linker picks some functions from the Windows CRT that should instead be picked from `libmingwex.a`.

It took quite some time to figure out that this is causing an odd behavior when calling `std::asin` with a complex double precision value with a real part larger than 1 in such a library.
See also: https://octave.discourse.group/t/6026/75

Also don't strip `-lmoldname` from these flags because there is no reason to do that as far as I can tell.

This essentially reverts the upstream changes from: https://lists.gnu.org/r/autoconf-patches/2021-09/msg00000.html

### DIFF
--- a/autoconf2.72/0006-fortran-mingw-libs.patch
+++ b/autoconf2.72/0006-fortran-mingw-libs.patch
@@ -1,0 +1,14 @@
+Do not strip mingw libraries from list of Fortran standard linker flags.
+
+diff -urN autoconf-2.72/lib/autoconf/fortran.m4.orig autoconf-2.72/lib/autoconf/fortran.m4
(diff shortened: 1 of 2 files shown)


mmuetzel:
Just for info:
This change was merged upstream: https://git.savannah.gnu.org/cgit/autoconf.git/commit/?id=9ff9c567b1a7a7e66fa6523d4ceff142b86bddaa

PR #5102, opened 2024-12-26
ccache: unset CI env variable before running cmake

jeremyd2019:
ccache's cmake files will turn on a "dev mode" if CI is set in the environment, turning on -Werror among other things.  This was causing the build on i686 to fail due to a warning with fmt 11.  We don't really want ccache's dev mode, so unset it explicitly.

### DIFF
--- a/ccache/PKGBUILD
+++ b/ccache/PKGBUILD
@@ -3,7 +3,7 @@
 
 pkgname=ccache
 pkgver=4.10.2
-pkgrel=3
+pkgrel=4
 pkgdesc="A compiler cache (mingw-w64)"
 arch=('i686' 'x86_64')
 url="https://ccache.samba.org/"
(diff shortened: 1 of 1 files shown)


jeremyd2019:
> msys2-autobuild unsets most GHA-set environment variables (including CI), so this is not something that should affect the resulting binary

Oops, I may be mistaken on that: https://github.com/msys2-arm/msys2-autobuild/blob/35ff0b71b62cde6288cee94b82979cfcc017d16b/msys2_autobuild/build.py#L106 shows it just removing variables whose names start with `GITHUB_` or `RUNNER_`, not `CI`.

ognevny:
`-DCCACHE_DEV_MODE=OFF` doesn't help?
https://github.com/msys2/MINGW-packages/blob/63f6456b655cf57e159610d799a2cfb9828e0c42/mingw-w64-ccache/PKGBUILD#L68

jeremyd2019:
Yeah, that would probably work too, as would the `-DWARNINGS_AS_ERRORS=OFF` and `-Wno-error=stringop-overflow`

PR #5110, opened 2025-01-04
nano: Update to 8.3

lazka:


luau-project:
I tried to build nano 8.3 and got the same issue. Then, I updated gettext to 0.22.5 here on MSYS2 following Cygwin sources to see if it would solve the problem, and gettext 0.22.5 didn't solve too. Also tried to build nano 8.3 on Cygwin (which has gettext 0.22.5) and got the same behavior.

Do you think is it reasonable to open a PR posting my gettext 0.22.5 update or do you prefer to update it yourself?

lazka:
> Do you think is it reasonable to open a PR posting my gettext 0.22.5 update or do you prefer to update it yourself?

sure. I remember trying it once and also hitting some errors there, but I don't remember details.

PR #5114, opened 2025-01-07
guile: backport commits to fix 32-bit

jeremyd2019:
https://debbugs.gnu.org/cgi/bugreport.cgi?bug=71891 says the issue was fixed in aff9ac968840e9c86719fb613bd2ed3c39b9905c, but to avoid conflicts (which I don't know how to resolve, not knowing scheme), also pull in 5e6288c9304b60f1875a44808ee3858e3d6efc83
0dab58fc2a6ac6a8354439749d598f8c24f57ddd and
e45b70dcded37eb77e71ec30445b12040b6bb1b7.

I don't know if we want to do this or just wait for 3.0.11.

/cc @h-c-mueller


jeremyd2019:
git.savannah.gnu.org is being unreliable for me, I've added the patch files in this PR and just commented where they came from.

jeremyd2019:
I just noticed that it looks like Gentoo used the same 4 patches (though with one in a different order) to fix 32-bit - https://github.com/gentoo/gentoo/blob/a7505369249cc783af7abb00869da6f4311acd87/dev-scheme/guile/files/guile-3.0.10-backport-issue72913.patch

PR #5129, opened 2025-01-12
ninja: Fix bash completion installation directory

### LINKED ISSUE #5125: pacman: `ninja` bash completion copied to non-standard directory
https://github.com/msys2/MSYS2-packages/blob/93a84cd2dd81439a3e1320c211c16154eae44f6f/ninja/PKGBUILD#L38

The `/etc/bash-completion.d/` directory is non-standard, so it's not being sourced by bash. The correct path would be `/etc/bash_completion.d/`, but from what I could gather, this is a legacy path and should not be used in modern installations.

More info at the [official repo](https://github.com/scop/bash-completion/blob/main/README.md) and [here](https://unix.stackexchange.com/a/605051).

The ninja `PKGBUILD` at [AUR](https://gitlab.archlinux.org/archlinux/packaging/packages/ninja/-/blob/main/PKGBUILD?ref_type=heads#L47) can be used as an example.

```
install -m644 -D misc/bash-completion "$pkgdir/usr/share/bash-completion/completions/ninja"
```

I'm not very experienced with pacman packaging, but I can push a quick pull request with the correct path (the same used by the AUR package). But I'm not sure how to handle the case where the user is upgrading the `ninja` package. Should I include a `.install` that tests for the old path and delete it? Something like

```bash
# ninja.install

post_upgrade() {
    if [ -f etc/bash-completion.d/ninja ]; then
        rm etc/bash-completion.d/ninja
    fi
    # remove directory if empty
    if [ -z $(ls -A etc/bash-completion.d/) ]; then
        rmdir etc/bash-completion.d
    fi
}

post_remove() {
    post_upgrade
}
```

euamotubaina:
Fixes #5125

### DIFF
--- a/ninja/PKGBUILD
+++ b/ninja/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=ninja
 pkgname=("${_realname}" "${_realname}-vim" "${_realname}-emacs")
 pkgver=1.12.1
-pkgrel=1
+pkgrel=2
 _descr="Ninja is a small build system with a focus on speed"
 pkgdesc="${_descr}"
 arch=('i686' 'x86_64')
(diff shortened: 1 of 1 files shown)


lazka:
thanks

PR #5144, opened 2025-01-16
ruby: update to 3.4.1

jeremyd2019:
* Update to 3.4.1
* remove unnecessary manual extraction of sources, the default works just fine
* patch syslog gem to build on cygwin (it builds fine)
* patch win32/resolv to include headers when checking for GetNetworkParams function (it otherwise fails to find it on i686, presumably due to stdcall)
* limit build's make to -j2, due to errors apparently due to exhaustion of page file space


jeremyd2019:
Weirdnesses:
1. if a module tries to be disabled in its extconf.rb, it fails with:
```
../ruby-3.4.1/ext/extmk.rb:279:in 'Array#-': no implicit conversion of nil into Array (TypeError)
        from ../ruby-3.4.1/ext/extmk.rb:279:in 'Object#extmake'
        from ../ruby-3.4.1/ext/extmk.rb:659:in 'block in <main>'
        from ../ruby-3.4.1/ext/extmk.rb:653:in 'Array#each'
        from ../ruby-3.4.1/ext/extmk.rb:653:in '<main>'
```
2. syslog module was disabled on cygwin.  I patched to turn that on.
3. win32/resolv module was being disabled on i686 because it wasn't finding `GetNetworkParams` in `libiphlpapi.a`.  This was due to it not including the header, so it didn't know it was looking for a stdcall function.  I have a patch for that, but
4. apparently random `Segmentation fault (core dumped)` on i686

jeremyd2019:
on i686 it looks like it's segfauling in `system`

jeremyd2019:
Ugh, my patch to fix i686 broke x86_64 (and of course, details of what went wrong are not in the CI log, have to build locally to see what's going on).  More work to do here

jeremyd2019:
it builds locally for me now for both i686 and x86_64 (and passes tests), but the build is still unreliable.

yasuhirokimura:
I tried to build with `makepkg -cfLsC` but it freezed at check phase as following.

```
linking shared-library msys-ruby340.dll
linking rubyw.exe
linking ruby.exe
2025-01-19 09:18:22 +0900
Driver is ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-cygwin]
Target is ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-cygwin]

 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ..............................................................................
 ....................................................F.F.
```

My build environment is

* CPU: Intel Core i7 12700 (12 cores)
* Memory: 96GB
* Disk: 1TB NVMe SSD
* OS: Windows 11 2023H2

jeremyd2019:
Hmm, I've had a lot of issues with the build phase, but check has generally been OK (I've seen a couple of spurious failures in fork/signal tests under load).

yasuhirokimura:
I also tried `makepkg -cfLsC --nocheck` and it finished successfully.
Just FYI.

jeremyd2019:
I added a `-j1` on this PR in `build()` - that seemed to work around the issues I was seeing.  `-j1` on the check would probably work around the issues there too, but I really need to figure out what's going wrong

yasuhirokimura:
> I added a `-j1` on this PR in `build()` - that seemed to work around the issues I was seeing. `-j1` on the check would probably work around the issues there too, but I really need to figure out what's going wrong

I tried with HEAD of  your branch. So it means the freeze happened with `make -j1`.

jeremyd2019:
I meant adding `-j1` to the `make test` in `check()`.  But like I said, we really need to figure out why it fails with higher `-j` values.

yasuhirokimura:
> I meant adding `-j1` to the `make test` in `check()`. 

Got it.  After applying following change `check()` also finished successfully.

```
diff --git a/ruby/PKGBUILD b/ruby/PKGBUILD
index 60af4787..1874f3e8 100644
--- a/ruby/PKGBUILD
+++ b/ruby/PKGBUILD
@@ -98,7 +98,7 @@ build() {

 check() {
   cd ${srcdir}/build-${CHOST}
-  make test
+  make -j1 test
 }

 package_ruby() {
```

> But like I said, we really need to figure out why it fails with higher `-j` values.

Agreed. I don't know the detail of `make test` at all but it seems like deadlock happens between running tests.

jeremyd2019:
I saw the same fixup_mmaps_after_fork issues with ruby 3.3.7, and with upstream cygwin.  https://cygwin.com/pipermail/cygwin/2025-January/257155.html suggests it's due to pagefile size?

jeremyd2019:
I figure -j2 is a reasonable compromise to avoid pagefile exhaustion while still getting some parallelism in the build.  Once (assuming) this is merged, I figure I should try to get my patches upstream, and report a bug for the extmk.rb error at least.

jeremyd2019:
-j2 still fails an uncomfortable amount of the time on i686, but as that's unsupported, I can live with having to retry a few times there.

jeremyd2019:
> I figure I should try to get my patches upstream, and report a bug for the extmk.rb error at least.


https://github.com/ruby/syslog/pull/16, https://github.com/ruby/ruby/pull/12645 (and https://github.com/ruby/resolv/pull/81, not sure where they want PRs against that code opened), and https://bugs.ruby-lang.org/issues/21092 for the underlying extmk.rb issue

PR #5147, opened 2025-01-18
msys2-runtime: test Takashi's patches

jeremyd2019:
As these are for testing, apply them separately and leave the msys2-runtime PR as a draft for now
(https://github.com/msys2/msys2-runtime/pull/253).


jeremyd2019:
My idea is that we merge this and give it some exercise in CI to see if we see any more hangs.  I think Git for Windows' tests sounded like a better indicator of whether the hangs are fixed or not, though.  To that end, I opened https://github.com/git-for-windows/msys2-runtime/pull/81 as well.

jeremyd2019:
These patches are still a moving target 🤷

jeremyd2019:
These are now in cygwin/master and cygwin-3_5-branch (well, except for the revert, which the original patch was never applied to 3.5).  There is still one more hang-related patch pending review though...

jeremyd2019:
All the patches are now in cygwin/master and cygwin-3_5-branch.

Issue #4712, opened 2024-06-16
[RFC] Dropping old msys/i686 packages from repo.msys2.org [edit:] sync instead


lazka:
Context: Motivated by https://github.com/jeremyd2019/msys2-build32/issues/2#issuecomment-2170938529

We currently still have 32bit msys packages hosted on repo.msys2.rog: https://repo.msys2.org/msys/i686/

We dropped support for them 4 years ago, but updated them once more about 3 years ago based on jeremy's builds, to keep them somewhat usable.

Since @jeremyd2019 is still putting work into them and building all packages we could just remove our remaining packages and point to his repo instead:

* Check the logs if there are any users of those packages
* (optional) rename the dir to see if anyone complains
* Drop msys/i686 and the 32bit sfx/tarball
* Point users to https://github.com/jeremyd2019/msys2-installer/releases if they need 32bit still

jeremyd2019:
I would have to adjust some scripts, currently I still have the i686 `[msys]` section in pacman.conf, but my new `[build32]` section is first, but that repo now has everything that the msys repo has so it'd just cause errors syncing.

My 'main' 32-bit tag of jeremyd2019/setup-msys2 uses the original 32-bit sfx and repos to install (and then I modify pacman.conf later on), so there will be some accesses from that.  I also have another branch that uses my installer and build32 repo directly, and that would probably be a better choice to use moving forward - but I never switched on jeremyd2019/msys2-build32 because I want to exercise the update path.

lazka:
> I would have to adjust some scripts, currently I still have the i686 `[msys]` section in pacman.conf, but my new `[build32]` section is first, but that repo now has everything that the msys repo has so it'd just cause errors syncing.

I could leave an empty db there, if it helps.

lazka:
> Check the logs if there are any users of those packages

Here are the 32bit logs for last week (we had an outage, so it's not complete)

```
--------  ----------------------------------------------------------------------
Start     2024-06-12T12:25:36Z
End       2024-06-17T05:22:06Z
Requests  2480
Clients   122 (Clients are grouped by IP+WinVer+Arch, which is far from perfect)
--------  ----------------------------------------------------------------------

     Repo    Type    % Requests    Requests
---------  ------  ------------  ----------
msys/i686     pkg       100.00%        1936

     Repo    Type    % Requests    Requests
---------  ------  ------------  ----------
msys/i686      db       100.00%         544

[...]

PR #5177, opened 2025-02-01
perl: build for cygwin

lazka:
The goal is for perl to report that it was built for cygwin ($^O gives "cygwin" for example) and to adjust all the code depending on that and remove all the patches for handing the "msys" case. Everything else should work as before.

* In case cygwin check was extended with msys, remove msys case
* In case msys check was added, replace with cygwin check
* In case msys specific checks/code was added, merge that into the cygwin paths

We still check/set the MSYS env var instead of the CYGWIN env var, since msys2-runtime doesn't handle both yet, so keep that as is for now.

Part of #3012

edit: confirmed with the test suite that there are no changes in the test results


lazka:
@dscho since I've always copied the perl work from git4win, any thoughts/feedback on this? It should make updating perl easier in the future.

perl/0006-perl-5.36.0-msys2.patch
```diff
@@ -2241,23 +415,54 @@ index 4715e9d..5da0af0 100644
...
+ fi
+ so='dll'
+-# - eliminate -lc, implied by gcc and a symlink to libcygwin.a
++# - eliminate -lc, implied by gcc and a symlink to libmsys-2.0.a
+ libswanted=`echo " $libswanted " | sed -e 's/ c / /g'`
+-# - eliminate -lm, symlink to libcygwin.a
++# - eliminate -lm, symlink to libmsys-2.0.a
+ libswanted=`echo " $libswanted " | sed -e 's/ m / /g'`
+-# - eliminate -lutil, symbols are all in libcygwin.a
++# - eliminate -lutil, symbols are all in libmsys-2.0.a
```
jeremyd2019, line 435: Is it really worth patching comments?

perl/0006-perl-5.36.0-msys2.patch
```diff
@@ -2241,23 +415,54 @@ index 4715e9d..5da0af0 100644
...
++libc='/usr/lib/libmsys-2.0.a'
+ loclibpth=' '
+ glibpth=' '
+ plibpth=' '
+@@ -38,9 +38,9 @@
+ test -z "$optimize" && optimize='-O3'
+ man3ext='3pm'
+ test -z "$use64bitint" && use64bitint='define'
+-test -z "$useithreads" && useithreads='define'
++test -z "$usethreads" && usethreads='define'
```
jeremyd2019, line 449: ?

perl/0006-perl-5.36.0-msys2.patch
```diff
@@ -2241,23 +415,54 @@ index 4715e9d..5da0af0 100644
...
+ #lddlflags="$lddlflags -s"
+-
+-# Seems that exporting _Thread_local doesn't work on cygwin. This 6 year old
+-# gcc bug suggests that maybe the problem really is binutils, but either way
+-# it still doesn't work, despite our probes looking good:
+-# https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64697
+-d_thread_local=undef
+-
+-# Broken: https://sourceware.org/pipermail/cygwin/2022-August/252043.html */
+-d_newlocale=undef
```
jeremyd2019, line 468: why remove these?

perl/0006-perl-5.36.0-msys2.patch
```diff
@@ -1364,124 +211,22 @@ index 2bdf799..3637dc0 100644
        *Term::ReadLine::Stub::devtty = sub { $donotexist };
      is( Term::ReadLine->devtty(), $donotexist, "devtty mocked" );
  
-+    skip('console is CONIN$/CONOUT$ in MSYS2, not /dev/tty', 3) if $^O eq 'msys';
++    skip('console is CONIN$/CONOUT$ in MSYS2, not /dev/tty', 3) if $^O eq 'cygwin';
```
jeremyd2019, line 214: I wonder if this works now.

perl/0006-perl-5.36.0-msys2.patch
lazka: Fair point, but I just copied over the `hints/msys.sh` changes to `hints/cygwin.sh` (and adjusting the arch strings) to avoid any other side effects of this PR.

perl/0006-perl-5.36.0-msys2.patch
lazka: see above

perl/0006-perl-5.36.0-msys2.patch
lazka: see above

perl/0006-perl-5.36.0-msys2.patch
jeremyd2019: so cygwin.sh probably diverged since msys.sh was created

lazka: yeah, likely: https://github.com/Perl/perl5/blame/blead/hints/cygwin.sh

perl/0006-perl-5.36.0-msys2.patch
lazka: indeed, it works now

dscho:
> @dscho since I've always copied the perl work from git4win, any thoughts/feedback on this?
> 
> It should make updating perl easier in the future.

That's a good goal!

As for Git for Windows, I recognize that for i686 builds, we're stuck with requiring the `msys2` patches, though, and therefore the updating will be a bit harder in the future.

Having said that, it's a good idea for MSYS2 to go ahead with this.

In Git for Windows, I'll have to adjust a couple of lines:

- in `git cvsexportcommit`: https://github.com/git-for-windows/git/blob/v2.47.1.windows.2/git-cvsexportcommit.perl#L40
- in `git send-email`: https://github.com/git-for-windows/git/blob/v2.47.1.windows.2/git-send-email.perl#L1486
- in `git svn`: https://github.com/git-for-windows/git/blob/v2.47.1.windows.2/git-svn.perl#L68

Note how inconsistently the affected Git code is: Even the underlying theme is to handle DOS/Unix paths issues, the handling is ad-hoc and not standardized in `Git.pm`. Unfortunately, this kind of inconsistency is pretty consistent in Git's source code, and a consequence of [Conway's Law](https://en.wikipedia.org/wiki/Conway's_law).

tl;dr no objections from my side!

jeremyd2019:
> As for Git for Windows, I recognize that for i686 builds, we're stuck with requiring the `msys2` patches, though, and therefore the updating will be a bit harder in the future.

Why would i686 builds still require the msys2 patches?  I've been building i686 packages, including those that remove msys2 patches, without issues

lazka:
> Why would i686 builds still require the msys2 patches? I've been building i686 packages, including those that remove msys2 patches, without issues

I'd guess because the MSYSTEM=CYGWIN logic is not in the current i686 repo. I need to look into #4712 ...

dscho:
> > Why would i686 builds still require the msys2 patches? I've been building i686 packages, including those that remove msys2 patches, without issues
> 
> hmmm, I'd guess because the MSYSTEM=CYGWIN logic is not in the current i686 repo.

It's a bit worse than that: [Git for Windows'](https://github.com/git-for-windows/MSYS2-packages/commits/HEAD/msys2-runtime-3.3) and [MSYS2's](https://github.com/msys2/MSYS2-packages/commits/HEAD/msys2-runtime-3.3) commit histories of the `msys2-runtime-3.3/` MSYS2-package are completely divergent. I do not want to put any more work into it, really: As documented in https://gitforwindows.org/32-bit Git for Windows is near/at its end of limited 32-bit support, and will soon drop building anything except the MinGit packages (which notably do not include Perl). I guess Git for Windows v2.48.1 will be the last Git for Windows release with i686 installers...

PR #5188, opened 2025-02-03
msys2-runtime-3.3: build as msys2-runtime on i686

lazka:
So there is an automated upgrade path and users don't have to manually switch to msys2-runtime-3.3.

### DIFF
--- a/msys2-runtime-3.3/PKGBUILD
+++ b/msys2-runtime-3.3/PKGBUILD
@@ -2,9 +2,13 @@
 # Contributor: Ray Donnelly <mingw.android@gmail.com>
 
 pkgbase=msys2-runtime-3.3
-pkgname=('msys2-runtime-3.3' 'msys2-runtime-3.3-devel')
+if [ "$CARCH" == "i686" ]; then
+  pkgname=('msys2-runtime' 'msys2-runtime-devel')
+else
+  pkgname=('msys2-runtime-3.3' 'msys2-runtime-3.3-devel')
+fi
 pkgver=3.3.6
-pkgrel=13
+pkgrel=14
 pkgdesc="Cygwin POSIX emulation engine"
 arch=('i686' 'x86_64')
 url="https://www.cygwin.com/"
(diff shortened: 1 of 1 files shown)


lazka:
@jeremyd2019 does this make sense/help?

jeremyd2019:
I don't know.  Would it be possible to do a conflicts/replaces thing, either on msys2-runtime-3.3, or on this new msys2-runtime against msys2-runtime-3.3, or would that cause issues with pacman?

lazka:
Ah, sure, like this?

jeremyd2019:
I'll try to test this today

msys2-runtime-3.3/PKGBUILD
```diff
@@ -300,6 +304,15 @@ package_msys2-runtime-3.3() {
   cp -rf "${srcdir}"/dest/usr/share "${pkgdir}"/usr/
 }
 
+package_msys2-runtime() {
+  pkgdesc="Posix emulation engine for Windows"
+  options=('!strip')
+  conflicts=('catgets' 'libcatgets' 'msys2-runtime-3.3')
+  replaces=('catgets' 'libcatgets' 'msys2-runtime-3.3')
+
+  package_msys2-runtime-3.3
```
jeremyd2019, line 313: As I kind of expected, this didn't work:
```console
$ pacman -Qip msys2-runtime-3.3.6-14-i686.pkg.tar.zst
Name            : msys2-runtime
Version         : 3.3.6-14
Description     : Posix emulation engine for Windows
Architecture    : i686
URL             : https://www.cygwin.com/
Licenses        : GPL
Groups          : None
Provides        : msys2-runtime=3.3.6
Depends On      : None
Optional Deps   : None
Conflicts With  : catgets  libcatgets  msys2-runtime  msys2-runtime-3.5
Replaces        : catgets  libcatgets
Compressed Size : 2.57 MiB
Installed Size  : 9.79 MiB
Packager        : Unknown Packager
Build Date      : Tue, Feb 4, 2025 1:14:50 PM
Install Script  : No
Validated By    : None
Signatures      : None
```

In some cases, makepkg effectively greps the function for the variable definitions, but in others it relies on their values after running the function.  In the former case, it would see the definitions here, but in the latter the values set in `package_msys2-runtime-3.3` would have overwritten them.  It seems the former is done for making the `.SRCINFO` and the latter is done for making the binary packages themselves.

jeremyd2019:
This seems to work

msys2-runtime-3.3/PKGBUILD
lazka: thanks, I need to keep this in mind :/

lazka:
thanks for testing! (I should have done so myself..)

PR #5189, opened 2025-02-03
gnu-netcat: build for cygwin

lazka:


ognevny:
> this also seems to fix dynamic linking to gettext

installed size is now only 100 KiB! is this ok?

lazka:
yes, previously is statically linked libintl

PR #5190, opened 2025-02-03
bash: build for cygwin

lazka:
* the only difference should be the target triplet printed by "--version"
* configure is the same except WOE32 is set to yes, but from what I see this isn't used
* remove cross compile cache, doesn't look used


ognevny:
bash-devel seems empty after an update

lazka:
thanks, that's just a side effect of bash install being skippe in CI I think

lazka:
Another difference:

```console
$ echo $OSTYPE # before
msys
$ echo $OSTYPE # after
cygwin
```

PR #5194, opened 2025-02-05
util-linux: update to 2.40.2

ognevny:
sync patches and some configure opts with cygwin.

@omikrun is this enough for https://github.com/msys2/MINGW-packages/pull/23280? ~I didn't manage to find cygwin patches for more recent version~


omikrun:
Yes, `rename` work properly.

```text
$ ls
libressl-openssl.1.gz
```

Thanks for your update. :)

omikrun:
Additionally, could you mind update it to 2.40.2? Here is the source code of it.
https://sourceware.org/ftp/cygwin/x86_64/release/util-linux/util-linux-2.40.2-2-src.tar.xz

jeremyd2019:
It looks like you lost col.exe and taskset.exe.  I know col.exe was just mentioned on the Cygwin mailing list: https://cygwin.com/pipermail/cygwin/2025-February/257225.html

ognevny:
> It looks like you lost col.exe and taskset.exe

taskset.exe is lost because one of the options lost while copy-pasting configure options from cygwin

jeremyd2019:
configure was saying it couldn't enable taskset due to not finding `sched_setaffinity` ? (this was before your change to switch `--enable-schedutils` back on)

ognevny:
> configure was saying it couldn't enable taskset due to not finding `sched_setaffinity` ? (this was before your change to switch `--enable-schedutils` back on)

--enable-schedutils pulls ionice with it, which seems unsupported for cygwin. I put link above for a configure.ac line with schedutils option

jeremyd2019:
I think you're still going to need something like the cygwin-taskset patch that was deleted for taskset to be enabled.

ognevny:
cygwin doesn't provide a patch for taskset...

jeremyd2019:
```patch
--- util-linux-2.40.2/configure.ac.orig      2025-02-06 11:29:31.296507100 -0800
+++ util-linux-2.40.2/configure.ac   2025-02-06 11:33:08.141674200 -0800
@@ -2535,9 +2535,8 @@
 UL_ENABLE_ALIAS([taskset], [schedutils])
 UL_BUILD_INIT([taskset])
 UL_REQUIRES_HAVE([taskset], [cpu_set_t], [cpu_set_t type])
-UL_REQUIRES_SYSCALL_CHECK([taskset],
-       [UL_CHECK_SYSCALL([sched_getaffinity])],
-       [sched_getaffinity])
+AC_CHECK_FUNCS([sched_getaffinity sched_setaffinity], [have_schedaffinity=yes],[have_schedaffinity=no])
+UL_REQUIRES_HAVE([taskset], [schedaffinity], [sched affinity functions])
 AM_CONDITIONAL([BUILD_TASKSET], [test "x$build_taskset" = xyes])
```

> cygwin doesn't provide a patch for taskset...

:shrug: maybe they need to?

ognevny:
feel free to push it then. I'll go sleep soon 😴

jeremyd2019:
ok.  oddly, taskset.exe is present in their package.  wonder how they did it?  https://www.cygwin.com/cgi-bin2/package-cat.cgi?file=x86_64%2Futil-linux%2Futil-linux-2.40.2-2&grep=taskset

jeremyd2019:
Their cygport expects sys/syscall.h to exist, and it doesn't for us.  it doesn't seem to be in any package... https://www.cygwin.com/cgi-bin2/package-grep.cgi?grep=syscall.h

jeremyd2019:
ok, looks like all the exes are present that are in the Cygwin util-linux (plus uuidd.exe, which is in its own package on Cygwin)

jeremyd2019:
> Their cygport expects sys/syscall.h to exist, and it doesn't for us. it doesn't seem to be in any package... https://www.cygwin.com/cgi-bin2/package-grep.cgi?grep=syscall.h

https://cygwin.com/pipermail/cygwin/2025-February/257250.html

jeremyd2019:
According to https://inbox.sourceware.org/cygwin/e817d642-da7d-471d-83b3-0a5294fb4f4c@maxrnd.com/, it is generated by a small shell script:
```sh
#!/bin/sh
#
#       gen-syscall-h.sh
#       Generates #defines for a plausible Cygwin <sys/syscall.h> on stdout
#       2023/12/30 Original version by Mark A. Geisert <mark@maxrnd.com>
#
#       This currently works by creating a list of all symbols exported
#       by libcygwin.a and subtracting out the symbols also exported
#       by libm.a.  After that, a symbol is kept only if it starts with
#       a lowercase letter and does not start with "cygwin_" or "main".
#
TMPF=/tmp/.$$.
trap 'rm -f $TMPF' EXIT
nm /usr/lib/libm.a | awk '/ T [a-z]/    {print $3}' | sort > $TMPF
nm /usr/lib/libcygwin.a | awk '/ T [a-z]/       {print $3}' | \
         grep -vE '^cygwin_|^main' | sort | \
         diff --suppress-common-lines -y - $TMPF | \
         awk '/^[a-z]/   {printf "#define SYS_%s %d\n", $1, ++RECNO}'
exit $?
```

Not sure we want to "Direct stdout to /usr/include/sys/syscall.h if you dare ;-).".  I'm thinking we merge this PR as-is, and punt on this.

ognevny:
> Not sure we want to "Direct stdout to /usr/include/sys/syscall.h if you dare ;-).".  I'm thinking we merge this PR as-is, and punt on this.

I agree :)

jeremyd2019:
OK, nobody seems to have any objections

PR #5201, opened 2025-02-07
mintty: default to Term=xterm-256color

### LINKED ISSUE #4973: [question] Should not we enable xterm-256color on mintty by default?
## Description

> [!NOTE]
> 
> Since I am relatively new to terminal-related topics, I opened this issue to read opinions from people with more experience on this subject, because I don't know very well the possible impacts of the proposed changes.

**My question is**: Given that [Mintty](https://mintty.github.io/) is the default terminal emulator for MSYS2 environments and has support even for 24-bit colors, should not we set ```Term=xterm-256color``` on ```/etc/minttyrc```, which seems to be described on this file [https://github.com/msys2/MSYS2-packages/blob/master/mintty/minttyrc](https://github.com/msys2/MSYS2-packages/blob/master/mintty/minttyrc)?

> [!TIP]
> 
> I am aware that each user can make such config changes by placing ```Term=xterm-256color``` on ```~/.minttyrc```.

My question is grounded on a few facts:

* MSYS2 supports Windows 10+ (maybe 8.1?). So, I don't see much of a problem on setting ```xterm-256color``` on such recent OSes;

* Ubuntu 24.04 (gnome terminal) uses ```xterm-256color``` by default. Probably, most recent distros too.

* If you open a ```cmd``` prompt on MSYS2 install folder (*assuming an unmodified standard install at ```C:\msys64```*), and echo the TERM environment variable on bash through mintty, you get ```xterm``` as a response. On the other hand, if you echo TERM through bash directly, you get ```xterm-256color```. So, it seems like that the default TERM used by bash is  ```xterm-256color```, and mintty is just overriding it:

Screenshot from 2024-10-25 11-00-57

## Drawbacks

Quoting from [mintty wiki](https://mintty.github.io/mintty.1.html#Terminal):

> The xterm−256color setting enables 256-color mode in some applications, but may not be recognised at all by others, which is why plain xterm is the default.

However, I'm not sure how well it applies for the MSYS2 environments.

## Benefits

Bash script used below:

```bash
#!/usr/bin/bash

for c in {0..255}; do
    tput setaf ${c};
    echo -n "$c ";
done;

tput sgr0;
echo;
```

* In the current setup with ```xterm```

Screenshot from 2024-10-25 11-12-28

* When ```xterm-256color``` is enabled

Screenshot from 2024-10-25 11-13-41

lazka:
instead of plain xterm. xterm-256color is the default on most modern systems, so give it a try.

Fixes #4973

### DIFF
--- a/mintty/PKGBUILD
+++ b/mintty/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=mintty
 pkgver=3.7.7
-pkgrel=1
+pkgrel=2
 epoch=1
 pkgdesc="Terminal emulator with native Windows look and feel"
 arch=('i686' 'x86_64')
@@ -18,7 +18,7 @@ backup=('etc/minttyrc')
 source=(${pkgname}-${pkgver}.tar.gz::https://github.com/mintty/mintty/archive/${pkgver}.tar.gz
         "minttyrc")
 sha256sums=('6f1a03e3fdab51927a1eb7415a1478f694aeb57d37edcee8eca50ed3a7e044a0'
-            'aabad49568755a05e76b2581eb6061e27c3ab099aa0f4eb536dfdbcac99cac1f')
+            'ce43a9b0c8eb01a695d0543ceb56c9281708a02aaf92021b72d0f0b69a7feeb9')
 
 build() {
   cd "${srcdir}/${pkgname}-${pkgver}"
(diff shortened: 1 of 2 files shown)


luau-project:
I've been using ```xterm-256color``` as my setting since I opened #4973 , and I didn't face any issue so far.

PR #5221, opened 2025-02-20
filesystem: Remove /etc/bash.bash_logout

lhmouse:
By adding `Hold=always` in ~/.minttyrc, it is possible to keep MinTTY open after all processes in it have terminated, which is useful for examination of outputs. However this script used to clear the screen, leaving a blank window, and causing confusion.

Another reason for removing this script is that it's not found on fresh installation of Debian or Ubuntu.

Reference: https://superuser.com/questions/1262655/mintty-how-to-keep-my-output-visible-after-processes-terminate/1263001#1263001


lazka:
I can't confirm the Ubuntu/Debian statement:

Ubuntu 24.04:

```console
user@desktop:~$ cat /etc/skel/.bash_logout
# ~/.bash_logout: executed by bash(1) when login shell exits.

# when leaving the console clear the screen to increase privacy

if [ "$SHLVL" = 1 ]; then
    [ -x /usr/bin/clear_console ] && /usr/bin/clear_console -q
fi
```

Debian: https://sources.debian.org/src/bash/5.2.37-1.1/debian/skel.bash_logout/

lazka:
But I agree that it's more confusing than useful.

If someone complains we can add it back as `/etc/skel/.bash_logout` like Debian/Ubuntu, so users can delete it themselves.

lhmouse:
Yes on Debian there's a per-user .bash_logout, unlike MSYS2 where there's a global bash.bash_logout and the per-user one is initially empty.

By default mintty exits if the shell process in it terminates, so the change should be invisible for default setups.

cbrt64:
I know I'm a little late here, but in general shouldn't config files in `backup` be kept there, at least for a while, to keep pacman from removing custom user scripts at package upgrade?

lazka:
> I know I'm a little late here, but in general shouldn't config files in `backup` be kept there, at least for a while, to keep pacman from removing custom user scripts at package upgrade?

Good question. I would have assumed it would backup the old file when removing the old version, but I haven't tested it.

cbrt64:
> I would have assumed it would backup the old file when removing the old version

Not having time to read the docs or the source, I went with a test PKGBUILD, it works just like you suspected. Sorry for the noise.

PR #5224, opened 2025-02-21
nano: follow git-for-windows and paint interface elements

luau-project:
## Description

As the title says: follow git-for-windows and paint interface elements of nano

[https://github.com/git-for-windows/git-sdk-64/blob/869fd0088bed870075a3a8e109367c0920625b57/etc/nanorc#L219-L233](https://github.com/git-for-windows/git-sdk-64/blob/869fd0088bed870075a3a8e109367c0920625b57/etc/nanorc#L219-L233)

> [!NOTE]
> 
> I did not use the same colors as git-for-windows to differentiate the MSYS2 shells. The ```titlecolor``` was set as magenta (resembles MSYS2 icon color), and orange / yellow on other parts to resemble the recommended environment (UCRT64) icon color.

## Screenshots

Once this PR is merged, a file opened on nano through any MSYS2 shell would look like the following image

Screenshot from 2025-02-21 11-58-12

## Wishes

In my opinion, MSYS2 should go even further enabling the default syntax highlighting files by uncommenting the line 252

### DIFF
--- a/nano/0002-paint-interface-elements.patch
+++ b/nano/0002-paint-interface-elements.patch
@@ -1,0 +1,32 @@
+diff -Naur nano-8.3-orig/doc/sample.nanorc.in nano-8.3/doc/sample.nanorc.in
+--- nano-8.3-orig/doc/sample.nanorc.in	2024-12-20 07:38:10.000000000 -0300
++++ nano-8.3/doc/sample.nanorc.in	2025-02-21 11:33:55.337769515 -0300
+@@ -219,17 +219,17 @@
+ 
+ ## Paint the interface elements of nano.  These are examples; there are
+ ## no colors by default, except for errorcolor and spotlightcolor.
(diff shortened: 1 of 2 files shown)


Biswa96:
Software distributions should not modify user configured options except when it causes issue with the basic use case.

luau-project:
> Software distributions should not modify user configured options except when it causes issue with the basic use case.

The proposed changes do not modify any user configured options. As far as I know, user configured options for nano goes on ```~/.nanorc```. Any option set on  ```~/.nanorc``` will take precedence over the default settings on ```/etc/nanorc```.

lazka:
Puh, with it being off by default in all distros I can imagine people not liking it/not being used to it, or it being distracting in combination with syntax highlighting.

Note that you changed the default colors, so they don't match git4win.

PR #5233, opened 2025-02-27
genromfs: close() fds after use

jhol:
The current version of genromfs has a bug whereby it leaks fd handles by not closing them after use. This results in the tool failing with large sets of files (approximately 9,000) displaying an error: `Too many open files`.

Linux users are typically less affected by this bug because the operating system in most cases allows a far greater number of simultaneously open file handles than is allowed on Windows.

A patch to correct the issue has been provided by @NomanKhan777 [here](https://github.com/chexum/genromfs/pull/5)

However, the maintainer does not appear to be very responsive. The most recent commit in the repository was made in 2011. Unless and until this patch is accepted upstream, or a new maintainer is found for the project, this commit adds the patch to MSYS2-packages to correct the issue.

### DIFF
--- a/genromfs/0001-genromfs.c-alignnode-close-fd-after-read.patch
+++ b/genromfs/0001-genromfs.c-alignnode-close-fd-after-read.patch
@@ -1,0 +1,35 @@
+From 5c995f8ebf22a790374bf1964c1cf135c132842d Mon Sep 17 00:00:00 2001
+From: nkhan <nkhan@nvidia.com>
+Date: Tue, 11 Feb 2025 15:47:02 -0800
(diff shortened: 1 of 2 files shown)


jhol:
Looking again at this, I think the issue is that the recipe packages the v0.5.7 tag - which, very confusingly, was never officially released. The most recent *true* release is v0.5.2 which does not appear to have this bug.

@lazka If we can confirm that the v0.5.2 does not have the bug, is it possible to roll out a downgrade to the package?

jhol:
This is now superseded by this PR: https://github.com/msys2/MSYS2-packages/pull/5234

PR #5240, opened 2025-03-05
enscript: add patch, update url

jeremyd2019:
Build as cygwin.

Based on a review of Cygwin's source package.

Note that there is a .sig file, but the signing key has been revoked
(possibly because it is a dsa1024 key?)

### DIFF
--- a/enscript/PKGBUILD
+++ b/enscript/PKGBUILD
@@ -3,25 +3,27 @@
 
 pkgname=enscript
 pkgver=1.6.6
-pkgrel=1
+pkgrel=2
 pkgdesc="Convert ASCII files to PostScript suitable for printing"
 arch=('i686' 'x86_64')
-url="https://git.savannah.gnu.org/cgit/enscript.git"
+url="https://www.gnu.org/software/enscript/"
 msys2_references=(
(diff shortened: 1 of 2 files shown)


jeremyd2019:
Cygwin also has a patch to the tests that might be relevant for us, but since we don't generally run the check phase, probably not worth the effort:
```patch
--- origsrc/enscript-1.6.4/src/tests/defs       2003-03-05 08:26:33.000000000 +0100
+++ src/enscript-1.6.4/src/tests/defs   2013-04-28 19:59:44.237652400 +0200
@@ -10,7 +10,7 @@ export ENSCRIPT_LIBRARY
 output=testout.ps

 asciifiles="$srcdir/defs $srcdir/ascii.test $srcdir/../psgen.c"
-binaryfiles="../enscript ../mkafmmap"
+binaryfiles="../enscript.exe ../mkafmmap.exe"

 enscript="../enscript"
```

jeremyd2019:
```
  ==> Verifying source file signatures with gpg...
      enscript-1.6.6.tar.gz ... FAILED (public key 38734D7787901452 has been revoked)
```
OK then, no pgp verification.

PR #5244, opened 2025-03-10
Upgrade GNU Privacy Guard to v2.5.5

dscho:
This makes the big jump from v2.4.* to v2.5.*.

This requires an upgrade of `libassuan`, too. Which might cause some problems, as the current DLL is called `msys-assuan-0.dll` whereas the new DLL is called `msys-assuan-9.dll` (which means that e.g. `dirmngr.exe` simply won't start anymore if `libassuan` is upgraded in isolation...).

### DIFF
--- a/gnupg/0001-gnupg-2.2.8-msys2.patch
+++ b/gnupg/0001-gnupg-2.2.8-msys2.patch
@@ -1,7 +1,7 @@
 diff -ru gnupg-2.2.8-orig/g10/gpg.c gnupg-2.2.8/g10/gpg.c
 --- gnupg-2.2.8-orig/g10/gpg.c	2018-06-06 11:59:06.000000000 +0200
 +++ gnupg-2.2.8/g10/gpg.c	2018-06-21 01:00:49.167416100 +0200
-@@ -1508,7 +1508,7 @@
+@@ -1676,7 +1676,7 @@ check_permissions (const char *path, int item)
  	{
  	  if(statbuf.st_uid==getuid())
  	    {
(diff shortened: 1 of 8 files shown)


dscho:
@lazka FYI here I proofed out that the development version of `gnupg` can be built relatively easily, in preparation for v2.6.0 (which according to https://gnupg.org/download/index.html#end-of-life was born last year but there [still is no v2.6.0 tag](https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg.git;a=tags)).

Since I just realized that all of this effort was not really needed because [v2.5.x are the _development_ versions](https://gnupg.org/download/index.html#sec-1-1) (and are not recommended for end users), I will close this PR, but want you to know about it for future reference, once v2.6.0 comes out.

lazka:
thanks.

yeah, re libassuan, I was worried that it might break updates, which is why I haven't updated yet, but not actually sure if it is a problem assuming users don't do partial upgrades.

dscho:
> Worst case we'll have to ship both DLLs for a while to make autobuild work.

@lazka right, that's what https://github.com/msys2/MSYS2-packages/pull/5244/commits/8cf4dbd7ec8d49983e47d07666a9312e82d7b92a is about, in an easy-to-revert commit.

I guess I could back out the `gnupg` upgrade and turn it into a rebuild (i.e. `2.4.7-2`) instead, then reopen this PR. Would you like that? Or would you rather wait until GnuPG v2.6.0 is available before doing that?

dscho:
> Or would you rather wait until GnuPG v2.6.0 is available before doing that?

I guess a natural progression would be to upgrade `libassuan` _now_, with the ugly work-around, and then use the `gnupg-2.6.0` PR (when GnuPG v2.6.0 is available) as a fine opportunity to drop the work-around.

lazka:
I'll have a look.
 
Note that I usually wait until Arch updates gnupg since there have been compat issues in the past.

PR #5254, opened 2025-03-13
fmt: Update to 11.1.4

Gork336:


fmt/PKGBUILD
```diff
@@ -1,7 +1,8 @@
-# Maintainer: Christoph Reiter <reiter.christoph@gmail.com>
+# Maintainer : Gork             <bertsmithzp@gmail.com>
```
lazka, line 1: What's this about?

Gork336: Hi @lazka 
I’m trying to update the `fmt` package, but I’m still new to this process and not entirely sure about the correct workflow. I added myself as a maintainer (`Gork`) alongside the original maintainer (`Christoph Reiter`), but I’m not sure if this is the right approach. Should I instead:
1. Keep the original maintainer and add myself as a contributor, or
2. Not add myself at all if I’m just making a one-time update?

Please let me know the correct way to handle this. I want to make sure I’m following the proper conventions.

Thanks for your guidance!

lazka: I was just wondering if you had any bigger plans for fmt, since you removed the original maintainer.

We don't have any rules really around package maintainers/contributors, and the git history is more helpful in figuring out who works on what.

Gork336: Hi @lazka ,Thanks for the  guidance.  I’ll go ahead and update the PKGBUILD to:

- Keep `Christoph Reiter` as the main Maintainer.
- Add myself (`Gork`) as a one-time Contributor for this update.

If that's okay, I'll change it and submit it again

Also, I’d love to do more for the fmt package and the `MSYS2-packages` repositories.

PR #5260, opened 2025-03-15
libblake3: use windows asm under cygwin; disable avx512

lazka:
the first one made it crash when using asm, the later avoids it failing to link at build time

This fixes ccache segfaulting in some cases.


ognevny:
filled a patch upstream

PR #5263, opened 2025-03-17
libcbor: update to v0.12.0

dscho:


lazka:
Fixed with #5299. The error was due to the source gaining a "build" file in this release which conflicts with the cmake build dir afaik.

PR #5264, opened 2025-03-18
udis86: new package

jeremyd2019:
There was a suggestion at one point that udis86 could be used for walking machine code in cygwin/msys2-runtime when looking for fast cwd.  This builds the library for prototype purposes.


jeremyd2019:
See https://cygwin.com/pipermail/cygwin-patches/2025q1/013506.html

jeremyd2019:
https://cygwin.com/cgit/cygwin-packages/udis86

PR #5268, opened 2025-03-19
zsh: fix performance issues related to PATH

samhocevar:
When `$PATH` contains `/` and command `foo` is run, zsh will try to execute `//foo` instead of `/foo`, which will attempt to contact a network share on server `foo`, causing DNS and CIFS requests that will greatly slow down the shell. And if `/foo` actually exists, it will not be found.

As a workaround, we do not insert the additional `'/'` if the path element already ends with a slash.

Another workaround for the users would be to put `/.` instead of `/` in `$PATH`.

### DIFF
--- a/zsh/PKGBUILD
+++ b/zsh/PKGBUILD
@@ -23,6 +23,7 @@ source=("https://www.zsh.org/pub/zsh-${pkgver}.tar.xz"{,.asc}
         msysize.patch
         zsh-5.0.6-1.patch
         msys-symlink-hack.patch
+        path-performance-issues.patch
         0004-pcre2.patch)
 sha256sums=('9b8d1ecedd5b5e81fbf1918e876752a7dd948e05c1a0dba10ab863842d45acd5'
             'SKIP'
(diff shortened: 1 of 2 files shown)


samhocevar:
A very easy repro example (using $RANDOM to bypass hashing):

```
% for d in / /.; do echo PATH=$d; for _ in 1 2 3; do time PATH=$d unknowncommand$RANDOM 2>/dev/null; done; done
PATH=/
PATH=$d unknowncommand$RANDOM 2> /dev/null  0.01s user 0.00s system 1% cpu 1.303 total
PATH=$d unknowncommand$RANDOM 2> /dev/null  0.03s user 0.00s system 2% cpu 1.306 total
PATH=$d unknowncommand$RANDOM 2> /dev/null  0.01s user 0.01s system 2% cpu 1.310 total
PATH=/.
PATH=$d unknowncommand$RANDOM 2> /dev/null  0.00s user 0.00s system 0% cpu 0.010 total
PATH=$d unknowncommand$RANDOM 2> /dev/null  0.00s user 0.00s system 0% cpu 0.010 total
PATH=$d unknowncommand$RANDOM 2> /dev/null  0.00s user 0.00s system 0% cpu 0.010 total
%
```

Biswa96:
Should it be reported to upstream zsh project?

samhocevar:
I don’t really know. Do they care about the Cygwin emulation layer? Because it’s Cygwin that causes `//foo` to not be the same as `/foo` like on most Unix systems, so they aren’t wrong in making the assumption.

lazka:
Just wondering, why do you have "/" in PATH?

lazka:
Since this is an uncommon use case, has a workaround, and should ideally be reported upstream since double slashes should also not be used on other systems, I tend towards closing this.

samhocevar:
I have `/` in `PATH` because:
- my Windows system has `C:\tools\msys64\` in `%PATH%`
- I use `MSYS2_PATH_TYPE=inherit`
- MSYS2 simplifies `C:\tools\msys64\` to `/` when building the Unix-style `$PATH`

PR #5265, opened 2025-03-18
msys2-runtime: update to 3.6.0

jeremyd2019:
msys2/msys2-runtime#260

The HEAD commit is recorded by update-patches.sh in msys2-runtime.commit, and the contents of that file are passed to the --with-msys2-runtime-commit configure argument, so that uname -r shows the msys2-runtime commit hash rather than the commit hash of the upstream Cygwin tag.

Fixes #5223

PR #5269, opened 2025-03-21
msys2-runtime: really restores the previous console mode after spawning a non-MSYS process

dscho:
This is a companion of https://github.com/msys2/msys2-runtime/pull/269.

While at it, also fix a problem in the `update-patches.sh` that led to the wrong commit to be recorded in `msys2-runtime.commit` (i.e. in the output of `uname -a`).


jeremyd2019:
patience... after this PR is merged, the package will be built automatically.  At some point after that, it will be manually uploaded to the repos.  It may take longer than usual since this corresponds with a 'major version' update to Cygwin, so it may warrant more testing/caution before uploading.

dscho:
> so what happens at this stage? Is this the workflow that uploads v3.6.0 to the MSYS2 repo?

This is not _quite_ the workflow that uploads the `msys2-runtime` package to the MSYS2 repo. But it is close. Once this PR is merged (which is only a matter of time, now that the PR build has succeeded), it will be queued for upload.

Please note that the upgrade to v3.6.0 was already merged in #5265 and is currently [queued for upload](https://packages.msys2.org/queue):

image

This queue is where you can monitor whether the update has made it (it will be tracked as 3.6.0-2; note the "-2" suffix, that's the `pkgrel` that indicates that a patch has been applied, in this case it's _your_ patch 😉).

lazka:
There are regressions, so this will sadly have to wait: https://github.com/msys2/msys2-runtime/issues/270

PR #5282, opened 2025-03-25
nnn: Update to 5.1

lazka:


jeremyd2019:
I see several lines of
```
/bin/sh: line 1: pkg-config: command not found
```
building this for i686.  Should pkgconf be a makedepends of this package?

lazka:
thanks, see 8d22551c2ab5e9ca3cc6084f46f3065a6ebaa804

PR #5291, opened 2025-03-29
msys2-runtime: backport pipe hang fix

lazka:
See https://github.com/msys2/msys2-runtime/pull/271

So we can build GIMP until we have an upstream fix


dscho:
I guess I should merge https://github.com/msys2/msys2-runtime/pull/271, then ;-)

PR #5296, opened 2025-03-30
msys2-runtime: more backports

lazka:
See https://github.com/msys2/msys2-runtime/issues/272


dscho:
@lazka should we cherry-pick this into `msys2/msys2-3.6.0`, too? Otherwise the next time I run [`update-patches.sh`](https://github.com/lazka/MSYS2-packages/blob/master/msys2-runtime/update-patches.sh), this patch would be dropped inadvertently.

dscho:
> should we cherry-pick this into `msys2/msys2-3.6.0`, too?

I do think we should, so I opened https://github.com/msys2/msys2-runtime/pull/273

PR #5297, opened 2025-03-30
Add msys2-runtime-3.6

lazka:
and revert msys2-runtime, it's too broken for staging


msys2-runtime/update-patches.sh
```diff
@@ -56,7 +56,7 @@ do
 	sed -i 's/^\(Subject: \[PATCH [0-9]*\/\)[1-9][0-9]*/\1N/' $p ||
 	die "Could not fix Subject: line in $p"
 done &&
-git -C msys2-runtime rev-parse --verify $msys2_branch >msys2-runtime.commit &&
+git -C msys2-runtime rev-parse --verify HEAD >msys2-runtime.commit &&
```
jeremyd2019, line 59: Maybe don't revert this commit, and re-run update-patches.sh so the correct commit is recorded

lazka: ok, thanks

lazka: the commit was already correct

lazka:
I'm gonna "revert" this later today if no one else does, since things look stable now.

jeremyd2019:
> I'm gonna "revert" this later today if no one else does, since things look stable now.

what's the plan for this and the update to 3.6.1?

lazka:
Sorry, I got distracted, I'll do it later.

PR #5308, opened 2025-04-04
llvm: update to 20.1.2

mati865:
Based on mingw-w64 package.


jeremyd2019:
sweet, I was hoping upstream was getting better cygwin support when I read about a cygwin rust target.

mati865:
Actually, right now it's even worse than when I touched it last time...
I'm hoping to upstream some of these changes soon: https://github.com/llvm/llvm-project/pull/134458

llvm/PKGBUILD
```diff
@@ -0,0 +1,280 @@
...
+         "clang-analyzer"
+         "compiler-rt"
+         "lld")
+_version=20.1.2
+_rc=""
+_tag=llvmorg-${_version}${_rc}
+pkgver=${_version}${_rc/-/}
+pkgrel=1
+pkgdesc="C language family frontend for LLVM"
+arch=('x86_64')
```
jeremyd2019, line 17:
```suggestion
arch=('i686' 'x86_64')
```

llvm/PKGBUILD
```diff
@@ -0,0 +1,280 @@
...
+  else
+    common_cmake_args+=(-DCMAKE_BUILD_TYPE=Release
+                        -DLLVM_ENABLE_ASSERTIONS=OFF
+                        -DLLVM_ENABLE_DUMP=ON)
+  fi
+  common_cmake_args+=(-Wno-dev
+    -DCMAKE_INSTALL_PREFIX=/usr
+    -DCOMPILER_RT_BUILD_ORC=OFF
+    -DCOMPILER_RT_DEFAULT_TARGET_ONLY=ON
+    -DLLVM_DEFAULT_TARGET_TRIPLE=x86_64-pc-cygwin
```
jeremyd2019, line 153:
```suggestion
    -DLLVM_DEFAULT_TARGET_TRIPLE=${CARCH}-pc-cygwin
```

llvm/PKGBUILD
```diff
@@ -0,0 +1,280 @@
...
+    -DLLVM_ENABLE_BACKTRACES=OFF
+    -DLLVM_ENABLE_THREADS=ON
+    -DLLVM_HOST_TRIPLE=x86_64-pc-cygwin
```
jeremyd2019, line 156:
```suggestion
    -DLLVM_HOST_TRIPLE=${CARCH}-pc-cygwin
```

jeremyd2019 (commented):
I'd be happy to help try to get this working for 32-bit too, but at least these changes would be required to even try

mati865:
Sure, I'll apply those changes in the morning. There is also one patch that I made as a last-minute fix for building with Clang, but it breaks the build with GCC.

I'm not sure if there is any benefit to keeping LLD. It won't work with Cygwin unless somebody implements linker scripts or incorporates the logic from Cygwin's linker script into LLD.

jeremyd2019:
LLD is at least useful to still be able to cross compile to mingw targets.

llvm/PKGBUILD
```diff
@@ -0,0 +1,280 @@
...
+pkgrel=1
+pkgdesc="C language family frontend for LLVM"
+arch=('x86_64')
+url="https://llvm.org/"
+msys2_references=(
+  "cpe: cpe:/a:llvm:llvm"
+)
+license=("spdx:Apache-2.0 WITH LLVM-exception")
+makedepends=('cmake'
+             'ninja'
```
jeremyd2019, line 24:
```suggestion
             'gcc'
             'ninja'
```

jeremyd2019:
obviously needs a rebase too before CI will run

jeremyd2019:
```
FAILED: tools/clang/tools/libclang/CMakeFiles/obj.libclang.dir/CIndexer.cpp.o 
  /usr/bin/g++.exe -DCINDEX_NO_EXPORTS -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -I/d/_/llvm/src/build/tools/clang/tools/libclang -I/d/_/llvm/src/clang/tools/libclang -I/d/_/llvm/src/clang/include -I/d/_/llvm/src/build/tools/clang/include -I/d/_/llvm/src/build/include -I/d/_/llvm/src/llvm/include -march=nocona -msahf -mtune=generic -O2 -pipe -Wa,-mbig-obj -Werror=date-time -fno-lifetime-dse -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wno-missing-field-initializers -pedantic -Wno-long-long -Wimplicit-fallthrough -Wno-maybe-uninitialized -Wno-nonnull -Wno-class-memaccess -Wno-redundant-move -Wno-pessimizing-move -Wno-noexcept-type -Wdelete-non-virtual-dtor -Wsuggest-override -Wno-comment -Wno-misleading-indentation -Wctad-maybe-unsupported -fdiagnostics-color -fno-common -Woverloaded-virtual -fno-strict-aliasing -O3 -DNDEBUG -std=gnu++17  -fno-exceptions -funwind-tables -MD -MT tools/clang/tools/libclang/CMakeFiles/obj.libclang.dir/CIndexer.cpp.o -MF tools/clang/tools/libclang/CMakeFiles/obj.libclang.dir/CIndexer.cpp.o.d -o tools/clang/tools/libclang/CMakeFiles/obj.libclang.dir/CIndexer.cpp.o -c /d/_/llvm/src/clang/tools/libclang/CIndexer.cpp
  /d/_/llvm/src/clang/tools/libclang/CIndexer.cpp: In member function ‘const std::string& clang::CIndexer::getClangResourcesPath()’:
  /d/_/llvm/src/clang/tools/libclang/CIndexer.cpp:130:3: error: ‘Dl_info’ was not declared in this scope
    130 |   Dl_info info;
        |   ^~~~~~~
  /d/_/llvm/src/clang/tools/libclang/CIndexer.cpp:132:63: error: ‘info’ was not declared in this scope
    132 |   if (dladdr((void *)(uintptr_t)clang_createTranslationUnit, &info) != 0) {
        |                                                               ^~~~
  /d/_/llvm/src/clang/tools/libclang/CIndexer.cpp:132:7: error: ‘dladdr’ was not declared in this scope
    132 |   if (dladdr((void *)(uintptr_t)clang_createTranslationUnit, &info) != 0) {
        |       ^~~~~~
```

jeremyd2019:
That error was probably what you were mentioning.  It should be fine to use dladdr for this purpose on Cygwin, but _GNU_SOURCE needs to be defined for it to be defined by the header.

jeremyd2019:
> I'm not sure if there is any benefit to keeping LLD. It won't work with Cygwin unless somebody implements linker scripts or incorporates the logic from Cygwin's linker script into LLD.

I don't remember seeing a Cygwin-specific *default* linker script.  The cygwin.sc.in linker script for the Cygwin DLL itself is complicated, but I think "normal" Cygwin images are pretty straightforward.

jeremyd2019:
```
FAILED: lib/ExecutionEngine/JITLink/CMakeFiles/LLVMJITLink.dir/JITLink.cpp.o 
/usr/bin/g++.exe -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -I/t/r/llvm/src/build/lib/ExecutionEngine/JITLink -I/t/r/llvm/src/llvm/lib/ExecutionEngine/JITLink -I/t/r/llvm/src/build/include -I/t/r/llvm/src/llvm/include -march=pentium4 -mtune=generic -O2 -pipe -D_GNU_SOURCE=1 -Wa,-mbig-obj -Werror=date-time -fno-lifetime-dse -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wno-missing-field-initializers -pedantic -Wno-long-long -Wimplicit-fallthrough -Wno-maybe-uninitialized -Wno-nonnull -Wno-class-memaccess -Wno-redundant-move -Wno-pessimizing-move -Wno-noexcept-type -Wdelete-non-virtual-dtor -Wsuggest-override -Wno-comment -Wno-misleading-indentation -Wctad-maybe-unsupported -fdiagnostics-color -O3 -DNDEBUG -std=gnu++17  -fno-exceptions -funwind-tables -MD -MT lib/ExecutionEngine/JITLink/CMakeFiles/LLVMJITLink.dir/JITLink.cpp.o -MF lib/ExecutionEngine/JITLink/CMakeFiles/LLVMJITLink.dir/JITLink.cpp.o.d -o lib/ExecutionEngine/JITLink/CMakeFiles/LLVMJITLink.dir/JITLink.cpp.o -c /t/r/llvm/src/llvm/lib/ExecutionEngine/JITLink/JITLink.cpp
/t/r/llvm/src/llvm/include/llvm/ExecutionEngine/JITLink/i386.h:19:26: error: namespace name required before numeric constant
   19 | namespace llvm::jitlink::i386 {
      |                          ^~~~
/t/r/llvm/src/llvm/include/llvm/ExecutionEngine/JITLink/i386.h:19:26: error: expected ‘{’ before numeric constant
/t/r/llvm/src/llvm/include/llvm/ExecutionEngine/JITLink/i386.h:19:26: error: expected unqualified-id before numeric constant
/t/r/llvm/src/llvm/lib/ExecutionEngine/JITLink/JITLink.cpp: In function ‘llvm::jitlink::AnonymousPointerCreator llvm::jitlink::getAnonymousPointerCreator(const llvm::Triple&)’:
/t/r/llvm/src/llvm/lib/ExecutionEngine/JITLink/JITLink.cpp:468:12: error: could not convert ‘1’ from ‘int’ to ‘llvm::jitlink::AnonymousPointerCreator’ {aka ‘llvm::unique_function<llvm::jitlink::Symbol&(llvm::jitlink::LinkGraph&, llvm::jitlink::Section&, llvm::jitlink::Symbol*, long long unsigned int)>’}
  468 |     return i386::createAnonymousPointer;
      |            ^~~~
      |            |
      |            int
/t/r/llvm/src/llvm/lib/ExecutionEngine/JITLink/JITLink.cpp:468:16: error: expected ‘;’ before ‘::’ token
  468 |     return i386::createAnonymousPointer;
      |                ^~
/t/r/llvm/src/llvm/lib/ExecutionEngine/JITLink/JITLink.cpp:468:18: error: ‘::createAnonymousPointer’ has not been declared
  468 |     return i386::createAnonymousPointer;
      |                  ^~~~~~~~~~~~~~~~~~~~~~
/t/r/llvm/src/llvm/lib/ExecutionEngine/JITLink/JITLink.cpp:468:18: note: suggested alternatives:
In file included from /t/r/llvm/src/llvm/lib/ExecutionEngine/JITLink/JITLink.cpp:16:
/t/r/llvm/src/llvm/include/llvm/ExecutionEngine/JITLink/aarch64.h:721:16: note:   ‘llvm::jitlink::aarch64::createAnonymousPointer’
  721 | inline Symbol &createAnonymousPointer(LinkGraph &G, Section &PointerSection,
      |                ^~~~~~~~~~~~~~~~~~~~~~
In file included from /t/r/llvm/src/llvm/lib/ExecutionEngine/JITLink/JITLink.cpp:18:
/t/r/llvm/src/llvm/include/llvm/ExecutionEngine/JITLink/loongarch.h:596:16: note:   ‘llvm::jitlink::loongarch::createAnonymousPointer’
  596 | inline Symbol &createAnonymousPointer(LinkGraph &G, Section &PointerSection,
      |                ^~~~~~~~~~~~~~~~~~~~~~
In file included from /t/r/llvm/src/llvm/lib/ExecutionEngine/JITLink/JITLink.cpp:19:
/t/r/llvm/src/llvm/include/llvm/ExecutionEngine/JITLink/x86_64.h:608:16: note:   ‘llvm::jitlink::x86_64::createAnonymousPointer’
  608 | inline Symbol &createAnonymousPointer(LinkGraph &G, Section &PointerSection,
      |                ^~~~~~~~~~~~~~~~~~~~~~
/t/r/llvm/src/llvm/lib/ExecutionEngine/JITLink/JITLink.cpp: In function ‘llvm::jitlink::PointerJumpStubCreator llvm::jitlink::getPointerJumpStubCreator(const llvm::Triple&)’:
/t/r/llvm/src/llvm/lib/ExecutionEngine/JITLink/JITLink.cpp:484:12: error: could not convert ‘1’ from ‘int’ to ‘llvm::jitlink::PointerJumpStubCreator’ {aka ‘llvm::unique_function<llvm::jitlink::Symbol&(llvm::jitlink::LinkGraph&, llvm::jitlink::Section&, llvm::jitlink::Symbol&)>’}
  484 |     return i386::createAnonymousPointerJumpStub;
      |            ^~~~
      |            |
      |            int
/t/r/llvm/src/llvm/lib/ExecutionEngine/JITLink/JITLink.cpp:484:16: error: expected ‘;’ before ‘::’ token
  484 |     return i386::createAnonymousPointerJumpStub;
      |                ^~
/t/r/llvm/src/llvm/lib/ExecutionEngine/JITLink/JITLink.cpp:484:18: error: ‘::createAnonymousPointerJumpStub’ has not been declared
  484 |     return i386::createAnonymousPointerJumpStub;
      |                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/t/r/llvm/src/llvm/lib/ExecutionEngine/JITLink/JITLink.cpp:484:18: note: suggested alternatives:
/t/r/llvm/src/llvm/include/llvm/ExecutionEngine/JITLink/aarch64.h:750:16: note:   ‘llvm::jitlink::aarch64::createAnonymousPointerJumpStub’
  750 | inline Symbol &createAnonymousPointerJumpStub(LinkGraph &G,
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/t/r/llvm/src/llvm/include/llvm/ExecutionEngine/JITLink/loongarch.h:609:16: note:   ‘llvm::jitlink::loongarch::createAnonymousPointerJumpStub’
  609 | inline Symbol &createAnonymousPointerJumpStub(LinkGraph &G,
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/t/r/llvm/src/llvm/include/llvm/ExecutionEngine/JITLink/x86_64.h:636:16: note:   ‘llvm::jitlink::x86_64::createAnonymousPointerJumpStub’
  636 | inline Symbol &createAnonymousPointerJumpStub(LinkGraph &G,
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[3327/4238] Building CXX object lib/ExecutionEngine/JITLink/CMakeFiles/LLVMJITLink.dir/JITLinkMemoryManager.cpp.o
ninja: build stopped: subcommand failed.
```

So something has defined `i386` as an int (or maybe `#define i386 1`), breaking its use as a namespace?

jeremyd2019:
```console
$ gcc -dM -E - < /dev/null | grep i386
#define __i386 1
#define i386 1
#define __i386__ 1
```

mati865:
> > I'm not sure if there is any benefit to keeping LLD. It won't work with Cygwin unless somebody implements linker scripts or incorporates the logic from Cygwin's linker script into LLD.
> 
> I don't remember seeing a Cygwin-specific _default_ linker script. The cygwin.sc.in linker script for the Cygwin DLL itself is complicated, but I think "normal" Cygwin images are pretty straightforward.

There are symbols created by linker scripts in `/usr/x86_64-pc-msys/lib/ldscripts/`, so unless the logic is ported to LLD, the linker scripts are necessary:
```
$ bat -p hello.cc
#include <iostream>

int main() {
    std::cout << "Hello" << std::endl;
    return 0;
}

$ g++ hello.cc -fuse-ld=lld
ld.lld: error: undefined symbol: __data_start__
>>> referenced by /d/S/B/src/msys2-runtime/winsup/cygwin/lib/_cygwin_crt0_common.cc:158
>>>               libmsys-2.0.a(_cygwin_crt0_common.o):(.text)

ld.lld: error: undefined symbol: __bss_start__
>>> referenced by /d/S/B/src/msys2-runtime/winsup/cygwin/lib/_cygwin_crt0_common.cc:158
>>>               libmsys-2.0.a(_cygwin_crt0_common.o):(.text)

ld.lld: error: undefined symbol: __data_end__
>>> referenced by /d/S/B/src/msys2-runtime/winsup/cygwin/lib/_cygwin_crt0_common.cc:158
>>>               libmsys-2.0.a(_cygwin_crt0_common.o):(.text)

ld.lld: error: undefined symbol: __bss_end__
>>> referenced by /d/S/B/src/msys2-runtime/winsup/cygwin/lib/_cygwin_crt0_common.cc:158
>>>               libmsys-2.0.a(_cygwin_crt0_common.o):(.text)
collect2: error: ld returned 1 exit status
```

jeremyd2019:
ah, yeah, cygwin would need to know that to know what to copy in their fork emulation.  I didn't notice this because those symbols are present in *all* binutils pe linker scripts, not just cygwin, so there's not a special case cygwin linker script.

jeremyd2019:
@mstorsjo how hard would it be to define `__data_start__` `__data_end__` `__bss_start__` `__bss_end__` in LLD, and then probably include `.data_cygwin_nocopy` input sections in the output `.data` section after `__data_end__` (assuming LLD's not handling that yet either)?  Probably not very.  I'm currently bogged down with some binutils patches, really quite remarkably slow to get through review largely due to timezone differences (I only tend to get replies around midnight my time, so it ends up being a 24-hour round trip on every message).

jeremyd2019:
build successfully for i686 with the `#undef i386` - now I'm trying to use it to build itself.  Note the clang and llvm packages are *huge*
```
445M Apr  6  2025 clang-20.1.2-1-i686.pkg.tar.zst
77K Apr  6  2025 clang-analyzer-20.1.2-1-i686.pkg.tar.zst
112K Apr  6  2025 compiler-rt-20.1.2-1-i686.pkg.tar.zst
116M Apr  6  2025 lld-20.1.2-1-i686.pkg.tar.zst
97M Apr  6  2025 llvm-20.1.2-1.src.tar.zst
364M Apr  6  2025 llvm-20.1.2-1-i686.pkg.tar.zst
```

jeremyd2019:
nope, llvm-min-tblgen seems to crash:
```console
(gdb) bt
#0  0x5e92e028 in ?? () from /usr/bin/msys-stdc++-6.dll
#1  0x004e6983 in void llvm::cl::initializer<char [2]>::apply<llvm::cl::opt<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, false, llvm::cl::parser<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > >(llvm::cl::opt<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, false, llvm::cl::parser<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&) const
    ()
#2  0x004e53ac in llvm::cl::opt<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, false, llvm::cl::parser<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >::opt<char [2], llvm::cl::desc, llvm::cl::value_desc, llvm::cl::initializer<char [2]> >(char const (&) [2], llvm::cl::desc const&, llvm::cl::value_desc const&, llvm::cl::initializer<char [2]> const&) ()
#3  0x004e50c9 in _GLOBAL__sub_I_Main.cpp ()
#4  0x61008052 in do_global_ctors (force=0,
    in_pfunc=0x543f70 <_rt_psrelocs_end>)
    at /t/r/msys2-runtime-3.3/src/msys2-runtime/winsup/cygwin/dcrt0.cc:78
#5  __main ()
    at /t/r/msys2-runtime-3.3/src/msys2-runtime/winsup/cygwin/dcrt0.cc:1167
#6  0x80049657 in ?? ()
#7  0x61009bdf in dll_crt0_1 ()
    at /t/r/msys2-runtime-3.3/src/msys2-runtime/winsup/cygwin/dcrt0.cc:1055
#8  0x610076f2 in _cygtls::call2(unsigned long (*)(void*, void*), void*, void*)@16 (this=0x159ce64, func=0x61008af0 <dll_crt0_1(void*)>, arg=0x0, buf=0x0)
    at /t/r/msys2-runtime-3.3/src/msys2-runtime/winsup/cygwin/cygtls.cc:40
#9  0x61008af0 in child_info_fork::handle_fork() ()
```

jeremyd2019:
hmm
```
   0x004e6976 <+118>:   lea    0x4c(%esi),%ecx
   0x004e6979 <+121>:   lea    0x4(%esp),%edi
   0x004e697d <+125>:   push   %edi
   0x004e697e <+126>:   call   0x5273e4 <_ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE9_M_assignERKS4_>
```
Maybe a `thiscall` mismatch?  It seems maybe the clang-generated code wants to put `this` in ecx, but the g++ dll expects it on the stack?  Just guessing at this point

llvm/0102-Clang-Cygwin-Enable-few-conditions-that-are-shared-w.patch
```diff
@@ -0,0 +1,183 @@
...
+diff --git a/clang/lib/AST/ItaniumCXXABI.cpp b/clang/lib/AST/ItaniumCXXABI.cpp
+index a1b2551419f5..d866f6023e14 100644
+--- a/clang/lib/AST/ItaniumCXXABI.cpp
++++ b/clang/lib/AST/ItaniumCXXABI.cpp
+@@ -236,7 +236,7 @@ public:
+ 
+   CallingConv getDefaultMethodCallConv(bool isVariadic) const override {
+     const llvm::Triple &T = Context.getTargetInfo().getTriple();
+-    if (!isVariadic && T.isWindowsGNUEnvironment() &&
++    if (!isVariadic && T.isOSCygMing() &&
```
jeremyd2019, line 28: Random thought: maybe this change was wrong?  I'll try reverting this

mati865: Very likely, I didn't test 32-bit binaries at all. I just assumed Cygwin would behave just like MinGW in this case.

mstorsjo:
> @mstorsjo how hard would it be to define `__data_start__` `__data_end__` `__bss_start__` `__bss_end__` in LLD, and then probably include `.data_cygwin_nocopy` input sections in the output `.data` section after `__data_end__` (assuming LLD's not handling that yet either)? Probably not very.

Yeah that's probably not very hard - but it's probably best if it's done by someone who is otherwise more familiar with the cygwin efforts.

So this PR tries to build LLVM/Clang as msys2 binaries (using a preexisting msys2 targeting GCC?), and then intending to be able to use that Clang for targeting msys2?

mati865:
Yes, it's kind of upgrade. Cygwin/MSYS2 LLVM/Clang are already packaged and sort of work (it sometimes segfaults), but the version is terribly out of date. It was used only to bootstrap CLANG* environments.

Times has changed and with the creation of Rust's Cygwin target it'd be useful to have working LLVM/Clang. I think LLD is not that important (unless we'd want to build Cygwin libraries with LLVM toolchain, but I don't see it happening soon).

> Note the clang and llvm packages are _huge_

Yeah, you have to link statically with GCC.

jeremyd2019:
> > @mstorsjo how hard would it be to define `__data_start__` `__data_end__` `__bss_start__` `__bss_end__` in LLD, and then probably include `.data_cygwin_nocopy` input sections in the output `.data` section after `__data_end__` (assuming LLD's not handling that yet either)? Probably not very.
> 
> Yeah that's probably not very hard - but it's probably best if it's done by someone who is otherwise more familiar with the cygwin efforts.

I might try to hack something together for a patch for this package anyway

jeremyd2019:
> > Note the clang and llvm packages are _huge_
> 
> Yeah, you have to link statically with GCC.

Possibly also with binutils rather than LLD?

mati865:
Martin implemented the support for excluding the symbols to Binutils in 2.40 release, so it shouldn't be the problem. That saidm I couldn't get the visibility to work with Clang when targeting Cygwin.

llvm/0102-Clang-Cygwin-Enable-few-conditions-that-are-shared-w.patch
jeremyd2019: With this change backed out, clang can build itself on i686

jeremyd2019:
> > Note the clang and llvm packages are _huge_
> 
> Yeah, you have to link statically with GCC.

built with clang:
```console
$ pacman -U llvm-20.1.2-1-i686.pkg.tar.zst clang-20.1.2-1-i686.pkg.tar.zst compiler-rt-20.1.2-1-i686.pkg.tar.zst
loading packages...
warning: llvm-20.1.2-1 is up to date -- reinstalling
warning: clang-20.1.2-1 is up to date -- reinstalling
warning: compiler-rt-20.1.2-1 is up to date -- reinstalling
resolving dependencies...

Packages (3) clang-20.1.2-1  compiler-rt-20.1.2-1  llvm-20.1.2-1

Total Installed Size:  2989.12 MiB
Net Upgrade Size:      -589.69 MiB
```

jeremyd2019:
@mstorsjo in 54ac10d I hacked something together that seems to work (except for https://github.com/llvm/llvm-project/issues/134558).  I'd appreciate some feedback.  Is the empty chunk approach the best way, or should I follow more like edataSec (using `->chunks.front()` and `->chunks.back()`, and passing `->getSize()` to the optional `offset` parameter to `replaceSymbol<DefinedSynthetic>` ?

jeremyd2019:
still get too many exported symbols even with LLD.

jeremyd2019:
ugh, building with `-DLLVM_ENABLE_LLD` I'm seeing issues now due to the PE flags/DLL characteristics.  For Cygwin targets, those nxcompat/dynamicbase flags were backed out in binutils.  I'll have to add yet another hack to change the default there.

jeremyd2019:
I'm getting one linker error for c-index-test.exe:
```
/usr/lib/gcc/i686-pc-msys/13.3.0/../../../../i686-pc-msys/bin/ld: lib/libclangSerialization.a(PCHContainerOperations.cpp.o):PCHContainerOperations.cpp:(.text+0xa0): multiple definition of
 `clang::PCHContainerOperations::PCHContainerOperations()';
lib/libclang.dll.a(msys_clang_20_1_dll_d011134.o):(.text+0x0): first defined here
collect2: error: ld returned 1 exit status
```

I still have more screwing around with lld to do.

jeremyd2019:
can anyone explain why I'm seeing this:
```c
    warn("Triple: " + sys::getDefaultTargetTriple());
  bool isCygwinTarget = Triple(sys::getDefaultTargetTriple())
                        .isWindowsCygwinEnvironment();

    warn(StringRef("isCygwinTarget: ") + (isCygwinTarget ? "true" : "false"));
```

```
$ PATH=./src/build/bin:$PATH g++ hello.cc -fuse-ld=lld
lld: warning: Triple: i686-pc-cygwin
lld: warning: isCygwinTarget: false
```

llvm/PKGBUILD
```diff
@@ -0,0 +1,289 @@
...
+  else
+    common_cmake_args+=(-DCMAKE_BUILD_TYPE=Release
+                        -DLLVM_ENABLE_ASSERTIONS=OFF
+                        -DLLVM_ENABLE_DUMP=ON)
+  fi
+  common_cmake_args+=(-Wno-dev
+    -DCMAKE_INSTALL_PREFIX=/usr
+    -DCOMPILER_RT_BUILD_ORC=OFF
+    -DCOMPILER_RT_DEFAULT_TARGET_ONLY=ON
+    -DLLVM_DEFAULT_TARGET_TRIPLE=${CARCH}-pc-cygwin
```
jeremyd2019, line 160: I wonder if this shouldn't be
```suggestion
    -DLLVM_DEFAULT_TARGET_TRIPLE=${CARCH}-pc-windows-cygnus
```
as the normalized triple

jeremyd2019:
OK, dunno what I screwed up that's making it try to build a shared libclang now.

mstorsjo:
> can anyone explain why I'm seeing this:
> 
> ```c
>     warn("Triple: " + sys::getDefaultTargetTriple());
>   bool isCygwinTarget = Triple(sys::getDefaultTargetTriple())
>                         .isWindowsCygwinEnvironment();
> 
>     warn(StringRef("isCygwinTarget: ") + (isCygwinTarget ? "true" : "false"));
> ```
> 
>  ```
> $ PATH=./src/build/bin:$PATH g++ hello.cc -fuse-ld=lld
> lld: warning: Triple: i686-pc-cygwin
> lld: warning: isCygwinTarget: false
> ```

Yes - unfortunately, `Triple()` parses things strictly, you need to do `Triple(Triple::normalize(sys::getDefaultTargetTriple()))`.

This is a bit unfortunate, because the common way of writing triples like `i686-w64-mingw32` does work for `LLVM_HOST_TRIPLE` (as the default host/target triple) in _most_ cases (e.g. Clang does normalize it somewhere along the way so it always works), but I do remember that some tools, like `llc`, don't normalize it, so one ends up with odd results.

So ideally, most uses of `getDefaultTargetTriple()` probably needs to explicitly normalize it. The odd thing is that `getProcessTriple()` does return normalized triples, see https://github.com/llvm/llvm-project/blob/llvmorg-20.1.2/llvm/lib/TargetParser/Host.cpp#L2269-L2271 - in contrast to https://github.com/llvm/llvm-project/blob/llvmorg-20.1.2/llvm/lib/TargetParser/Windows/Host.inc#L24-L35 and https://github.com/llvm/llvm-project/blob/llvmorg-20.1.2/llvm/lib/TargetParser/Unix/Host.inc#L75-L87.

mstorsjo:
> @mstorsjo in [54ac10d](https://github.com/msys2/MSYS2-packages/commit/54ac10ddf9edbac31d8a3077d9b7d0d93a17a329) I hacked something together that seems to work (except for [llvm/llvm-project#134558](https://github.com/llvm/llvm-project/issues/134558)). I'd appreciate some feedback. Is the empty chunk approach the best way, or should I follow more like edataSec (using `->chunks.front()` and `->chunks.back()`, and passing `->getSize()` to the optional `offset` parameter to `replaceSymbol<DefinedSynthetic>` ?

Hmm. Offhand this does seem fine. I don't really remember any specific case why one would use one or the other approach. (We don't really seem to use `EmptyChunk` all that much though - I wonder why. It seems like I added `EmptyChunk` for the runtime pseudo reloc stuff, and nothing else has picked up using it.) @cjacek has poked such things a bit more recently than me, do you remember any reason for doing this one way or the other?

cjacek:
AFAICS, `EmptyChunk` would mostly work, but I'm not sure if it's desired for corner cases. If the section is empty, adding such chunks would make it no longer empty, so `removeUnusedSections()` would not remove it, `assignAddresses()` would assign it addresses, but `removeEmptySections()` would remove later, so those symbols would point to address occupied by other section. In such case start and end would have the same address still, so maybe that's not a problem, I don't know what's expected behavior in such case.

llvm/PKGBUILD
jeremyd2019: oops, should have done the same to the host triple too, I missed that

jeremyd2019:
> > OK, dunno what I screwed up that's making it try to build a shared libclang now. Seems `LLVM_ENABLE_PIC` is `ON`
> 
> That shouldn't be the problem. PIC should be on (not that it changes anything on Windows, but mingw-w64 builds also enable it) and I have the patch for the shared lib: `0106-disable-shared-libs-on-Cygwin-by-default.patch`

See [b3eb870](https://github.com/msys2/MSYS2-packages/pull/5308/commits/b3eb87010f7adab1e91259444e1a8906885cc8d3)

> > still get too many exported symbols even with LLD.
> 
> Yeah, as I said, the problem is with handling symbol visibility by Clang. Both ld.bfd and LLD should have no problem. It might be easier to use the tests to find out the reason, but I don't want to invest time in it.

See [d27688e](https://github.com/msys2/MSYS2-packages/pull/5308/commits/d27688e0fe89834c66858237b7f83a4971ecbc92)

jeremyd2019:
Hmm, seems LLD ignores `--enable-auto-image-base`.  That's not gonna work for Cygwin.

mati865:
> > > OK, dunno what I screwed up that's making it try to build a shared libclang now. Seems `LLVM_ENABLE_PIC` is `ON`
> > 
> > 
> > That shouldn't be the problem. PIC should be on (not that it changes anything on Windows, but mingw-w64 builds also enable it) and I have the patch for the shared lib: `0106-disable-shared-libs-on-Cygwin-by-default.patch`
> 
> See [b3eb870](https://github.com/msys2/MSYS2-packages/pull/5308/commits/b3eb87010f7adab1e91259444e1a8906885cc8d3)
> 
> > > still get too many exported symbols even with LLD.
> > 
> > 
> > Yeah, as I said, the problem is with handling symbol visibility by Clang. Both ld.bfd and LLD should have no problem. It might be easier to use the tests to find out the reason, but I don't want to invest time in it.
> 
> See [d27688e](https://github.com/msys2/MSYS2-packages/pull/5308/commits/d27688e0fe89834c66858237b7f83a4971ecbc92)

Ah, I had them here https://github.com/llvm/llvm-project/compare/main...mati865:llvm-project:cygwin-more-fixes but didn't update the patches.

> Hmm, seems LLD ignores `--enable-auto-image-base`. That's not gonna work for Cygwin.

Do you mean `--disable-auto-image-base`? It's enabled by default.
Adding code to handle it won't be hard.

jeremyd2019:
> Ah, I had them here [llvm/llvm-project@main...mati865:llvm-project:cygwin-more-fixes](https://github.com/llvm/llvm-project/compare/main...mati865:llvm-project:cygwin-more-fixes) but didn't update the patches.

I extended/modified your patches.

> > Hmm, seems LLD ignores `--enable-auto-image-base`. That's not gonna work for Cygwin.
> 
> Do you mean `--disable-auto-image-base`? It's enabled by default. Adding code to handle it won't be hard.

I mean that both enable and disable do nothing - there is no hashing of the output path to come up with a (hopefully) unique image base.

jeremyd2019:
With these changes I can build shared with clang on i686 (with GNU ld, LLD is still busted for Cygwin target), but I still need `-DLIBCLANG_BUILD_STATIC=ON` to work around the multiple definition error linking c-index-test.exe.  When this PR run finishes I'll probably take the artifacts and see if that also happens on x86_64, and try to patch it somehow.

mstorsjo:
> > Yeah, as I said, the problem is with handling symbol visibility by Clang. Both ld.bfd and LLD should have no problem. It might be easier to use the tests to find out the reason, but I don't want to invest time in it.
> 
> See [d27688e](https://github.com/msys2/MSYS2-packages/pull/5308/commits/d27688e0fe89834c66858237b7f83a4971ecbc92)

That only fixes one aspect; there are a couple other places as well. See https://github.com/llvm/llvm-project/commit/2c2fb0c7375061147711cd21396d79faad7dfdfb where this was added.

jeremyd2019:
> > > Yeah, as I said, the problem is with handling symbol visibility by Clang. Both ld.bfd and LLD should have no problem. It might be easier to use the tests to find out the reason, but I don't want to invest time in it.
> > 
> > 
> > See [d27688e](https://github.com/msys2/MSYS2-packages/pull/5308/commits/d27688e0fe89834c66858237b7f83a4971ecbc92)
> 
> That only fixes one aspect; there are a couple other places as well. See [llvm/llvm-project@2c2fb0c](https://github.com/llvm/llvm-project/commit/2c2fb0c7375061147711cd21396d79faad7dfdfb) where this was added.

I *think* the other places were already dealt with in @mati865's patches, just this one cmake file was missing.  But it's possible something else was missed

mstorsjo:
> well adding this seems to work around it:
> 
> ```cmake
>   if(CYGWIN)
>     # Crazy linker issue
>     target_link_options(c-index-test PRIVATE LINKER:--allow-multiple-definition)
>   endif()
> ```
> 
> but, WTF? it's getting one definition from the libclang import library, and the other from the static clangSerialization library. I'm not sure what else it's pulling in from that archive member in the static clangSerialization though.

I'm not sure that I understand the setup here, but if you're using a dynamic libclang-cpp library, wouldn't all the components be linked from there, and not from a separate static clangSerialization library? Whenever the libllvm/libclang-cpp dylibs are enabled, it usually replaces all the regular small static libraries with those dylibs instead.

jeremyd2019:
I don't know why, but see https://github.com/llvm/llvm-project/blob/369c7739d0853b7931410037843d5a63f50bc0a1/clang/tools/c-index-test/CMakeLists.txt#L25-L33.  Removing those extra libs gives a bunch of undefined symbols.

jeremyd2019:
actually, manually stripping clangSerialization from the link in the generated ninja file for c-index-test results in just this undefined symbol:

```
/usr/lib/gcc/i686-pc-msys/13.3.0/../../../../i686-pc-msys/bin/ld: tools/clang/tools/c-index-test/CMakeFiles/c-index-test.dir/core_main.cpp.o:core_main.cpp:
(.rdata$.refptr.__ZTVN5clang28ObjectFilePCHContainerReaderE+0x0): undefined reference to `vtable for clang::ObjectFilePCHContainerReader'
```

mstorsjo:
I tested building `c-test-index.exe` in a mingw/dylib build, and it does link successfully, but I do see that it passes `lib/libclang.dll.a  lib/libclangAST.a  lib/libclangBasic.a  lib/libclangFrontend.a  lib/libclangIndex.a  lib/libclangSerialization.a  lib/libclangFrontend.a  lib/libclangDriver.a  -lversion  lib/libclangParse.a  lib/libclangSerialization.a  lib/libclangSema.a  lib/libclangAPINotes.a  lib/libclangEdit.a  lib/libclangAnalysis.a  lib/libclangASTMatchers.a  lib/libclangAST.a  lib/libclangSupport.a  lib/libclangFormat.a  lib/libclangToolingInclusions.a  lib/libclangToolingCore.a  lib/libclangRewrite.a  lib/libclangLex.a  lib/libclangBasic.a  lib/libLLVM-21git.dll.a` to the linker. So if symbols from `libclangSerialization.a` are provided by `libclang.dll.a` too (I didn't check), there's maybe a risk for some duplication, but if `libclang.dll.a` covers it all, nothing should be pulled in from `libclangSerialization.a`.

jeremyd2019:
Same error on x86_64:
```
/usr/lib/gcc/x86_64-pc-msys/13.3.0/../../../../x86_64-pc-msys/bin/ld:
lib/libclangSerialization.a(PCHContainerOperations.cpp.o):PCHContainerOperations.cpp:(.text+0xc0):
multiple definition of `clang::PCHContainerOperations::PCHContainerOperations()';
lib/libclang.dll.a(msys_clang_20_1_dll_d011128.o):(.text+0x0): first defined here
```

jeremyd2019:
```
$ llvm-nm.exe -B /clang64/lib/libclang-cpp.dll.a | grep ZTVN5clang28ObjectFilePCHContainerReader
00000000 D __imp__ZTVN5clang28ObjectFilePCHContainerReaderE

$ llvm-nm.exe -B src/build/lib/libclang-cpp.dll.a | grep ZTVN5clang28ObjectFilePCHContainerReader
00000000 I __imp__ZTVN5clang28ObjectFilePCHContainerReaderE
00000000 I __nm__ZTVN5clang28ObjectFilePCHContainerReaderE
```

maybe it's the auto-import then?

jeremyd2019:
LLD (broken as it is) is able to link this file without error.  binutils screwup I guess.

jeremyd2019:
I have a hacky patch to lld/MinGW to make it MinGW/Cygwin, but now I'm getting this with clang linked with LLD (dylib):
```
FAILED: CMakeFiles/clang_rt.builtins-x86_64.dir/absvdi2.c.o
/d/S/llvm/src/build/bin/clang.exe   -march=nocona -msahf -mtune=generic -O2 -pipe -D_GNU_SOURCE=1 -o CMakeFiles/clang_rt.builtins-x86_64.dir/absvdi2.c.o -c /d/S/llvm/src/compiler-rt/lib/builtins/absvdi2.c
: CommandLine Error: Option 'verify-analysis-invalidation' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
```

Also, I'm not sure I like having "Cygwin behavior" switched by the default target triple, because I'd still like it to be usable for cross-compiling to MinGW normally.

Note that I implemented `--enable-auto-image-base` using an LLVM hash thingy rather than trying to replicate (or copy/paste) the hash function from binutils.  I figure it doesn't really matter if the hash algorithm is the same or not.

<details>
<summary>Hacky patch to turn MinGW driver into Cygwin</summary>

```patch
--- a/lld/MinGW/Driver.cpp	2025-04-07 14:55:18.420801800 -0700
+++ b/lld/MinGW/Driver.cpp	2025-04-07 17:27:27.318509600 -0700
@@ -177,6 +177,24 @@
   return "";
 }
 
+static inline bool isI386Target(const opt::InputArgList &args,
+                                const Triple &defaultTarget) {
+  auto *a = args.getLastArg(OPT_m);
+  if (a)
+    return StringRef(a->getValue()) == "i386pe";
+  return defaultTarget.getArch() == Triple::x86;
+}
+
+static inline bool is64bitTarget(const opt::InputArgList &args,
+                                 const Triple &defaultTarget) {
+  auto *a = args.getLastArg(OPT_m);
+  if (a)
+    return StringRef(a->getValue()) == "i386pep" ||
+					 StringRef(a->getValue()) == "arm64pe" ||
+					 StringRef(a->getValue()) == "arm64ecpe";
+  return defaultTarget.isArch64Bit();
+}
+
 namespace lld {
 namespace coff {
 bool link(ArrayRef<const char *> argsArr, llvm::raw_ostream &stdoutOS,
@@ -222,6 +240,9 @@
     return false;
   }
 
+  Triple defaultTarget(Triple::normalize(sys::getDefaultTargetTriple()));
+  bool isCygwinTarget = defaultTarget.isWindowsCygwinEnvironment();
+
   std::vector<std::string> linkArgs;
   auto add = [&](const Twine &s) { linkArgs.push_back(s.str()); };
 
@@ -230,7 +251,7 @@
 
   if (auto *a = args.getLastArg(OPT_entry)) {
     StringRef s = a->getValue();
-    if (args.getLastArgValue(OPT_m) == "i386pe" && s.starts_with("_"))
+    if (isI386Target(args, defaultTarget) && s.starts_with("_"))
       add("-entry:" + s.substr(1));
     else if (!s.empty())
       add("-entry:" + s);
@@ -287,6 +308,25 @@
     add("-output-def:" + StringRef(a->getValue()));
   if (auto *a = args.getLastArg(OPT_image_base))
     add("-base:" + StringRef(a->getValue()));
+  else if (isCygwinTarget) {
+	 if (args.hasFlag(OPT_enable_auto_image_base, OPT_disable_auto_image_base,
+				 false) && args.hasArg(OPT_shared)) {
+		 hash_code outhash;
+		 uint64_t autobase;
+		 if (auto *a = args.getLastArg(OPT_o))
+			 outhash = hash_value(StringRef(a->getValue()));
+		 else
+			 outhash = hash_value(StringRef("a.dll"));
+		 if (is64bitTarget(args, defaultTarget))
+			 autobase = 0x400000000ULL + (((uint64_t)outhash << 16) & 0x1ffff0000ULL);
+		 else
+			 autobase =
+				 (uint32_t)(0x61500000 + (((uint32_t)outhash << 16) & 0x0FFC0000));
+		 add("-base:0x" + Twine::utohexstr(autobase));
+	 } else if (is64bitTarget(args, defaultTarget)) {
+		add(args.hasArg(OPT_shared) ? "-base:0x400000000" : "-base:0x100400000");
+	 }
+	}
   if (auto *a = args.getLastArg(OPT_map))
     add("-lldmap:" + StringRef(a->getValue()));
   if (auto *a = args.getLastArg(OPT_reproduce))
@@ -373,14 +413,17 @@
   if (args.hasFlag(OPT_no_seh, OPT_disable_no_seh, false))
     add("-noseh");
 
-  if (args.getLastArgValue(OPT_m) != "thumb2pe" &&
-      args.getLastArgValue(OPT_m) != "arm64pe" &&
-      args.getLastArgValue(OPT_m) != "arm64ecpe" &&
-      args.hasFlag(OPT_disable_dynamicbase, OPT_dynamicbase, false))
+  if ((args.getLastArgValue(OPT_m) != "thumb2pe" &&
+       args.getLastArgValue(OPT_m) != "arm64pe" &&
+       args.getLastArgValue(OPT_m) != "arm64ecpe" &&
+       args.hasFlag(OPT_disable_dynamicbase, OPT_dynamicbase, false)) ||
+      (isCygwinTarget && !args.hasArg(OPT_dynamicbase)))
     add("-dynamicbase:no");
-  if (args.hasFlag(OPT_disable_high_entropy_va, OPT_high_entropy_va, false))
+  if (args.hasFlag(OPT_disable_high_entropy_va, OPT_high_entropy_va, false) ||
+      (isCygwinTarget && !args.hasArg(OPT_high_entropy_va)))
     add("-highentropyva:no");
-  if (args.hasFlag(OPT_disable_nxcompat, OPT_nxcompat, false))
+  if (args.hasFlag(OPT_disable_nxcompat, OPT_nxcompat, false) ||
+      (isCygwinTarget && !args.hasArg(OPT_nxcompat)))
     add("-nxcompat:no");
   if (args.hasFlag(OPT_disable_tsaware, OPT_tsaware, false))
     add("-tsaware:no");
@@ -527,7 +570,7 @@
   for (auto *a : args.filtered(OPT_Xlink))
     add(a->getValue());
 
-  if (args.getLastArgValue(OPT_m) == "i386pe")
+  if (isI386Target(args, defaultTarget))
     add("-alternatename:__image_base__=___ImageBase");
   else
     add("-alternatename:__image_base__=__ImageBase");
--- a/lld/MinGW/Options.td	2025-04-07 14:55:18.388986300 -0700
+++ b/lld/MinGW/Options.td	2025-04-07 17:03:20.201184300 -0700
@@ -59,6 +59,7 @@
 defm demangle: B<"demangle",
     "Demangle symbol names (default)",
     "Do not demangle symbol names">;
+def disable_auto_image_base: F<"disable-auto-image-base">;
 def disable_auto_import: F<"disable-auto-import">,
     HelpText<"Don't automatically import data symbols from other DLLs without dllimport">;
 def disable_runtime_pseudo_reloc: F<"disable-runtime-pseudo-reloc">,
@@ -68,6 +69,7 @@
 defm dll_search_prefix: EEq<"dll-search-prefix",
     "Prefer DLL with this prefix if importlib is missing">;
 defm dynamicbase: B_disable<"dynamicbase", "Enable ASLR", "Disable ASLR">;
+def enable_auto_image_base: F<"enable-auto-image-base">;
 def enable_auto_import: F<"enable-auto-import">,
     HelpText<"Automatically import data symbols from other DLLs where needed">;
 def enable_runtime_pseudo_reloc: F<"enable-runtime-pseudo-reloc">,
@@ -238,8 +240,6 @@
 // Ignored options
 def: Joined<["-"], "O">;
 def: F<"as-needed">;
-def: F<"disable-auto-image-base">;
-def: F<"enable-auto-image-base">;
 def: F<"end-group">;
 def: Flag<["--"], "full-shutdown">;
 defm: EqNoHelp<"major-image-version">;
```
</details>

jeremyd2019:
I'm thinking it might be better for now to punt on lld support for Cygwin target, and instead make sure the existing mingw-w64-cross packages that use clang work being updated/rebuilt with this new clang.

mstorsjo:
> I have a hacky patch to lld/MinGW to make it MinGW/Cygwin, but now I'm getting this with clang linked with LLD (dylib):
> 
> ```
> FAILED: CMakeFiles/clang_rt.builtins-x86_64.dir/absvdi2.c.o
> /d/S/llvm/src/build/bin/clang.exe   -march=nocona -msahf -mtune=generic -O2 -pipe -D_GNU_SOURCE=1 -o CMakeFiles/clang_rt.builtins-x86_64.dir/absvdi2.c.o -c /d/S/llvm/src/compiler-rt/lib/builtins/absvdi2.c
> : CommandLine Error: Option 'verify-analysis-invalidation' registered more than once!
> LLVM ERROR: inconsistency in registered CommandLine options
> ```

Yeah that sounds like some other inconsistency around the dylib build, with things getting pulled in both from static libraries while also being present in the dylib.

> Also, I'm not sure I like having "Cygwin behavior" switched by the default target triple, because I'd still like it to be usable for cross-compiling to MinGW normally.

Indeed, it doesn't seem ideal... It may be ok to pick some implicit defaults based on the default triple, but it should also be possible to do the same while cross compiling, and with a build of LLD that doesn't default to a cygwin target. So I'm not sure if that would require inventing new `-m` parameter names for the cygwin configurations, or if there are other ways of achieving that? Inventing new `-m` parameter names is not ideal wrt GCC interop, as they probably won't take on that. (I tested a cygwin cross compiler build, and it does pass an `-m` option to `ld`, but it just uses the regular `i386pep` on x86_64.)

jeremyd2019:
got as far as mingw-w64-cross-clang but the runtimes are really fighting me.  Getting
```
In file included from /d/S/mingw-w64-cross-clang/src/libcxx/src/charconv.cpp:12:
/d/S/mingw-w64-cross-clang/src/libcxx/src/include/from_chars_floating_point.h:13:10: fatal error: 'shared/fp_bits.h' file not found
   13 | #include "shared/fp_bits.h"
```
thinking I need to give in and use the full llvm-project src.tar.xz

jeremyd2019:
Also getting
```
CMake Error at /d/S/mingw-w64-cross-clang/src/llvm-project-20.1.2.src/libunwind/src/CMakeLists.txt:107 (message):
  Compiler doesn't support generation of unwind tables if exception support
  is disabled.  Building libunwind DSO with runtime dependency on C++ ABI
  library is not supported.
```

it appears the cmake check for the `-fno-exceptions` flag is trying to link to libc++ that isn't built yet

llvm/0202-hack-add-system-search-paths.patch
```diff
@@ -0,0 +1,27 @@
...
+--- a/lld/MinGW/Driver.cpp
++++ b/lld/MinGW/Driver.cpp
+@@ -546,6 +546,10 @@ bool link(ArrayRef<const char *> argsArr, llvm::raw_ostream &stdoutOS,
+     add("-exclude-symbols:" + StringRef(a->getValue()));
+ 
+   std::vector<StringRef> searchPaths;
++  #ifdef __CYGWIN__
++    searchPaths.push_back("/usr/lib");
++    searchPaths.push_back("/usr/lib/w32api");
++  #endif
```
jeremyd2019, line 21: This  breaks cross-compilation.  Shouldn't this be done in the compiler, anyway?

mati865: IIRC, GCC doesn't pass `-L` so without this change `gcc -fuse-ld=lld` won't work.

jeremyd2019: so this is another case where GNU ld hardcodes things because it knows its target at build time, but LLD is left to divine its target from its inputs, and there's no real way to tell the difference between mingw and cygwin from them at this point?

jeremyd2019:
now error on aarch64:
```
        ld.lld: error: unable to automatically import from _fpreset with relocation type IMAGE_REL_ARM64_BRANCH26 in /opt/aarch64-w64-mingw32/lib/crt2.o
        ld.lld: error: unable to automatically import from _fpreset with relocation type IMAGE_REL_ARM64_BRANCH26 in libmingw32.a(libarm64_libmingw32_a-crt_handler.o)
        ld.lld: error: unable to automatically import from _fpreset with relocation type IMAGE_REL_ARM64_BRANCH26 in libmingw32.a(libarm64_libmingw32_a-tlsthrd.o)
        clang: error: linker command failed with exit code 1 (use -v to see invocation)
        ninja: build stopped: subcommand failed.
```

jeremyd2019:
I went with an older revision of mingw-w64 (matching what's used in the mingw-w64-cross-headers package).

llvm/0202-hack-add-system-search-paths.patch
mati865: This needs a double check but I think so.

mstorsjo:
> now error on aarch64:
> 
> ```
>         ld.lld: error: unable to automatically import from _fpreset with relocation type IMAGE_REL_ARM64_BRANCH26 in /opt/aarch64-w64-mingw32/lib/crt2.o
>         ld.lld: error: unable to automatically import from _fpreset with relocation type IMAGE_REL_ARM64_BRANCH26 in libmingw32.a(libarm64_libmingw32_a-crt_handler.o)
>         ld.lld: error: unable to automatically import from _fpreset with relocation type IMAGE_REL_ARM64_BRANCH26 in libmingw32.a(libarm64_libmingw32_a-tlsthrd.o)
>         clang: error: linker command failed with exit code 1 (use -v to see invocation)
>         ninja: build stopped: subcommand failed.
> ```


> I went with an older revision of mingw-w64 (matching what's used in the mingw-w64-cross-headers package).

Yeah this would be caused by a recent change in mingw-w64-headers/crt, you can probably get that if they aren't in sync.

jeremyd2019:
@mati865 does this look reasonable now?  I think most things work:
1. clang can build itself
2. clang can build llvm/clang dylib
3. clang/lld still works to cross compile mingw targets (though I haven't tested that they generate working executables)

What doesn't work
1. lld doesn't work for the native/host/default cygwin target.
I backed out a patch that was trying to help this, but broke lld for mingw cross targets.  I also had more hacks to improve cygwin at the expense of mingw, but even then there were still issues using it to build llvm/clang dylib, so I abandoned that idea and decided it was better to support mingw cross since we can just use GNU ld for cygwin.

mati865:
@jeremyd2019 thank you for basically taking over. Yes, that looks good to me.

LLD fixes were half-baked anyway. I think this is basically not fixable without introducing a new `-m` variant or prefixing LLD binary with the target triple. Neither feels like a nice solution.

jeremyd2019:
OK, let's talk dependencies.  llvm doesn't depend on anything?  And I think *something* (probably clang) should depend on compiler-rt.

jeremyd2019:
I successfully built llvm with msys2-runtime 3.6.1.  I wanted to note this because it failed to build with whatever test version of 3.6.0 I had installed (and I went back to the released 3.5 version rather than figure out if my 3.6 was out of date or not).

mati865:
Optdepends sounds good.

PR #5317, opened 2025-04-10
openssh: Update to 10.0p1

lazka:


Biswa96:
[openssh-10.0p2](https://github.com/openssh/openssh-portable/releases/tag/V_10_0_P2) (identical to 10.0p1)

lazka:
thanks, there is no release tarball though

lazka:
hm, or should we fake the version number of the package to match the reported version?

PR #5321, opened 2025-04-11
msys2-runtime: update patches

jeremyd2019:
* msys2/msys2-runtime#278
* msys2/msys2-runtime#279


jeremyd2019:
probably wait until the massive rebuild is done, msys2-runtime would implicitly be a dependency of everything, so would stop it in its tracks until built.

lazka:
Any objections to releasing it from staging after this?

jeremyd2019:
The ctrl-c thing was all I noticed.  It seems to be doing well for building lots of packages so far (it survived building llvm from MSYS2-packages as well as the llvm update and lots of rebuilds ongoing for MINGW-packages).

PR #5322, opened 2025-04-11
openssh: Update to 10.0p1 (v2)

lazka:
I had to pull the last update, f97d8e521886b021d868 because it was no longer working with cygwin 3.5

### DIFF
--- a/openssh/PKGBUILD
+++ b/openssh/PKGBUILD
@@ -1,8 +1,8 @@
 # Maintainer: Alexey Pavlov <alexpux@gmail.com>
 
 pkgname=openssh
-pkgver=9.9p2
-pkgrel=1
+pkgver=10.0p1
+pkgrel=2
 pkgdesc='Free version of the SSH connectivity tools'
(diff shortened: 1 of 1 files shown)


jeremyd2019:
Just a hazard of building MSYS2-packages against an msys2-runtime (major) version in staging that's not yet available to everyone.  Similarly a hazard of downgrading an msys2-runtime (major) version

lazka:
Yeah, I usually don't push cygwin package updates out of order, except for msys2-runtime to get more build testing there (which has been very helpful for 3.6)

jeremyd2019:
I wonder if you could hack autobuild to not add msys2-runtime-devel updates to the local repo.  That way it would still have the live headers and import library, so configure et al wouldn't see new APIs

lazka:
hm, yeah. If we remove the strict version dependency between them I could just remove the -devel package from the PKGBUILD so it's not built until we decide to commit.

lazka:
I'll try the no-headers approach the next time.

PR #5327, opened 2025-04-12
llvm: remove unnecessary MINGW-packages detritus

jeremyd2019:
I don't know if this warrants a pkgrel bump.  I think the only real change would be the LLVMConfig.cmake file.


Biswa96:
I thought detritus is a package name.

jeremyd2019:
detritus is basically a real-word synonym for "cruft".

PR #5359, opened 2025-04-27
cross-mingwarm64-gcc: Update

lazka:
the upstream fork now contains the cygwin patches, so uncomment them for now.

enable libstdcxx-backtrace to match the other gcc cross compilers


lazka:
thanks for the hint

PR #5368, opened 2025-05-02
llvm: update to 20.1.4

jeremyd2019:
update patches from upstream

Making some progress on upstreaming things.  I tried using github commit
 .patch URLs but it gave me 429 errors, I guess GitHub doesn't appreciate downloading a bunch of commits as patches in a loop.


jeremyd2019:
I'm considering making a README.md like we had with mingw-w64-clang for keeping track of all these patches and their upstreaming state.

PR #5370, opened 2025-05-04
emacs: Update to 30.1

lazka:


jeremyd2019:
```
configure: error: Native compilation is not supported on 32-bit Cygwin.
If you really want to try it anyway, use the configure option
'--with-cygwin32-native-compilation'.
```

Not sure what they want here - turn off native compilation or add this option?

lazka:
> Not sure what they want here - turn off native compilation or add this option?

I guess turning it off would be safer but slower (?)

jeremyd2019:
I don't think native compilation isn't on on 64-bit either, see check output:
```
configure: WARNING: Elisp native compiler can't be enabled as libgccjit was not
  found.
  Please try installing libgccjit or a similar package if you want to have it
  enabled.
  checking for libgccjit.h... no
  configure: WARNING: Elisp native compiler can't be enabled as libgccjit header files
  were not found.
  Please try installing libgccjit-dev or a similar package if you want to have it
  enabled.
```

lazka:
hm, ok. then I also don't know what it wants.

jeremyd2019:
It wants libgccjit.  I saw a patch for gcc for libgccjit, but it's not present in the msys gcc package.

PR #5371, opened 2025-05-04
emacs: explicitly disable native compilation

jeremyd2019:
It is implicitly disabled due to configure not finding libgccjit.h, and is not supported (and causes configure to fail) on 32-bit, so just disable it.


lazka:
thanks

PR #5375, opened 2025-05-07
libxml2: Update to 2.14.2

lazka:


lazka:
```
src/lxml/etree.o:etree.c:(.text+0x49f72): undefined reference to `__imp_htmlDefaultSAXHandler'
```

Paebbels:
Is this the cause why `lxml` was updated in MSYS2, but not in MinGW64 and UCRT64?

* #5478

lazka:
the mingw version fails similarly.

lazka:
https://gitlab.gnome.org/GNOME/libxml2/-/issues/961

Issue #5285, opened 2025-03-26
allow for easier 3rd-party addition of MSYSTEMs


jeremyd2019:
in https://github.com/msys2/MINGW-packages/discussions/13259#discussioncomment-12115759, @revelator wrote:
> ok ill see what i can do also need to repack the modified filesystem pacman and msys2-launcher packages otherwise it wont be a part of the toolchain.

This got me to thinking: could we make it easier for 3rd-party packages to drop in additional MSYSTEMs?  Something like:
`/etc/msystem.d/ucrt32`
`/etc/makepkg_mingw.d/ucrt32.conf`

and then have the logic in `/etc/msystem` and `/etc/makepkg_mingw.conf` to either source everything in the directory, or else look for `${MSYSTEM,,}.conf`?  (Keep in mind `/etc/msystem` should probably not have bash-isms like that, so we'd need something POSIXy to do any lower-casing there)

revelator:
that would be an interresting approach to it :) im game.

revelator:
something like a plugin system for toolchains i reckon thats the idea ?. ill see if i can cook something up as it would be an immense help,

jeremyd2019:
I didn't contemplate anything different for pacman.conf.

jeremyd2019:
This idea was implemented in #5381 and #5382

PR #5381, opened 2025-05-09
filesystem: make msystems pluggable.

jeremyd2019:
The /etc/msystem script sources /etc/msystem.d/${MSYSTEM,,} (but using only POSIX shell features rather than a bash-ism to lower-case since /etc/msystem is included in /etc/profile, so may be used by a non-bash shell).  /usr/bin/shell validates an MSYSTEM by checking for the existence of that file.

Addresses #5285 (next up for that issue is pacman's makepkg-mingw and /etc/makepkg_mingw.conf)

Not handled is msys2_shell.cmd.  It should be possible though.


MehdiChinoune:
Still don't understand that extra `msystem.d` prefix!

filesystem/msystem
```diff
@@ -20,56 +20,10 @@ unset MINGW_CHOST
...
-        export MSYSTEM_PREFIX MSYSTEM_CARCH MSYSTEM_CHOST MINGW_CHOST MINGW_PREFIX MINGW_PACKAGE_PREFIX
-        ;;
-    *)
-        MSYSTEM_PREFIX='/usr'
-        MSYSTEM_CARCH="$(/usr/bin/uname -m)"
-        MSYSTEM_CHOST="${MSYSTEM_CARCH}-pc-msys"
-        export MSYSTEM_PREFIX MSYSTEM_CARCH MSYSTEM_CHOST
-        ;;
-esac
+_msystem_d="/etc/msystem.d/$(echo "${MSYSTEM}" | LC_ALL=C /usr/bin/tr A-Z a-z)"
```
lazka, line 23: This creates 3 new processes on every shell startup, no? Can we optimize this?

jeremyd2019: I don't know.  I did it that way because msystem is sourced from profile, and that was the way I found that would work on any POSIX shell.  I just found another way that seems like it would work without subprocesses - `typeset -l`.  But it is also not strictly POSIX, though it should work with bash and ksh so it's probably reasonably portable.

lazka: hm ok. We already have various shell specific things in profile, so that is always an option too.

What if we just uppercase the filenames?

jeremyd2019: > What if we just uppercase the filenames?

That'd be the easiest, if we're ok with the "ugliness" of the uppercase filenames.  Then `shell` can use a bash-ism to uppercase its parameter, since it already requires bash.  msys2_shell.cmd would already need to uppercase to set MSYSTEM if/when that gets updated.

lazka: hm, yeah.

I measured and this line takes ~30ms here.

jeremyd2019: uppercased the filenames

lazka:
I've created https://github.com/msys2/msys2-tests/pull/89 for this

PR #5413, opened 2025-05-21
makepkg confs: apply +crt-static in RUSTFLAGS for CLANG*

ognevny:
currently there is an issue with `crt-static` option being uncontrollable for windows-gnu targets: https://github.com/rust-lang/rust/issues/137389. this behavour is bad because libunwind.dll is linked by default for windows-gnullvm, but no gcc dynamic libs are linked in case of windows-gnu. I tried to use bootstrap.toml option, but it doesn't do what is expected. this leads to confusion because there is a need to pull libunwind package separately for all rust-based packages...

this change should be temp before the issue is fixed. this flag can be changed in case of package which has cc-libs as their dependency anyway. didn't add the flag for GCC envs because nothing changes with it

I'll block some rust updates/additions before this is merged

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=pacman
 pkgver=6.1.0
-pkgrel=13
+pkgrel=14
 pkgdesc="A library-based package manager with dependency support (MSYS2 port)"
 arch=('i686' 'x86_64')
 url="https://www.archlinux.org/pacman/"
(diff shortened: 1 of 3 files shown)


ognevny:
@jeremyd2019 can you look at it, please? 🙏

jeremyd2019:
I don't know about the rust flags themselves, but the change makes sense structurally

ognevny:
can you merge it, please?

jeremyd2019:
Sorry, I thought you had rights to merge

PR #5415, opened 2025-05-21
Duplicate libmsys-2.0 static/import library as libcygwin

mati865:
A step towards #3012


jeremyd2019:
Is there a particular use case where just making libcygwin.a contain
```
INPUT(libmsys-2.0.a)
```
wouldn't be sufficient?

mati865:
Duplicated library, unlike the linker script, works with LLD.

PR #5418, opened 2025-05-25
doxygen: Update to 1.14.0

ReenigneArcher:


ReenigneArcher:
Did the release fail? https://packages.msys2.org/base/doxygen is still showing 1.13.2.

PR #5419, opened 2025-05-26
grep: Update to 3.12

Suyun114:


jeremyd2019:
the last time we tried to update grep it broke libtool, I think because grep upstream decided it'd be fun to change how crlf line endings were being handled.

lazka:
It also resulted in gcc exception handling being broken (when built with newer grep, we never debugged why)

Some users complained that it broke their workflow, but most of it can be worked around by piping through dos2unix, or disabling autocrlf in git.

Not sure what we should do about it.

PR #5420, opened 2025-05-26
crosstool-ng: Update to 1.27.0

Suyun114:


lazka:
Just wondering, why was the patch dropped?

lazka:
Since I think you asked in chat: I tested things, and yes, the patch is still needed from what I see, and `os=-mingw32` should be `os=mingw32`

PR #5437, opened 2025-06-05
llvm: update to 20.1.6

jeremyd2019:


lazka:
(feel free to merge any time)

jeremyd2019:
yeah, I've started trying to fix some of the test failures, so now I have more patches pending :wink:

jeremyd2019:
This should really have a double-rebuild, since we use clang to build clang and there's an ABI change in here (both fixing `va_list` and `wint_t`)... But there are still more patches pending so adding them can be the second rebuild :grin:

Issue #5372, opened 2025-05-05
Failure buildin Emacs [MSYS2]


angelog0:
### Description / Steps to reproduce the issue

Current Emacs master fails to build for MSYS2:

```
[...]
Configured for 'x86_64-pc-cygwin'.

  Where should the build process find the source code?    .
  What compiler should emacs be built with?               gcc -O2
  Should Emacs use the GNU version of malloc?             no
    (The GNU allocators don't work with this system configuration.)
  Should Emacs use a relocating allocator for buffers?    no
  Should Emacs use mmap(2) for buffer allocation?         no
  What window system should Emacs use?                    none
  What toolkit should Emacs use?                          none
  Where do we find X Windows header files?                NONE
  Where do we find X Windows libraries?                   NONE
  Does Emacs use -lXaw3d?                                 no
  Is Emacs being built for Android?                       no
  Does Emacs use the X Double Buffer Extension?           no
  Does Emacs use -lXpm?                                   no
  Does Emacs use -ljpeg?                                  no
  Does Emacs use -ltiff?                                  no
  Does Emacs use a gif library?                           no
  Does Emacs use a png library?                           no
  Does Emacs use -lrsvg-2?                                no
  Does Emacs use -lwebp?                                  no
  Does Emacs use -lsqlite3?                               no
  Does Emacs use cairo?                                   no
  Does Emacs use -llcms2?                                 no
  Does Emacs use imagemagick?                             no
  Does Emacs use native APIs for images?                  no
  Does Emacs support sound?                               yes
  Does Emacs use -lgpm?                                   no
  Does Emacs use -ldbus?                                  no
  Does Emacs use -lgconf?                                 no
  Does Emacs use GSettings?                               no
  Does Emacs use a file notification library?             no
  Does Emacs use access control lists?                    yes
  Does Emacs use -lselinux?                               no
  Does Emacs use -lgnutls?                                yes
  Does Emacs use -lxml2?                                  yes
  Does Emacs use -lfreetype?                              no
  Does Emacs use HarfBuzz?                                no
  Does Emacs use -lm17n-flt?                              no
  Does Emacs use -lotf?                                   no
  Does Emacs use -lxft?                                   no
  Does Emacs use -lsystemd?                               no
  Does Emacs use -ltree-sitter?                           no
  Does Emacs use the GMP library?                         yes
  Does Emacs directly use zlib?                           yes
  Does Emacs have dynamic modules support?                yes
[...]

PR #5450, opened 2025-06-11
gcc: Update to 13.4.0

lazka:
the removed change was fixed upstream via
https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=1283b9f946eea07573be5ba0e0785c9e9279b3be


jeremyd2019:
Assuming the analysis by the gnulib folks was correct (and I'm pretty sure it was), this should fix the error building emacs from master branch in #5372 @angelog0

angelog0:
> this should fix the error building emacs from master branch in https://github.com/msys2/MSYS2-packages/issues/5372 @angelog0

Indeed, confirmed. Thanks!

PR #5458, opened 2025-06-18
gcc: update to 15.1.0

MehdiChinoune:
patches copied from Cygwin.


lazka:
I guess we could. But it might also mean more work being the first to update; cygwin is still on gcc 12.

With aarch64 support progressing it might help there though.

jeremyd2019:
> cygwin is still on gcc 12.

They're planning to update to gcc 13.4.0 real soon now.

MehdiChinoune:
We were ahead of CygWin for many years, I think since gcc11, they were always behind us by one or two versions.

lazka:
At that (v11) time they had no one updating their cross compiler afair, but yes, they are always a bit behind on purpose, for some reason.

OK, I propose to wait a bit in case there is any fallout from the triplet change, and update to 15 next week.

We also need to backport https://cygwin.com/git/?p=newlib-cygwin.git;a=commit;h=83af2fd6d1018bfe1c08c0028f65076f57684474 in msys2-runtime

lazka:
Building ninja fails with:



```

2025-06-26T10:27:19.0541758Z bootstrapping ninja...
2025-06-26T10:27:57.5829820Z /usr/lib/gcc/x86_64-pc-cygwin/15.1.0/../../../../x86_64-pc-cygwin/bin/ld: cannot find -lintl: No such file or directory
2025-06-26T10:27:57.6023521Z /usr/lib/gcc/x86_64-pc-cygwin/15.1.0/../../../../x86_64-pc-cygwin/bin/ld: cannot find -liconv: No such file or directory
2025-06-26T10:27:57.6500627Z collect2: error: ld returned 1 exit status
2025-06-26T10:27:57.6521311Z when running:  g++ -Lbuild -march=nocona -msahf -mtune=generic -O2 -pipe -march=nocona -msahf -mtune=generic -O2 -pipe -D_XOPEN_SOURCE=700  -o ninja build/ninja.o -lninja
```

I guess due to 0301-libstdc-Implement-most-of-locale-features-for-newlib.patch

lazka:
Should we add a runtime dep on libiconv-devel and gettext-devel, or drop the patch for now?

I tend towards dropping it for now.

MehdiChinoune:
I am with dropping

lazka:
OK, I'll push something

lazka:
Looks good now

PR #5478, opened 2025-06-26
python-lxml: Update to 6.0.0

lazka:


Paebbels:
I see MSYS2 was updated, but MinGW64 and UCRT64 didn't receive updates.

Was this forgotten or is there something I can try to help with?

lazka:
I wanted to update cython first, but that broke various things. But let me try.

PR #5503, opened 2025-07-09
gnutls: Update to 3.8.10

lazka:


jeremyd2019:
huh, that wint_t thing was supposed to be fixed by the update to gcc 13.4.0.  Did the update to gcc 15 regress that fix?

lazka:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114870 right? Not sure

lazka:
gnutls-3.8.9 also fails the same way

jeremyd2019:
> https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114870 right? Not sure

yeah, but really gnulib's attempt to work around it IMHO.  But that shouldn't affect gcc 15.  Maybe this has an older version of gnulib with an even more broken attempted workaround? (https://github.com/coreutils/gnulib/commit/e529f4fbd08fa9e60c7092ba0f8cc976a735077c ?)

jeremyd2019:
Yep,
```
  checking whether nullptr_t needs <stddef.h>... yes
  checking for clean definition of __STDC_VERSION_STDDEF_H__... no
```

Could probably override it with `gl_cv_clean_version_stddef=yes` or patch gnulib stddef_h.m4...

lazka:
thanks, I'll try `gl_cv_clean_version_stddef=yes`

lazka:
thanks again

Issue #2570, opened 2021-07-03
no msys version of jq package


igagis:
## Describe the issue

There is a `mingw-w64-x86_64-jq` package for `mingw64` env, but there is no `jq` package for `msys` env. Is there a reason for this?

Need to add the `jq` package.

Biswa96:
msys2 packages are just to support build environment for mingw packages.

igagis:
Well, that's exactly why I need the `jq` package, as I use jq utility in my CI process for building my mingw packages.

Biswa96:
Then install mingw-w64-jq package.

igagis:
This way jq is not available in msys environment, and I'm doing mingw package deployment from msys env, not from mingw env, so I need jq in msys env

Biswa96:
> I'm doing mingw package deployment from msys env

Confusing. All the mingw packages should be compiled in mingw environment. For example, see the [main workflow][1] in mingw-packages repository.

> not from mingw env

Why not from mingw env?

[1]: https://github.com/msys2/MINGW-packages/blob/master/.github/workflows/main.yml

igagis:
I build packages in mingw64/32 env, but then I need to deploy those packages to repository and that I do from msys env as in this case I need as close as possible to linux-like environment to execute scripts which also work for linux and macos. Using mingw64 env for this does not look right.

Biswa96:
Shell scripts can be run from any mingw environment. If you provide the build log or the CI yaml file, CI experts can help.

igagis:
I am a CI expert.

It looks really strange that there is a curl package for msys env, but no jq package, so I can make rest requests with curl in msys but cant parse its json response because there is no jq...

I don't get your concepts on this.

oech3:
How about symlinking native `jq` to your `/usr/local/bin/jq.exe` and removing `jq` from MSYS2's repository?
I have no idea why this is needed even it is not a library package.

igagis:
it is needed because `msys` provides linux-like environment, that is it's purpose. `jq` is part of the linux-like environment.

oech3:
Sorry. I don't understand. You can just symlink or alias native one.

igagis:
what is the package manager for then? I don't want to symlink anything by hand in system path's.
If you propose me a workaround, then I have an idea how to do that. This issue is not about that.

oech3:
[...]

PR #5526, opened 2025-07-22
jq: New Package

bwijen:
Port jq (and dependency oniguruma) from MINGW:

Solves: #2570


lazka:
Do we really need to make it look for Windows USERPROFILE? Our cygwin tools usually don't do that, see https://www.msys2.org/docs/configuration/

other than that I'm fine with adding it.

bwijen:
> Do we really need to make it look for Windows USERPROFILE? Our cygwin tools usually don't do that, see https://www.msys2.org/docs/configuration/
> 
> other than that I'm fine with adding it.

Indeed we shouldn't, we have `db_home` for that.
I've taken the actual commit from the jq devs and reverse applied it.
See: 
https://github.com/jqlang/jq/pull/3114
https://github.com/bwijen/MSYS2-packages/blob/485eec87/jq/PKGBUILD#L22-L23

lazka:
ah, revert. ok.

lazka:
I still don't quite understand, the upstream code looks fine as is. What are you trying to change with the revert?

jeremyd2019:
When I read it, it did look like the code was fine, it was just the test that was broken (it assumed running in an msys shell meant it was running a win32 binary)

lazka:
ok, thanks, avoided by changing uname to Cygwin

lazka:
thanks!

PR #5536, opened 2025-07-26
fish: use rustc completions from v4.0

ognevny:
they were broken in v3.x, and it was fixed only in 4.0 release

```shell
$ rustc +string split: - hashes of green query instances: unknown option

/usr/share/fish/completions/rustc.fish (line 1):
string split -m 1 ' ' $line
^
in command substitution
        called on line 43 of file /usr/share/fish/completions/rustc.fish
from sourcing file /usr/share/fish/completions/rustc.fish

(Type 'help string' for related documentation)
string split: - hash collisions of query keys: unknown option

/usr/share/fish/completions/rustc.fish (line 1):
string split -m 1 ' ' $line
^
in command substitution
        called on line 43 of file /usr/share/fish/completions/rustc.fish
from sourcing file /usr/share/fish/completions/rustc.fish

(Type 'help string' for related documentation)
string split: - hash collisions when creating dep-nodes: unknown option

/usr/share/fish/completions/rustc.fish (line 1):
string split -m 1 ' ' $line
^
in command substitution
        called on line 43 of file /usr/share/fish/completions/rustc.fish
from sourcing file /usr/share/fish/completions/rustc.fish

(Type 'help string' for related documentation)
```
(if you press Tab after typing '+')

### DIFF
--- a/fish/PKGBUILD
+++ b/fish/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=fish
 pkgver=3.7.1
-pkgrel=5
+pkgrel=6
 epoch=
 pkgdesc='Smart and user friendly shell intended mostly for interactive use'
 arch=('i686' 'x86_64')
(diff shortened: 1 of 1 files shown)


ognevny:
@lazka it may make sense to separately provide (or download) fish completions, so we can update them all to modern revision

lazka:
I'd prefer to keep it simple as here and backport into this package. (do we have to care if new completions require new fish? something to consider at least)

ognevny:
> (do we have to care if new completions require new fish? something to consider at least)

I guess no for rustc, unsure about other completions...

PR #5539, opened 2025-07-27
cross-mingwarm64-binutils: Update to 2.44dev

lazka:


lazka:
I guess we need some newer binutils fixes. Maybe we should just take the patches on top of the current release instead of taking them from git.

edit: https://github.com/bminor/binutils-gdb/commit/5c87b330e910f8be1443c881fd16a70e685f1f2f

PR #5542, opened 2025-07-28
pax: add msys2_references

striezel:


striezel:
Neither the old nor the newer version builds, and I currently do not have the time to look into this (there are quite a few errors), so I'll close it.

lazka:
I've pushed the first commit alone

striezel:
> I've pushed the first commit alone

Thanks. :+1: Just having the first commit was my original intention anyway.
But after I saw that the build failed I thought that maybe the newer version would not have such build problems and gave it a try. Well, I was wrong, there were problems, too.

PR #5552, opened 2025-08-05
perl: Update to 5.40.3

lazka:


lazka:
Not knowing perl very well, I was wondering if Term-Table and Test2-Suite are now in core only, but it seems users can decide which version to use, so the separate packages for them are fine. Let me know if I got that wrong.

lazka:
test suite gives the same results (i.e. same tests are failing as currently)

PR #5571, opened 2025-08-18
[new package] rust 1.89.0

ognevny:
- use Jeremy's build as a stage0 compiler
- based on MinGW package but with many things removed (e.g. wasm targets, src component, PGO...)

closes https://github.com/msys2/MINGW-packages/pull/23936

### DIFF
--- a/rust/PKGBUILD
+++ b/rust/PKGBUILD
@@ -1,0 +1,120 @@
+# Maintainer: Maksim Bondarenkov <maksapple2306@gmail.com>
+
+_bootstrapping=no
+
+rust_dist_server=https://static.rust-lang.org/dist
+#rust_dist_server=https://dev-static.rust-lang.org/dist/2025-08-04
+
+pkgname=rust
+pkgver=1.89.0
+pkgrel=1
+pkgdesc="Systems programming language focused on safety, speed and concurrency"
+arch=('x86_64')
+url="https://www.rust-lang.org/"
+msys2_repository_url="https://github.com/rust-lang/rust"
+msys2_references=(
+  'anitya: 7635'
+  'archlinux: rust'
+  "cpe: cpe:/a:rust-lang:rust"
+  'gentoo: dev-lang/rust'
+)
+license=('spdx:Apache-2.0 AND MIT')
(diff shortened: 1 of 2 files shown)


ognevny:
thanks @jeremyd2019 for a review. I'll push changes after local testing

ognevny:
hm?
```
ERROR: creating symbolic link `/home/maksa/forks/MSYS2-packages/rust/src/rustc-1.89.0-src/build/x86_64-pc-cygwin/stage2/lib/rustlib/src/rust` to `/home/maksa/forks/MSYS2-packages/rust/src/rustc-1.89.0-src` failed with Too many levels of symbolic links (os error 92)
Build completed unsuccessfully in 0:00:44
```

jeremyd2019:
I think I've seen that before but I don't remember what I did about it.

The issue is with deepcopy symlinks it obviously can't do that, so it fails with ELOOP.  Maybe put the build directory outside the rustc src directory?

ognevny:
> Maybe put the build directory outside the rustc src directory?

don't want to do out of tree builds... so I'll try to change winsymlinks option

jeremyd2019:
> > Maybe put the build directory outside the rustc src directory?
> 
> don't want to do out of tree builds... so I'll try to change winsymlinks option

be careful with that, symlinks are not allowed in packages so make sure they don't end up in the pkgdir accidentally

ognevny:
...and sometimes build just hangs 😄

jeremyd2019:
probably on fork, due to address space conflicts (usually it'll fail with an error but sometimes just hangs).  you could try to rebase but msys2 doesn't rebase on x86_64 generally so there won't be a rebase db.  https://www.cygwin.com/cgit/newlib-cygwin/log/?h=topic/posix_spawn is my attempt to avoid this, but not much help to you right now.

ognevny:
ready :)

ognevny:
I've succeeded to build fish with https://github.com/Berrysoft/fish-msys2 recipe, so we can move forward, I guess...

ognevny:
forgot to check earlier, there is no symlink in the package

PR #5572, opened 2025-08-18
fish: update to 4.0.2

ognevny:
based on https://github.com/Berrysoft/fish-msys2

- use Ninja
- change Cargo deps in a way so the build is reproducible

closes https://github.com/msys2/MINGW-packages/pull/22837

### DIFF
--- a/fish/0001-support-cygwin.patch
+++ b/fish/0001-support-cygwin.patch
@@ -1,0 +1,207 @@
+diff --git a/Cargo.toml b/Cargo.toml
+index d592a1aaf2d9..c942fcaee91b 100644
+--- a/Cargo.toml
++++ b/Cargo.toml
+@@ -71,6 +71,9 @@ serial_test = { version = "3", default-features = false }
+ cc = "1.0.94"
+ rsconf = "0.2.2"
+ 
++[target.'cfg(windows)'.build-dependencies]
++unix_path = "1.0.1"
++
+ [lib]
+ crate-type = ["rlib"]
+ path = "src/lib.rs"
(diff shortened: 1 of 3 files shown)


ognevny:
I can't undersand this error: `Error copying file "/d/_/fish/src/build/cargo/build/x86_64-pc-cygwin/release/fish" to "/d/_/fish/src/build".`

update: I somehow didn't catch the thing that CMake profile MUST be `Release`, otherwise wrong Cargo profile is used with `None`

ognevny:
thanks @Berrysoft and @jeremyd2019 for making this happen!

ahaoboy:
I've been using it for weeks, and noticed two issues.

1. Fish freezes for a few seconds while typing, then returns to normal. I haven't found a specific way to reproduce this issue.
2. VSCode can't split terminals. https://github.com/microsoft/vscode/issues/257263

Overall, Fish is the best shell experience on Windows. Thank you to all the contributors.

Berrysoft:
Hello, @ahaoboy . I haven't met the first issue so you might need to explain with more detail to reproduce that. For the second issue, I *strongly* suggest you reopen the issue in the vscode repo. That's the correct place, I think.

ahaoboy:
> Hello, @ahaoboy . I haven't met the first issue so you might need to explain with more detail to reproduce that. For the second issue, I _strongly_ suggest you reopen the issue in the vscode repo. That's the correct place, I think.

I haven't found a stable way to reproduce this issue yet. It occasionally freezes for 3-5 seconds while typing, with no error message.

I don't have permission to open vscode issues :(

PR #5574, opened 2025-08-18
rust: rebuild with itself

ognevny:
so we use our package as a stage0


MehdiChinoune:
You can't merge it until rust is deployed the stable repo.

ognevny:
oh, thanks. didn't thought about that

lazka:
you can now

PR #5586, opened 2025-08-22
maturin: Add 1.9.6

lazka:
Doesn't build, feel free to improve it. (I'd assume it needs rust ecosystem fixes, so not easy)

Disabling rusttls makes it fail in https://github.com/bytecodealliance/target-lexicon I think

### DIFF
--- a/maturin/PKGBUILD
+++ b/maturin/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgbase=maturin
 pkgname=("python-maturin" "maturin")
-pkgver=1.9.4
+pkgver=1.9.6
 pkgrel=1
 pkgdesc='Build and publish crates with pyo3, rust-cpython and cffi bindings'
 arch=('x86_64')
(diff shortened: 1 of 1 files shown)


ognevny:
some low-level rust packages require changes to make work on cygwin. in most cases it's adding some cfg's (like `#if #endif`) or even link specific Cygwin API...

edit: compile error output literally tells to add some extra cfg conditions

ognevny:
https://github.com/bytecodealliance/rustix/pull/1410
https://github.com/bytecodealliance/rustix/pull/1428

^ backport these + ensure that some deps are up-to-date, then try again...

lazka:
thanks.

I also got these some time (this log is from testing with another package):

```
error: failed to run custom build command for `target-lexicon v0.13.2`

Caused by:
  process didn't exit successfully: `/tmp/build-via-sdist-umdsri5d/bcrypt-4.3.0/src/_bcrypt/target/release/build/target-lexicon-567ae6ae729fbcec/build-script-build` (exit status: 101)
  --- stderr

  thread 'main' panicked at /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/target-lexicon-0.13.2/build.rs:52:54:
  Invalid target name: 'x86_64-pc-cygwin'
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

ognevny:
I've just looked at source. seems easy to fix

ognevny:
that's how far I get. the problem is that `getrandom` v0.3.2 is required, but reverse dependency is abandoned so it's stuck at 0.2.x

ognevny:
rustls is still compiled even when it's disabled

lazka:
thanks, blocked on https://github.com/PyO3/maturin/issues/1858 then I guess

PR #5598, opened 2025-08-26
llvm: update to 21.1.0

### LINKED ISSUE #5573: split llvm-libs from llvm package
so Rust installation won't take so much memory (only shared libraries are required at runtime)

jeremyd2019:
add an msysize patch to deal with the fallout of teaching the Cygwin driver how to call the linker directly instead of using GCC to do it

Split llvm-libs package fixes #5573


jeremyd2019:
any ideas on how to best rebuild rust for new llvm (assuming 1.89 even builds against llvm 21.1)?

jeremyd2019:
Ideas:
1. merge this, I build new dist tarballs against staging and we build rust over from those
2. Copy /usr/bin/msys-LLVM-20.dll into the llvm-libs package temporarily, rebuild rust, rebuild llvm without msys-LLVM-20.dll

I don't really want to have to build llvm twice to get the msys-LLVM-20.dll, so I'd probably just copy the one installed by the existing llvm package.

jeremyd2019:
that works for llvm, but getting the fork error rebuilding rust.

jeremyd2019:
Build successfully locally.  Maybe try in autobuild and see if it works or not?

jeremyd2019:
rust rebuilt in autobuild successfully.  llvm rebuild without old dll being packaged is running.  🚀

PR #5600, opened 2025-08-27
patchutils: Update to 0.4.3

jannick0:
@lazka The package cleanly compiles locally including all tests - with `--skippgpcheck`. However, the CI build currently fails because of an unknown PGP signature.

Is the package author's public key to be added to the PGP database? I cannot do that for CI supposedly.

~~~
==> Verifying source file signatures with gpg...
      patchutils-0.4.3.tar.xz ... FAILED (unknown public key B7C20D079491EA63)
  ==> ERROR: One or more PGP signatures could not be verified!
~~~

NB: The added patch is PRed upstream on top of version 0.4.3, cf. twaugh/patchutils#98.


lazka:
you have to add a validpgpkeys in the PKGBUILD containing the full key ID

PR #5605, opened 2025-08-29
pacman: restore COMPRESSZST defaults

lazka:
With https://github.com/msys2/msys2-autobuild/commit/dc632d99340da8b204fa2d7 we can now override COMPRESSZST in autobuild, to use higher than default compression when building packages for the repo.

This means we can revert the default COMPRESSZST back to the default zstd compression level (3) and make packaging faster for development/CI etc.


lazka:
@jeremyd2019 fyi, in case that affects your builds

jeremyd2019:
:shrug: I'm only going to be building 32-bit packages for another couple of months, and lower settings would probably help not run out of memory compressing anyway (on my physical machine I have to turn down the number of threads due to too many cores)

PR #5623, opened 2025-09-10
llvm: update to 21.1.1

jeremyd2019:
backport LLD patch for wrapped dllimported weak symbols (used by Cygwin for C++ operator new/delete)


jeremyd2019:
From my email to cygwin-announce regarding lld:

This release includes a backported patch that makes it *possible* to link
Cygwin binaries using clang -fuse-ld=lld.  The defaults are still biased
towards MinGW so using gcc -fuse-ld=lld requires more flags [1].
--enable-auto-image-base is not implemented, and there are known issues
using rebase on lld-generated binaries, so dynamicbase is
currently required for dlls.

1. at least
`-L/usr/w32api/lib -Wl,--disable-high-entropy-va,--disable-nxcompat,--dll-search-prefix=cyg`

(I forgot to mention `-Wl,-e,_cygwin_dll_entry` for linking dlls - and for msys they'd need to be `--dll-search-prefix=msys-` and `-e,_msys_dll_entry`)

PR #5626, opened 2025-09-11
mingw-w64-cross-compiler-rt/cross-clang: update to 21.1.1

jeremyd2019:
Deal with llvm package split, add third-party source tarball for siphash/SipHash.h in aarch64 compiler-rt


jeremyd2019:
I forgot to update these for llvm 21

PR #5630, opened 2025-09-12
llvm: backport fixes, fix 32-bit

jeremyd2019:
* --disable-high-entropy-va is not accepted on 32-bit ld, it should be added only on 64-bit.
* New Cygwin linker driver was not passing --large-address-aware on 32-bit
* Microsoft-style calling convention name mangling (such as stdcall Name@N) was disabled on Cygwin, regression from 21.1.  This would have been an issue on x86_64 as well with vectorcall I think

### DIFF
--- a/llvm/0008-LLVM-Coverage-Unittest-Fix-dangling-reference-in-uni.patch
+++ b/llvm/0008-LLVM-Coverage-Unittest-Fix-dangling-reference-in-uni.patch
@@ -1,0 +1,73 @@
+From 03731bf1dc4a9ab97a03bf4a5d6e33487ccb6845 Mon Sep 17 00:00:00 2001
+From: Tomohiro Kashiwada <kikairoya@gmail.com>
+Date: Fri, 12 Sep 2025 05:12:54 +0900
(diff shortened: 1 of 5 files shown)


jeremyd2019:
I'm still having issues with i686, running clang 21 built with clang 21 gives errors in forked child that cannot allocate heap with win32 error 487 (attempt to access invalid address).  We just found an apparent regression for 32-bit Cygwin targets in clang 21, but not sure how that would be it.

jeremyd2019:
Not applying the 0104 (and 0199) patches seems to fix the issue I'm seeing on i686, so that helps narrow it down some.  It also only happens when building clang with the affected clang while it's building the compiler-rt builtins (which uses the just-built clang).

jeremyd2019:
Appears to have been lack of `--large-address-aware` flag

jeremyd2019:
I'll leave this open for a bit because there are a couple more upstream pull requests that should be able to be merged soon, and I'd include them in this update too.  One appears to be a regression from 20.1.x.

PR #5633, opened 2025-09-13
global: update to 6.6.14

striezel:


striezel:
[Fails](https://github.com/msys2/MSYS2-packages/actions/runs/17703152515/job/50311308915?pr=5633#step:9:586) to compile:
```
   getopt.c: In function '_getopt_initialize':
  getopt.c:287:28: error: too many arguments to function 'getenv'; expected 0, have 1
    287 |   d->__posixly_correct = !!getenv ("POSIXLY_CORRECT");
        |                            ^~~~~~  ~~~~~~~~~~~~~~~~~
  getopt.c:150:14: note: declared here
    150 | extern char *getenv ();
        |              ^~~~~~
  getopt.c: At top level:
  getopt.c:1136:1: error: conflicting types for 'getopt'; have 'int(int,  char * const*, const char *)'
   1136 | getopt (int argc, char *const *argv, const char *optstring)
        | ^~~~~~
  regex.c: In function 're_set_syntax':
  In file included from getopt.c:102:
  getopt.h:155:12: note: previous declaration of 'getopt' with type 'int(void)'
    155 | extern int getopt ();
        |            ^~~~~~
```
Code seems to redefine `getenv()`, so it's should probably be removed / patched out.

piotrvv:
Use gcc 13.4.0-2.
[global-6.6.14-1-x86_64-build.log](https://github.com/user-attachments/files/23292158/global-6.6.14-1-x86_64-build.log)

striezel:
> Use gcc 13.4.0-2.

That's not really an option, because we currently have GCC 15.2.0 in MSYS2.
However, it could possibly still build, if we forced C17 instead of C23 as the default C standard during the build. GCC 15 uses C23 as the default standard version, and that can sometimes cause problems for older code bases.

jeremyd2019:
also if memory serves I believe those getopt files are copied around, and upstream probably has a new version of them.

PR #5638, opened 2025-09-15
nano-syntax-highlighting: update to 2025.07.01

ognevny:
mostly taken from Arch:

- use a fork with the patch applied
- generate nanorc.sample


lazka:
Does the nanorc.sample work for you? it just gives me errors in nano.

ognevny:
> Does the nanorc.sample work for you? it just gives me errors in nano.

hm, didn't check... what's the error? 

update: it's broken in Arch too...

lazka:
It just says " Mistakes in '/home/user/.nanorc' "

the sample file doesn't have any includes in them, so I think the sed is broken somehow

ognevny:
that's because find puts not an absolute path, it's just ${lang}.nanorc. I'll send a fix in the next 5 minutes

PR #5645, opened 2025-09-16
nano-syntax-highlighting: fix nanorc.sample

ognevny:
Arch did generation wrong, and I didn't test it (sorry :( ). `find` returns relative path, which is ${lang}.nanorc, so seding is just simply putting `include ` in the beginning of each line

cc @lazka, thanks for pointing to this!


lazka:
looks good, thanks!

PR #5677, opened 2025-09-28
fish: update to 4.1.0

Berrysoft:
* Remove all patches because v4.1.0 includes them.
* Remove dep on gettext because fish doesn't need it now.


ognevny:
gettext tools seem to be still required

> While we no longer have any runtime dependency on gettext, we still need gettext tools for building, most notably msgfmt. When building without msgfmt available, localization will not work with the resulting executable

Berrysoft:
Oh, yes. My fault.

Berrysoft:
I have created a PR #5679 to fix it.

I also checked the package from CI of this PR. It seems that the localization feature is enabled. `gettext` might have been pulled in by other packages in CI, I think.

ognevny:
> I also checked the package from CI of this PR. It seems that the localization feature is enabled. `gettext` might have been pulled in by other packages in CI, I think.

it's better when explicitly put on the list

PR #5750, opened 2025-10-28
rust: update to 1.91.0

ognevny:
remove backports


ognevny:
it doesn't fail locally, huh

ognevny:
I got this OS error when tried to run tests

jeremyd2019:
try changing the build-dir (maybe remove the `-2` or increment it or something like that).  The image base is based on a hash of the linker's output filename.

PR #5751, opened 2025-10-28
python: add a limited API library

lazka:
Similar to what we do for the native Python create a python3 dll
which just forwards the imports of the limited API subset to
the versioned DLL.

Extend the tool which generates the DLL code for Windows with a
cygwin variant, since we have to update the list on every major
Python release.

### DIFF
--- a/python-cryptography/PKGBUILD
+++ b/python-cryptography/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=cryptography
 pkgname=python-cryptography
 pkgver=46.0.3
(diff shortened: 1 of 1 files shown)


lazka:
I'll have to reach out to cygwin upstream If they want to provide this as well, since this affects how we need to patch python build tools which default to the limited api.

edit: though they can just remove `features = ["abi3"]` from the package they build if wanted. That also fixes the wheel tag which would otherwise point to abi3, while actually not using abi3.

PR #5756, opened 2025-10-29
maturin/python-cryptography: more fixes

lazka:
so we can build cryptography without any hacks


ognevny:
will all these patches be upstreamed?

lazka:
yes, I plan to

PR #5763, opened 2025-11-02
pacman: add RUST_CHOST to makepkg*.conf

### LINKED ISSUE #5762: [pacman] suggestion: introduce RUST_CHOST variable
this could be useful for packages using rustup to build (e.g. cargo-msrv), because it can't print the actual host of our compiler by default (it will be `${CARCH}-pc-windows-msvc`). so putting this variable along with `CHOST` will make scripts cleaner so we don't have to do if branches...

ognevny:
fixes #5762

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -4,7 +4,7 @@
 
 pkgname=pacman
 pkgver=6.1.0
-pkgrel=19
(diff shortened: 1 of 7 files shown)


ognevny:
@lazka that should be enough for me now I guess

PR #5786, opened 2025-11-11
maturin: update to 1.10.1

ognevny:
remove backport; packages are updated upstream

*0003-allow-msys-too.patch is a great name: it can be both "too" and "2"*


ognevny:
full and rustls features now work here...

lazka (approved):

PR #5799, opened 2025-11-17
base: Drop editor

oech3:


lazka:
It's quite handy to have a text editor available by default.

oech3:
How about using optdepends or base-devel?
Arch no longer have editor at base.
In MSYS2, we can use editor installed on Windows.

lazka:
Why do you want to get rid of it?

oech3:
Because I want to save size by default and specific editor in base looks unfair.

lazka:
nano was chosen over others because it is small, most likely familiar to users since it's use in Linux distros, while still being relatively beginner friendly.

It's just 1% of the base size, and getting an editor by default is imo worth it.

PR #5820, opened 2025-11-25
corutils, inetutils: Move hostname

oech3:
Be same with Arch Linux and most distributions.


oech3:
`error: unknown type name 'wint_t'`

PR #5828, opened 2025-11-26
ansible: Update to 13.0.0

lazka:


striezel (approved):
According to the release announcement at https://forum.ansible.com/t/release-announcement-ansible-community-package-13-0-0/44866 ...

> Ansible 13.0.0 depends on ansible-core 2.20.0

... the ansible-core should also be updated. We currently don't have any older ansible-core version in the repo anyway, but let's make sure that everyone updating to ansible 13 gets the proper ansible-core version, too.

lazka:
We don't support partial upgrades, but yeah, can't hurt.

PR #5833, opened 2025-11-28
bash: Update to 5.3.008

lazka:
* Remove all hunks which are included upstream now.
* Remove the hunk for the !BUFFERED_INPUT case, that code path/macro
  has been dropped upstream completely.
* Some minor refresh for the "set_minus_o_option" hunk
* Add "int" to "extern igncr;" due to stricter compiler checks

---

Here is the upstream commit for reference: https://cgit.git.savannah.gnu.org/cgit/bash.git/commit/?id=b8c60bc9ca365f8261fa97900b6fa939f6ebc303


striezel:
Good to see that somebody finally put in the work to get Bash 5.3 to MSYS2. :+1:

PR #5883, opened 2025-12-19
uutils-coreutils: make as a replacement of GNU coreutils

ognevny:
with unix progs added we can finally make this package as a compatible alternative to GNU coreutils

- better reproducibility (properly update deps)
- separate install for more, kill and uptime

cc @oech3

### DIFF
--- a/uutils-coreutils/.gitignore
+++ b/uutils-coreutils/.gitignore
@@ -1,0 +1,3 @@
+/nix
+/ioctl.patch
+/uu-cygwin.patch
--- a/uutils-coreutils/PKGBUILD
+++ b/uutils-coreutils/PKGBUILD
@@ -4,7 +4,7 @@
 _realname=coreutils
 pkgname="uutils-${_realname}"
 pkgver=0.5.0
-pkgrel=2
+pkgrel=3
 pkgdesc="Cross-platform Rust rewrite of the GNU coreutils"
 arch=('any')
 url='https://github.com/uutils/coreutils'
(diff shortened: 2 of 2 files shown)


oech3:
If we used `MULTICALL=y` at pkgbase, `coreutils` binary would be wrong.

oech3:
Oh, we had `uu-`.

ognevny:
> Oh, we had `uu-`.

yes, conflicting binaries have prefixes, so there are coreutils and uu-coreutils

ognevny:
there can be a separate package for conflicting trio

oech3:
Also I don't think ~`kill more uptime`~ `more` has cygpath issue. So we might just delete it.
(No `kill uptime` Windows support yet)

oech3:
`uutils-coreutils-extra` with individual `kill more uptime` binaries?

ognevny:
> `uutils-coreutils-extra` with individual `kill more uptime` binaries?

in our case these utils are not part of 'coreutils', so may be confusing

oech3:
Hmm... Everything in 1 package look OK. It would not be 100 MB+ even 3 utils are individual...

oech3:
I previously opened https://github.com/uutils/coreutils/issues/8975

MehdiChinoune:
As a first step make it a choice for the user by marking it as `conflicts`/`provides` but not `replaces` until making sure it is truly compatible with coreutils on Windows.

ognevny:
> As a first step make it a choice for the user by marking it as `conflicts`/`provides` but not `replaces` until making sure it is truly compatible with coreutils on Windows.

there is no `replaces` here

oech3:
Well... there is a usecase to install this package without replacing system coreutils. But I don't like to build same utils twice to provide both of drop-in and non-drop-in.

ognevny:
> Well... there is a usecase to install this package without replacing system coreutils. But I don't like to build same utils twice to provide both of drop-in and non-drop-in.

probably we don't understand each other... I mean that we make this package as an option of "different" coreutils, but not replacing totally (like in recent Ubuntu version). anyway it will be "hidden" (meaning not shown by pacman when you type `pacman -S coreutils`)

oech3:
I'm aware of this has near property with `zlib-ng-compat`. But someone might still hope installing both of GNU and uutils (with uu-)

ognevny:
> `zlib-ng-compat`. But someone might still hope installing both of GNU and uutils (with uu-)

that fits MINGW-packages more, I guess

oech3:
(I made `AUR/uutils-coreutils-git` drop-in only btw...)

ognevny:
I decided that it's better to provide `more`, `kill` and `uptime` when they become a part of larger packages. also let's try to change build dir to address base image issue in autobuild...

PR #5904, opened 2025-12-28
fish: update to 4.3.1

striezel:
Release tarballs no longer contain the documentation. Thererefore, they are not in the package anymore either. Building documentation requires Sphinx now.

See the paragraph "For distributors and developers" (<https://fishshell.com/docs/current/relnotes.html#for-distributors-and-developers>) in the release notes for Fish 4.3.0.

cc @Berrysoft @ognevny

### DIFF
--- a/fish/PKGBUILD
+++ b/fish/PKGBUILD
@@ -1,7 +1,7 @@
 # Contributor: 王宇逸 <Strawberry_Str@hotmail.com>
 
 pkgname=fish
-pkgver=4.3.0
+pkgver=4.3.1
 pkgrel=1
 pkgdesc='Smart and user friendly shell intended mostly for interactive use'
 arch=('x86_64')
@@ -16,7 +16,7 @@ msys2_references=(
 )
 license=('spdx:GPL-2.0-only')
 depends=('gcc-libs' 'libpcre2_32' 'libpcre2_8' 'man-db')
-makedepends=('gcc' 'gettext' 'pcre2-devel' 'cmake' 'pkgconf' 'rust' 'ninja')
+makedepends=('gcc' 'gettext' 'pcre2-devel' 'cmake' 'pkgconf' 'rust' 'ninja' 'python-sphinx')
 optdepends=('python: for manual page completion parser and web configuration tool')
 install=fish.install
 backup=('etc/fish/config.fish' 'etc/fish/msys2.fish' 'etc/fish/perlbin.fish')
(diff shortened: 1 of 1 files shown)


ognevny:
> I think you need `WITH_DOCS` option, and to add sphinx (previously I used the mingw one) to the makedepends.

I would prefer using cygwin version, but we haven't packaged it yet

Berrysoft:
If it is a policy of `MSYS2-packages`, I think we have to package sphinx first...

striezel:
Yes, not having Sphinx packaged for MSYS2 (but only [for MINGW64](https://packages.msys2.org/base/mingw-w64-python-sphinx)) is the whole point.

Currently, Sphinx 9.0.4 has the following dependencies in its pyproject.toml:
```toml
dependencies = [
    "sphinxcontrib-applehelp>=1.0.7",
    "sphinxcontrib-devhelp>=1.0.6",
    "sphinxcontrib-htmlhelp>=2.0.6",
    "sphinxcontrib-jsmath>=1.0.1",
    "sphinxcontrib-qthelp>=1.0.6",
    "sphinxcontrib-serializinghtml>=1.1.9",
    "Jinja2>=3.1",
    "Pygments>=2.17",
    "docutils>=0.20,<0.23",
    "snowballstemmer>=2.2",
    "babel>=2.13",
    "alabaster>=0.7.14",
    "imagesize>=1.3",
    "requests>=2.30.0",
    "roman-numerals>=1.0.0",
    "packaging>=23.0",
    "colorama>=0.4.6; sys_platform == 'win32'",
]
```
From that list we currently have [Jinja2](https://packages.msys2.org/base/python-jinja), [Pygments](https://packages.msys2.org/base/python-pygments), [docutils](https://packages.msys2.org/base/python-pygments) and [packaging](https://packages.msys2.org/base/python-packaging) in MSYS2. This means we would have to package more than ten Python packages just to get the documentation. Is it worth it?

lazka:
If someone wants to add those Python packages, feel free. I don't think it should block this update though. I can't imagine there being many users of the offline docs.

striezel:
I've added those packages in <https://github.com/msys2/MSYS2-packages/pull/5905>.

Rebuilding fish 4.3.x with Sphinx present generates the documentation files. The build process seems to detect Sphinx automatically. At least that's what happened when I built it on my machine.

striezel:
Now that Sphinx is available it also builds the documentation. :)

PR #5911, opened 2025-12-30
uv-build: Add 0.9.20

lazka:
This just adds the build backend, for starters.


ognevny:
bzip2-sys patch is not required. build-script will run pkgconfig on Cygwin because it's not `target_os = windows`

PR #5919, opened 2026-01-01
ncurses: update to 6.6

MehdiChinoune:
using cygwin triplet breaks the entire environment.

@lazka could you do something about it?


lazka:
Interesting, thanks, not sure yet

MehdiChinoune:
OK, I'll merge it then.

PR #5937, opened 2026-01-09
libblake3: update to 1.8.3

ognevny:
remove workaround, source should be fixed to not use AVX512


ognevny:
pushing this, anyway it's not mingw...

PR #5957, opened 2026-01-16
arturo: add first package 0.10.0

RickBarretto:
## Proposal

This PR proposes add [Arturo Programming Language](https://arturo-lang.io/) to MSYS2 repository.

Arturo on Windows is portable and all needed files are under our zip file, including our DLL dependencies. 

All our files, such as Arturo itself, DLLs and Certificates goes into `~/.arturo/bin/`.

### DIFF
--- a/arturo/PKGBUILD
+++ b/arturo/PKGBUILD
@@ -1,0 +1,31 @@
+# Maintainer: https://github.com/arturo-lang
+
+pkgname=arturo
+pkgver=0.10.0
+pkgrel=1
+pkgdesc="Simple, expressive & portable programming language for efficient scripting."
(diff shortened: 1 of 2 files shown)


lazka:
This is the wrong repo -> https://github.com/msys2/MINGW-packages

Also we build all our packages from source, while this is just unpacking a zip with pre-built binaries.

RickBarretto:
> This is the wrong repo -> https://github.com/msys2/MINGW-packages
> 
> Also we build all our packages from source, while this is just unpacking a zip with pre-built binaries.

Ohhh, I see... I'll open a PR there, compiling from source, so. :wink:

PR #5969, opened 2026-01-22
colordiff faild to downloading a URL

lygstate:
"/usr/bin/xmllint" --noout --nonet --xinclude --postvalid --noent  "/e/work/xemu/MSYS2-packages/colordiff/src/colordiff-1.0.21/colordiff.xml" xmlto: /e/work/xemu/MSYS2-packages/colordiff/src/colordiff-1.0.21/colordiff.xml does not validate (status 3) xmlto: Fix document syntax or use --skip-validation option /e/work/xemu/MSYS2-packages/colordiff/src/colordiff-1.0.21/colordiff.xml:6: I/O warning : failed to load "http://www.oasis-open.org/docbook/xml/4.1.2/docbookx.dtd": No such file or directory ]>
  ^
I/O warning : failed to load "http://www.oasis-open.org/docbook/xml/4.1.2/docbookx.dtd": No such file or directory Document /e/work/xemu/MSYS2-packages/colordiff/src/colordiff-1.0.21/colordiff.xml does not validate

seems the cause is libxml2, but don't know what's happened.

PR #5974, opened 2026-01-22
libxml2: apply patch Resolve URI after xmlCatalogResolve

lygstate:
Apply MR https://gitlab.gnome.org/GNOME/libxml2/-/merge_requests/378

Should be able Closes: https://github.com/msys2/MSYS2-packages/pull/5969


lygstate:
According to https://gitlab.gnome.org/GNOME/libxml2/-/issues/1046#note_2658627

Seems the valid fixes should update file `docbook-xml` instead.

PR #5976, opened 2026-01-22
breezy: Bump to version 3.3.20 that support python 3.12.x

lygstate:
The new versrion need package setuptools_gettext that not exist in msys repository


lygstate:
Still lack a `setuptools gettext` package

PR #5977, opened 2026-01-23
docbook-xml: Use file:/// instead of http:// in file etc/xml/docbook-xml

lygstate:
The rewriteSystem and rewriteURI would not take effect according to

https://gitlab.gnome.org/GNOME/libxml2/-/issues/1046#note_2658627

Closes: https://github.com/msys2/MSYS2-packages/pull/5974

Closes: https://github.com/msys2/MSYS2-packages/pull/5969


lazka:
nice, thanks

PR #5979, opened 2026-01-24
llvm: Bump to version 21.1.8 and also add dependencies to libxml2-devel and libedit-devel

lygstate:


lygstate:
@lazka hi, anything prevent this?

lazka:
No, thanks for the ping.

Just curious, what's the motivation for the new deps?

lygstate:
> No, thanks for the ping.
> 
> Just curious, what's the motivation for the new deps?

When those deps are intalled locally, then makepkg will automatilly link to those, for avoid annoy, I add then as a deps.
Another thing, as llvm can use them, I think it's fine add them as deps.

The annoy comes from switch `msys-2.0.dll` to `cygwin1.dll`
in repo:
https://github.com/lygstate/CYGWIN-packages

I am now able building all msys2 packages with `cygwin1.dll` and can dropping many `msysize` patches

PR #5991, opened 2026-01-29
mksh: Fixes SHA256 for http://www.mirbsd.org/TaC-mksh.txt

lygstate:
As the link http://www.mirbsd.org/TaC-mksh.txt is changing time to time, so downloading it instead.

Mainly because the lastline is time and keep changing

### DIFF
--- a/mksh/.gitignore
+++ b/mksh/.gitignore
@@ -1,1 +1,0 @@
-/TaC-mksh.txt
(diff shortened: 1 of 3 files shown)


lazka:
Why the change to a mirror?

lygstate:
> Why the change to a mirror?

because CI failed https://github.com/msys2/MSYS2-packages/actions/runs/21497917610

lazka:
ah, right

PR #5992, opened 2026-01-30
zsh: Enable compile with gcc 15.x

lygstate:
Also sync the patches from fedora

fixes symbolic linkage of help files

### DIFF
--- a/zsh/PKGBUILD
+++ b/zsh/PKGBUILD
@@ -3,7 +3,7 @@
 pkgbase=zsh
 pkgname=('zsh' 'zsh-doc')
 pkgver=5.9
-pkgrel=4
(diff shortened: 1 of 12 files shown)


zsh/PKGBUILD
```diff
@@ -3,8 +3,9 @@
 pkgbase=zsh
 pkgname=('zsh' 'zsh-doc')
 pkgver=5.9
-pkgrel=4
+pkgrel=5
 arch=('i686' 'x86_64')
+options=(!zipman)
```
lazka: why?

lygstate: It's stopped when packing docs before :( locally, let's me try again

PR #6089, opened 2026-03-07
uv: update to 0.10.9

striezel:


striezel:
Obsolete, due to <https://github.com/msys2/MSYS2-packages/pull/6108>.

PR #6179, opened 2026-04-15
openssl: Update to 4.0.2

lazka:


lazka:
let's try rebuilds after the next curl release

PR #6188, opened 2026-04-16
git-extras: fix build failure when MSYS=winsymlinks:native

lygstate:
…ile"

### DIFF
--- a/git-extras/PKGBUILD
+++ b/git-extras/PKGBUILD
@@ -4,7 +4,7 @@ pkgname=${_realname}
 replaces=("${_realname}-git")
 conflicts=("${_realname}-git")
 pkgver=7.4.0
-pkgrel=1
+pkgrel=2
 pkgdesc="GIT utilities -- repo summary, commit counting, repl, changelog population and more"
 arch=('any')
 license=('MIT')
@@ -17,6 +17,7 @@ sha256sums=('aaab3bab18709ec6825a875961e18a00e0c7d8214c39d6e3a63aeb99fa11c56e')
 
 build(){
   cd "${srcdir}"
+  MSYS=winsymlinks:deepcopy
   bsdtar -zxf "${_realname}-${pkgver}.tar.gz" || true
 
   cd "${srcdir}/${_realname}-${pkgver}"


ognevny:
cc @dscho

lygstate:
What's the default MSYS in CI?
I am using

```
$ echo $MSYS
winsymlinks:native
```
locally.

ognevny:
by default winsymlinks set to deep copy option, so all symlinks are actually copies of sources

lygstate:
OK, I switch it to MSYS=winsymlinks:native and now it's fine for bsdtar

git-extras/PKGBUILD
```diff
@@ -17,10 +17,9 @@ sha256sums=('aaab3bab18709ec6825a875961e18a00e0c7d8214c39d6e3a63aeb99fa11c56e')
 
 build(){
   cd "${srcdir}"
-  bsdtar -zxf "${_realname}-${pkgver}.tar.gz" || true
 
-  cd "${srcdir}/${_realname}-${pkgver}"
-  cp bin/git-scp bin/git-rscp
+  MSYS=winsymlinks:native \
+  bsdtar -zxf "${_realname}-${pkgver}.tar.gz"
```
dscho: Even if many people are lucky enough to run MSYS2 on a recent Windows machine in Developer Mode, not all MSYS2 users are that lucky. And those would no longer be able to use this recipe to build.

Maybe it would be better to go the other way round, using `MSYS=winsymlinks:deepcopy` and keep most of the original code as-is?

lygstate: `The difference between winsymlinks:native and winsymlinks:nativestrict is this: If the filesystem supports native symlinks and Cygwin fails to create a native symlink for some reason, it will fall back to creating Cygwin default symlinks with winsymlinks:native, while with winsymlinks:nativestrict the symlink(2) system call will immediately fail.
`
So I think `winsymlinks:native` is a better opton indeed. According to https://cygwin.com/cygwin-ug-net/using-cygwinenv.html

lygstate: MSYS=winsymlinks:deepcopy have the following error that are hidden.:

```
git-extras-7.4.0/bin/git-rscp: Can't create 'git-extras-7.4.0/bin/git-rscp': No such file or directory
bsdtar: Error exit delayed from previous errors
```

dscho (commented):
With substantially improved commit message, I would be totally comfortable with this change for now. https://github.blog/2022-06-30-write-better-commits-build-better-projects/ might be helpful, in particular this part:

  | What you’re doing | Why you’re doing it
-- | -- | --
High-level (strategic) | **Intent** (what does this accomplish?) | **Context** (why does the code do what it does now?)
Low-level (tactical) | **Implementation** (what did you do to accomplish your goal?) | **Justification** (why is this change being made?)

lygstate:
> With substantially improved commit message, I would be totally comfortable with this change for now. https://github.blog/2022-06-30-write-better-commits-build-better-projects/ might be helpful, in particular this part:
> 
>  	What you’re doing	Why you’re doing it
> High-level (strategic)	**Intent** (what does this accomplish?)	**Context** (why does the code do what it does now?)
> Low-level (tactical)	**Implementation** (what did you do to accomplish your goal?)	**Justification** (why is this change being made?)

Updated, don't know if it's OK to you.

PR #6197, opened 2026-04-18
For windows 10 and windows 11, fixes bulding msys2 under case sensitive folder

lygstate:
@lazka  this MR is not huge, only character case changes, so it's easy to review.


lygstate:
@lazka Do I need split this MR into multiple PR?

lazka:
no, it's fine

lygstate:
@lazka ping for merge

striezel:
I had to revert two commits before the merge:

* commit 2aebe95394bb6e4b3958f1379d7ebd1c2c63d601, because it is not necessary to update package versions, because the package content will not change despite the renames (as seen in the file-diff)
* commit d916682c2b25f0446446b2d60983508b60aa0fe6 because (and correct me if I'm wrong here) that would change the package name and I think this might cause problems for packages depending on it. We should probably use `replaces` and / or `provides` to give a smooth migration when updating to the new name.

PR #6207, opened 2026-04-20
ignore files.

lygstate:


.gitignore
```diff
@@ -20,3 +20,19 @@
 /*/*-git/
 /*/*-svn/
 /*/*-hg/
+/*/srclinks*
+/*/*/hooks
+/*/*/objects
+/*/*/HEAD
+/*/*/config
```
lazka, line 27: just wondering, is that all git stuff, and doesn't git auto ignore that anyway?

ognevny: in my machine not :(

lygstate: Unless the cloned dirname ends with `.git` then we can ignore with `/*/*.git`, otherwise we have to ignore with this.

lazka: ah right.

Maybe we should start appending a .git to them?

PR #6210, opened 2026-04-21
unzip: Fixes building with gcc 15.

lygstate:
These patches are synced from https://src.fedoraproject.org/rpms/unzip/blob/rawhide/f/unzip.spec

The build script are comes from cygwin

Fixes symbolic links when install

### DIFF
--- a/unzip/PKGBUILD
+++ b/unzip/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=unzip
 pkgver=6.0
-pkgrel=3
+pkgrel=4
 pkgdesc="Unpacks .zip archives such as those made by PKZIP"
 arch=('i686' 'x86_64')
 url="http://www.info-zip.org/"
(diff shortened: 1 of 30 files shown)


lazka:
oof

lygstate:
> oof

mmm, many cve patches synced.

oech3:
We should just drop this package. 1st step: https://github.com/msys2/MSYS2-packages/pull/6359

unzip/PKGBUILD
```diff
@@ -73,5 +156,10 @@ package() {
 
   # install the license file
   install -Dm644 LICENSE ${pkgdir}/usr/share/licenses/unzip/LICENSE
+
+  pushd ${pkgdir}/usr/bin
+  rm -f zipinfo
+  ln -s unzip.exe zipinfo
```
oech3:
```suggestion
  ln -f unzip.exe zipinfo
```

PR #6224, opened 2026-04-26
gnupg: Update to v2.5.19

dscho:
This upgrades GnuPG from v2.4.9 to v2.5.19, the current stable release.

## Why now?

GnuPG v2.5 has been promoted from "development" to **stable**. The [download page](https://gnupg.org/download/index.html) now labels v2.5.19 as "GnuPG (stable)" and v2.4.9 as "GnuPG (oldstable)".

Werner Koch's [release announcement for v2.5.19](https://lists.gnupg.org/pipermail/gnupg-announce/2026q2/000504.html) is explicit:

> Note that the old 2.4 series reaches end-of-life in just two months. Thus update to 2.5.19 in time.

The EOL date for v2.4 is **2026-06-30**, roughly two months from now.

## History

GnuPG v2.5 was initially considered a development branch (odd minor version, following the Linux kernel convention). In March 2025, [an earlier attempt to upgrade to v2.5.5](https://github.com/msys2/MSYS2-packages/pull/5244) was abandoned for this reason, and [the corresponding git-for-windows/git ticket](https://github.com/git-for-windows/git/issues/5471#issuecomment-2710934113) was closed with the note that "v2.5.5 is not on the stable version train."

The promotion to stable happened around the v2.5.16 release in late December 2025. The [GnuPG EOL table](https://gnupg.org/download/index.html#end-of-life) now lists "2.5, 2.6" together as the current Long Term Support track.

Arch Linux still ships v2.4.9 as of this writing. Their gnupg maintainer [has noted](https://github.com/msys2/MSYS2-packages/pull/5244#issuecomment-2713102678) that they "usually wait until Arch updates gnupg since there have been compat issues in the past." With upstream now unambiguously designating v2.5.19 as stable and v2.4's EOL imminent, there is no reason to wait for Arch to move first.

## What's new in v2.5?

The headline features of the v2.5 series compared to v2.4 are:

- **Post-quantum cryptography**: Kyber (ML-KEM / FIPS-203) encryption support
- **64-bit Windows improvements**
- Full compatibility with previous versions is maintained

Beyond PQC, the changes are mostly internal, leveraging newer features from supporting libraries.

## Patch changes

The [freepg project](https://gitlab.com/freepg/gnupg/-/tree/main) maintains the extra patches that both Arch and we carry. For v2.5, these come from the `master-freepg` series instead of `STABLE-BRANCH-2-4-freepg`, with different numbering.

**Dropped** (no longer needed):
- "tests: add test cases for import without uid" (test files, script, and Makefile.am entry all present in v2.5.19 upstream)
- "Sync compliance mode cleanup with master" (was a backport *from* master to 2.4)
- "Do not use a default when asking for another output filename" (upstreamed in [v2.5.14](https://dev.gnupg.org/T7908))

**Added** for the master series:
- Revert the removal of the default keyserver (0028)
- Add missing test files to EXTRA_DIST (0040)
- Skip trust packets during import/restore (0041, gpg.fail mitigation)
- Compat flag for truncated-line handling (0042)
- Fail on unprintable armor headers (0043, gpg.fail mitigation)

**Fixed**: The freepg 0042 patch as distributed defines `COMPAT_IGNORE_TRUNCATED_LINE` as 32, colliding with the upstream `COMPAT_ALLOW_NOT_DASH_ESCAPED` flag (also 32, added in v2.5.15). Fixed locally to 64.

**Refreshed**: The `avoid_beta_warning` patch was regenerated against v2.5.19's `autogen.sh` (context drift in `$matchexcl` and `cid` fields) and renamed to `gnupg-2.5-avoid_beta_warning.patch`.

The MSYS2-specific patch (`0001-gnupg-2.2.8-msys2.patch`) applies cleanly without modification.


lazka:
I'm a bit worried since Arch isn't motivated to update and we usually wait for them. But if upstream is forcing us we kinda have to.

dscho:
> I'm a bit worried since Arch isn't motivated to update and we usually wait for them. But if upstream is forcing us we kinda have to.

Right. I also checked whether Arch was moving, and I checked whether there had been any proper announcement of the GnuPG project that they were no longer considering the v2.4/v2.5 stable/development split, but true to form they mentioned this in passing.

Master-Hash:
https://lwn.net/Articles/1055053/

I heard that GnuPG 2.5 introduced a new LibrePGP protocol, which is incompatible with OpenPGP, and that's why Fedora, ... haven't updated to GnuPG 2.5 so far.

Anyone knows that is it solved now?

I only wish this pr would break nothing...

dscho:
@Master-Hash oh wow! I don't know how I missed this drama, thank you for pointing to this!

In light of the backward-incompatible change, I think I'll "soft-retract" this PR by demoting it to a draft.

My immediate reaction to reading up about  all this drama is to consider staying with 2.4 plus patches from https://gitlab.com/freepg/gnupg...

Master-Hash:
Latest progress: https://lwn.net/Articles/1072870/ (see comments)

dscho:
@Master-Hash thank you! It does seem as if we need to look for alternatives to suggest. Do you have any suggestions?

Master-Hash:
> @Master-Hash thank you! It does seem as if we need to look for alternatives to suggest. Do you have any suggestions?

I'm sorry, I'm not an expert about PGP ecosystem, just a normal user. I'll forward the latest news here.

> My immediate reaction to reading up about all this drama is to consider staying with 2.4 plus patches from [gitlab.com/freepg/gnupg](https://gitlab.com/freepg/gnupg)...

I agree with this.

PR #6241, opened 2026-04-29
curl: Update to 8.20.0

lazka:


lazka:
We need to have a watch at https://github.com/curl/curl/issues/21466#issuecomment-4349261084 before we rebuild pacman against this.

PR #6263, opened 2026-05-08
uv: Update to 0.11.10

lazka:


ognevny:
it could be 0.11.11 :D

lazka:
ah, right, next time :)

lazka:
the full uv fails in aws-lc-sys now, lots of errors, so I didn't bother.

ognevny:
it seems that aws-lc just doesn't support cygwin...

ognevny:
there was an attempt: https://github.com/aws/aws-lc/pull/2544

Issue #6264, opened 2026-05-08
Cleanup /etc/profile* and speedup bash


dragon-archer:
Currently /etc/profile* are very heavy, significantly increasing bash startup time. After some investigating, I found several potential cleanups.

To begin with, I use `hyperfine --warmup 3 "C:/msys64/usr/bin/bash.exe -l -c 'exit'"` to measure the startup time of bash.

Baseline results (clean installation with empty ~/.bashrc):
```
$ hyperfine --warmup 3 "C:/msys64/usr/bin/bash.exe -l -c 'exit'"
Benchmark 1: C:/msys64/usr/bin/bash.exe -l -c 'exit'
  Time (mean ± σ):     326.4 ms ±  17.1 ms    [User: 90.0 ms, System: 212.5 ms]
  Range (min … max):   300.7 ms … 354.0 ms    10 runs
```

**Cleanup 1: Optimizing `profile_d` in /etc/profile**
Before:
```
  local file=
  for file in $(export LC_COLLATE=C; echo /etc/profile.d/*.$1); do
    [ -e "${file}" ] && . "${file}"
  done
  
  if [ -n "${MINGW_MOUNT_POINT}" ]; then
    for file in $(export LC_COLLATE=C; echo ${MINGW_MOUNT_POINT}/etc/profile.d/*.$1); do
      [ -e "${file}" ] && . "${file}"
    done
  fi
```
After:
```
  for file in /etc/profile.d/*.$1; do
    [ -e "${file}" ] && . "${file}"
  done

  if [ -n "${MINGW_MOUNT_POINT}" ]; then
    for file in ${MINGW_MOUNT_POINT}/etc/profile.d/*.$1; do
      [ -e "${file}" ] && . "${file}"
    done
  fi
```
Results:
```
$ hyperfine --warmup 3 "C:/msys64/usr/bin/bash.exe -l -c 'exit'"
Benchmark 1: C:/msys64/usr/bin/bash.exe -l -c 'exit'
  Time (mean ± σ):     280.5 ms ±  28.6 ms    [User: 89.7 ms, System: 163.4 ms]
  Range (min … max):   237.6 ms … 331.9 ms    11 runs
```

**Cleanup 2: Commenting out installer-only logics**
I commented the following part in /etc/profile and removed /etc/profile.d/000-msys2.sh
```
for postinst in $(export LC_COLLATE=C; echo /etc/post-install/*.post); do
[...]

PR #6265, opened 2026-05-08
filesystem: speedup /etc/profile

dragon-archer:
Implement cleanup 1,3,5 in #6264

### DIFF
--- a/filesystem/bash.bashrc
+++ b/filesystem/bash.bashrc
@@ -9,11 +9,11 @@
 
 # System-wide bashrc file
 
-# Check that we haven't already been sourced.
-([[ -z ${CYG_SYS_BASHRC} ]] && CYG_SYS_BASHRC="1") || return
-
 # If not running interactively, don't do anything
 [[ "$-" != *i* ]] && return
+
+# Check that we haven't already been sourced.
+([[ -z ${CYG_SYS_BASHRC} ]] && CYG_SYS_BASHRC="1") || return
 
 # If started from sshd, make sure profile is sourced
 if [[ -n "$SSH_CONNECTION" ]] && [[ "$PATH" != *:/usr/bin* ]]; then
--- a/filesystem/bash.bashrc
+++ b/filesystem/bash.bashrc
@@ -9,11 +9,11 @@
 
 # System-wide bashrc file
 
-# Check that we haven't already been sourced.
-([[ -z ${CYG_SYS_BASHRC} ]] && CYG_SYS_BASHRC="1") || return
-
 # If not running interactively, don't do anything
 [[ "$-" != *i* ]] && return
+
+# Check that we haven't already been sourced.
+[[ -z "${CYG_SYS_BASHRC}" ]] && CYG_SYS_BASHRC="1" || return
 
 # If started from sshd, make sure profile is sourced
 if [[ -n "$SSH_CONNECTION" ]] && [[ "$PATH" != *:/usr/bin* ]]; then
--- a/filesystem/profile
+++ b/filesystem/profile
@@ -93,13 +93,12 @@ unset p
 # Shell dependent settings
 profile_d ()
 {
-  local file=
-  for file in $(export LC_COLLATE=C; echo /etc/profile.d/*.$1); do
+  for file in /etc/profile.d/*.$1; do
     [ -e "${file}" ] && . "${file}"
   done
-  
+
   if [ -n "${MINGW_MOUNT_POINT}" ]; then
-    for file in $(export LC_COLLATE=C; echo ${MINGW_MOUNT_POINT}/etc/profile.d/*.$1); do
+    for file in ${MINGW_MOUNT_POINT}/etc/profile.d/*.$1; do
       [ -e "${file}" ] && . "${file}"
     done
   fi
@@ -111,7 +110,13 @@ done
 
 if [ ! "x${BASH_VERSION}" = "x" ]; then
   [ -z "${HOSTNAME}" ] && HOSTNAME="$(exec /usr/bin/hostname)"
-  SHELL=`which bash`
+  case "${SHELL##*/}" in
+    bash*)
+      export SHELL="${SHELL%%.exe}"
+      ;;
+    *)
+      export SHELL="$(which bash)"
+  esac
   profile_d sh
   [ -f "/etc/bash.bashrc" ] && . "/etc/bash.bashrc"
 elif [ ! "x${KSH_VERSION}" = "x" ]; then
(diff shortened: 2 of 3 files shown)


dragon-archer:
Update: POSIX `test` or `[ ]` doesn't support regex matching, so use `case` instead.

dragon-archer:
@lazka How about merging this? I've tested it for a long time locally, and it worked perfectly.

filesystem/profile
```diff
@@ -93,13 +93,12 @@ unset p
 # Shell dependent settings
 profile_d ()
 {
-  local file=
```
Omikorin, line 96: If you removed `local` you should `unset` the `file` variable after the `if` block.

I changed execution to echo to make it more visible:

```text
me@universe UCRT64 ~
$ echo $file


me@universe UCRT64 ~
$ profile_d ()
{
  for file in /etc/profile.d/*.$1; do
    [ -e "${file}" ] && echo "${file}"
  done

  if [ -n "${MINGW_MOUNT_POINT}" ]; then
    for file in ${MINGW_MOUNT_POINT}/etc/profile.d/*.$1; do
      [ -e "${file}" ] && echo "${file}"
    done
  fi
}

me@universe UCRT64 ~
$ profile_d sh
/etc/profile.d/000-msys2.sh
/etc/profile.d/bash_completion.sh
/etc/profile.d/lang.sh
/etc/profile.d/perlbin.sh
/etc/profile.d/tzset.sh

me@universe UCRT64 ~
$ echo $file
/ucrt64/etc/profile.d/*.sh

me@universe UCRT64 ~
$ unset file

me@universe UCRT64 ~
$ echo $file


```

Omikorin: Moreover, removal of `LC_COLLATE` can be risky for some languages if execution order matters.
I didn't found any problems for current `profile.d` scripts with C, en_US.UTF-8, cs_CZ.UTF-8, da_DK.UTF-8 (random examples) but I guess when ordering matters then it should use an explicit index-name style.

Anyway, simple `for ... in ...` loop is much faster than spawning subshells.

dragon-archer: > If you removed `local` you should `unset` the `file` variable after the `if` block.

Thanks for pointing out this. I'll fix this.

> Moreover, removal of `LC_COLLATE` can be risky for some languages if execution order matters.
> I didn't found any problems for current `profile.d` scripts with C, en_US.UTF-8, cs_CZ.UTF-8, da_DK.UTF-8 (random examples) but I guess when ordering matters then it should use an explicit index-name style.

As all files in `profile.d` should be named using only ASCII characters and no space, I guess this shouldn't matter? A safer (and much more complicated) way might be saving LC_COLLATE, setting it to C, do `for ... in ...`, then restore it if not empty. But this is actually not safest about 'empty but set LC_COLLATE'.

Omikorin: > As all files in `profile.d` should be named using only ASCII characters and no space, I guess this shouldn't matter?

There are some quirks like [Some apps unexpectedly sorting "Aa" after "Z" (Danish Collation)](https://apple.stackexchange.com/questions/464492/some-apps-unexpectedly-sorting-aa-after-z-danish-collation). Same with "ch" and "h" in Czech language.

I believe nobody really uses `/etc/profile.d` on MSYS2 to that extent that it would affect them. After all, if you want an execution order, then you knowingly use the "0nnn-task.sh" naming (as I said before). I think this change is worth trying, but I'd like to hear what maintainers think. Fun fact, your change should be compatible even with names using spaces because of native for loop instead of that bash output insertion.

> As all files in `profile.d` should be named using only ASCII characters and no space, I guess this shouldn't matter? A safer (and much more complicated) way might be saving LC_COLLATE, setting it to C, do `for ... in ...`, then restore it if not empty. But this is actually not safest about 'empty but set LC_COLLATE'.

I think it is not worth trying. Although my experience with `profile.d` is small, I've never encountered `LC_COLLATE` there (but Windows is weird so who knows).

Recently, I informed one big C++ project about the MINGW64 deprecation. They had a problem with `LC_ALL` when migrating to UCRT64.

This is the code the maintainer come up with:

```cpp
void configureLocale() {
    #if defined (OS_WINDOWS)
        // MSYS2 UCRT64 runtime does not support any sort of locales except for the "C" locale.
        // See https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/config/locale/generic/c_locale.cc#L245
        std::setlocale(LC_ALL, "C");
    #else
        // Elsewhere, set the locale to en_US.UTF-8 to ensure that we can handle UTF-8 strings correctly
        // falling back to the C locale if the rest isn't available
        std::setlocale(LC_ALL, "en_US.UTF-8") ||
        std::setlocale(LC_ALL, "C.UTF-8")     ||
        std::setlocale(LC_ALL, "C");
    #endif
}
```

filesystem/bash.bashrc
```diff
@@ -9,12 +9,12 @@
...
 # System-wide bashrc file
 
-# Check that we haven't already been sourced.
-([[ -z ${CYG_SYS_BASHRC} ]] && CYG_SYS_BASHRC="1") || return
-
 # If not running interactively, don't do anything
 [[ "$-" != *i* ]] && return
 
+# Check that we haven't already been sourced.
+([[ -z ${CYG_SYS_BASHRC} ]] && CYG_SYS_BASHRC="1") || return
```
Omikorin:
```suggestion
[[ -n "${CYG_SYS_BASHRC}" ]] && return
CYG_SYS_BASHRC="1"
```

Omikorin: Solves https://github.com/msys2/MSYS2-packages/issues/6264#issuecomment-5515440416

 I chose a guard clause in place of a one-liner for readability.

You can click to accept changes, so it is quicker than copying stuff

Omikorin (changes requested):

filesystem/bash.bashrc
dragon-archer: Thanks, though "apply suggestion" is actually not very useful for MSYS2, because it won't automatically update sha256sums.

Omikorin: Yeah, it still needs some work

dragon-archer: Updated.

filesystem/profile
dragon-archer: Fixed `local file=`.

filesystem/bash.bashrc
Omikorin: Great! It really works now :)

Just a side note: force-pushing on PR makes it difficult to read changes since last review (also from notifications on GitHub). These changes are squashed often so it does not matter what happened on the branch. You may have a different reasoning, obviously

dragon-archer: > Just a side note: force-pushing on PR makes it difficult to read changes since last review (also from notifications on GitHub). These changes are squashed often so it does not matter what happened on the branch. You may have a different reasoning, obviously

Ah, you're right. I usually rewrite and force-push a PR to make each commit self-contained, so for new reviewers it'd be cleaner. Sadly this indeed makes old reviewers harder to find "what changed since last review". Sorry for the inconvenience.

Omikorin: Yeah, it's a matter of taste. I found out using atomic commits a game changer to explain my reasoning (okay, generic cleanup is difficult to explain in one line :D). 
In example, like I split my PR into updating the package along with metadata and a cleanup of my choice.

But it really depends on the project you worked on as it is not a golden rule of course.

PR #6285, opened 2026-05-16
update parallel to v20260422

oltolm:
I fixed two problems:
1. I hard coded a maximum command line length of 32000 because parallel was hanging indefinitely trying to determine it.
1. I hard coded the MSYS2 perl because it does not work with a native perl.


oltolm:
I did fix the hang, that made `parallel` unusable, but it still only works if NO native perl interpreter is installed. It break if for example the UCRT64 perl is installed. I tried restoring the deleted patch, but even with that patch it does not work and I don't know why.

PR #6316, opened 2026-05-27
zip: Fixes building with gcc 15.x

lygstate:
zip.h:728:8: error: conflicting types for 'memcmp'; have 'int(char *, char *, unsigned int)'
unix/unix.c:71:14: error: conflicting types for 'DIR'; have 'FILE'
unix/unix.c:244:13: error: implicit declaration of function 'isascii' [-Wimplicit-function-declaration]


oech3:
Is this package still needed?

lazka:
it's required by lots of packages at least

oech3:
Can we use Fedora's (un)zip repo as an upstream to simplify?

PR #6317, opened 2026-05-27
.patch files should be binary files

lygstate:
Because in patch files, there may line ending changes. Keep it in binary format to support for all kinds of patches.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -18,3 +18,4 @@
 *.lib binary
 *.a binary
 *.ico binary
+*.patch binary


striezel:
This change does not seem right.

Patch files are basically text files, not binary files. So they should be treated as text files and not as binary files. Furthermore, the suggested change would render `git diff` on patch files useless, because instead of showing the changes it would just print the "Binary files differ" message without showing the actual differences. That obscures changes to patch files and makes it harder to see what actually changed. All in all that makes things worse for developers / maintainers.

If you are concerned about line ending changes, then `git diff` has [the `-w` option](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt--w) which ignores whitespace changes.

I'm closing this.

lygstate:
I create this MR because I faced real issue, indeed the msys-runtime is binary, because it's commit with line ending modified, so when apply those patches, warnings rasied.

striezel:
If I remember this correctly, then those warnings were about trailing whitespace in some patch files for msys-runtime, not about some binary stuff. In that case the patches should probably be adjusted to remove the trailing whitespace characters so that the warnings are gone. Since those patches are generated from a separate repository, it should probably be fixed there.

PR #6323, opened 2026-05-28
apr:  fixes building with gcc 15

lygstate:
file_io/unix/filedup.c:40:14: error: implicit declaration of function 'dup3'; did you mean 'dup2'? [-Wimplicit-function-declaration]
misc/unix/errorcodes.c:385:9: error: assignment to 'const char *' from 'int' makes pointer from integer without a cast [-Wint-conversion]


lazka:
what about the dropped patches?

lygstate:
> what about the dropped patches?

Do not know why them there, it's not used before this PR, so I removed it

lazka:
ok, looks fine I guess.

PR #6359, opened 2026-06-11
libarchive: install bsdunzip

oech3:


oech3:
Do we need to split `bsdcpio` and `bsdtar` binaries even tyey are small enough? Arch does not do that.

lazka:
yeah, I was wondering the same. I think it's because it is part of the base set: https://packages.msys2.org/packages/base and we try to keep that lean.

oech3:
Hmm... I wanted to bundle bsdunzip to replace makedepends in https://packages.msys2.org/packages/unzip?variant=x86_64 without additional `makedepends=(bsdunzip)`...

lazka:
Could we use the mingw bsdunzip for mingw packages?

oech3:
https://github.com/msys2/MINGW-packages/pull/29920#issuecomment-4682974146

oech3:
SInce `bsdtar` supports *.zip, lacking MinGW bsdunzip is not a problem for build dep.
Some MSYS package might assume `bsdunzip` at `$PATH` as `unzip` to build.

lazka:
ah, right

PR #6366, opened 2026-06-14
crosstool-ng: drop unzip from makedepends

oech3:


oech3:
@ognevny

ognevny:
fyi bsdtar package is pulled via base-devel, so no need to put it in makedepends

oech3:
I'll try same things for other MinGW pkgs.

PR #6377, opened 2026-06-18
ci: use case-sensitive directories on Windows builds.

lygstate:
Windows CI uses case-insensitive NTFS by default, so linker and path issues are not caught until users build on case-sensitive folders. Enable NTFS case sensitivity for the checkout and MSYS2 install dirs.

uprev mintty to trigger the issue

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,6 +40,23 @@ jobs:
           # Needs to be C:, see https://github.com/al-cheb/configure-pagefile-action/issues/18
           disk-root: "C:"
 
+      # Windows is case-insensitive by default, so linkers and build scripts can
+      # succeed in CI while failing for users on case-sensitive directories (or
+      # Cygwin case-sensitive library lookups). Use NTFS per-directory flags so
+      # the checkout and MSYS2 install match that environment.
+      - name: Prepare case-sensitive directories
+        run: |
+          $drive = "${{ steps.facts.outputs.drive }}"
+          foreach ($dir in @("_", "M")) {
+            $path = Join-Path $drive $dir
+            if (Test-Path $path) { Remove-Item -Recurse -Force $path }
+            New-Item -ItemType Directory -Path $path | Out-Null
+            fsutil.exe file setCaseSensitiveInfo $path enable
+            if ($LASTEXITCODE -ne 0) { throw "Failed to enable case sensitivity on $path" }
+            $info = fsutil.exe file queryCaseSensitiveInfo $path
+            if ($info -notmatch 'enabled') { throw "Case sensitivity not enabled on $path" }
+          }
+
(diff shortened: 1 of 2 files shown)


lazka:
Just so I understand your motivation better, why don't you just build with a normal NTFS setup like everyone else?

lygstate:
> Just so I understand your motivation better, why don't you just build with a normal NTFS setup like everyone else?

of cause I can do, but Cygwin is for simulate POSIX, I want build toolchian for Linux, then we need Case sensitive folders.

striezel:
Making the CI use case-sensitive paths could potentially break dozens or even hundreds of packages for CI only while they continue to build just fine for the average Windows user who uses the default case-insensitive NTFS file system. This could mean that there is a lot of work coming towards us to fix all those packages just for the rare case where someone wants to run this on a case-sensitive file system. Is that really worth the effort?

lygstate:
> Making the CI use case-sensitive paths could potentially break dozens or even hundreds of packages for CI only while they continue to build just fine for the average Windows user who uses the default case-insensitive NTFS file system. This could mean that there is a lot of work coming towards us to fix all those packages just for the rare case where someone wants to run this on a case-sensitive file system. Is that really worth the effort?

I have already build all msys2-packages locally with case-sensative folder. And it's already green after https://github.com/msys2/MSYS2-packages/pull/6197

This mintty raised new error because there is no CI guard it after MR  https://github.com/msys2/MSYS2-packages/pull/6197

lygstate:
> Just so I understand your motivation better, why don't you just build with a normal NTFS setup like everyone else?

Another case to clarify, many people use Linux to cross-compiling msys/mingw target windows. And people also use msys to guard there project, if msys-ci can default to case-sensative folder, both of them will benifet from this. That's why mingw's lib/headers are always lower-cased.

lazka:
OK, that's a valid goal, in theory, but I'm not sure that it justifies making our lifes harder for that. Only a small subset of the software we build will be cross compiled and there is a chance we'll get random breakages without any benefit from fixing it.

lygstate:
> OK, that's a valid goal, in theory, but I'm not sure that it justifies making our lifes harder for that. Only a small subset of the software we build will be cross compiled and there is a chance we'll get random breakages without any benefit from fixing it.

I think it's should be rare case, most software is cross-platform support. see https://github.com/msys2/MSYS2-packages/pull/6197; most changes is msys2-packages only. not the specific package

lygstate:
> OK, that's a valid goal, in theory, but I'm not sure that it justifies making our lifes harder for that. Only a small subset of the software we build will be cross compiled and there is a chance we'll get random breakages without any benefit from fixing it.

the mintty pacth should be upstream. Just msys needs CI to guard them from to broekn.

lazka:
I don't think we should be the CI for other projects, I'm sorry.

I'm happy to take (not too complex) patches for fixing such issues though.

PR #6391, opened 2026-06-24
jsoncpp: Update to 1.9.8

lygstate:
- Bump pkgver from 1.9.6 to 1.9.8 and reset pkgrel to 1.
- Update source tarball sha256sum for the new release.
This is for fixes building with gcc 15.x


MehdiChinoune:
cmake should be rebuilt

lygstate:
oh cppdap also depends on this

MehdiChinoune:
> oh cppdap also depends on this

grokker shows only cmake.

lygstate:
cppdap is depends on jsoncpp, and cmake depends on cppdap, now cmake is broken.....

PR #6394, opened 2026-06-24
enscript: Fixes building with gcc 15.x

lygstate:
Oh, this is the case why patch should be binary files

https://github.com/msys2/MSYS2-packages/pull/6317

### DIFF
--- a/enscript/.gitattributes
+++ b/enscript/.gitattributes
@@ -1,0 +1,1 @@
+*.patch binary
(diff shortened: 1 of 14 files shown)


mmuetzel:
> patch should be binary files

Did you forget to set the following in your .gitconfig?
```
[core]
	autocrlf = false
```

lygstate:
it's false,

it's because of https://github.com/msys2/MSYS2-packages/blob/master/.gitattributes

https://github.com/msys2/MSYS2-packages/blob/529fe11e6bf7cf42132d837df89f373fc1879d29/.gitattributes#L1-L4

mmuetzel:
> it's because of https://github.com/msys2/MSYS2-packages/blob/master/.gitattributes

Why do we need ` * text eol=lf` at all? That's surprising. But probably off-topic for this PR.
Sorry for the noise.

lygstate:
Yeap, that's why I create this MR https://github.com/msys2/MSYS2-packages/pull/6317

PR #6395, opened 2026-06-24
Fix mingw-w64-cross-crt ISO vswprintf exact-fit check.

lygstate:
- Patch mingw_vsnprintf.c so __mingw_vswprintf returns -1 only when n or more data wide chars are needed, not when output exactly fills n-1.
- Change the post-decrement error test from >= to >; fixes swprintf(buf, 10, ...) returning -1 for a 9-char result used by msys2-launcher AppID hash.
- Bump mingw-w64-cross-crt pkgrel to 3 and apply the patch in prepare().
- Use _patches array in PKGBUILD; keep patch file beside PKGBUILD.

### DIFF
--- a/mingw-w64-cross-crt/0001-crt-Fix-ISO-vswprintf-exact-fit-off-by-one-error-che.patch
+++ b/mingw-w64-cross-crt/0001-crt-Fix-ISO-vswprintf-exact-fit-off-by-one-error-che.patch
@@ -1,0 +1,65 @@
+From 29fe00be14f74a749aab86d3d0704d4580ed8374 Mon Sep 17 00:00:00 2001
+From: Yonggang Luo <luoyonggang@gmail.com>
+Date: Wed, 24 Jun 2026 02:06:57 +0800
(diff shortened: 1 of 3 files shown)


lygstate:
msys2-launcher is statically linkage to __mingw_vswprintf, so it's need rebuild.

lazka:
was this patch upstreamed?

lygstate:
> was this patch upstreamed?

Not yet, I've just sent it to upstream, not accepted yet, but that's a real bug.

lygstate:
upstream merged now, the commit is https://github.com/msys2-apiss/mingw-w64/commit/e7e21db45eaae989b04f3c24964dead5b53310de

lazka:
Great, thank you!

Would it be OK if I just update cross mingw-w64 to current git master?

lygstate:
> Great, thank you!
> 
> Would it be OK if I just update cross mingw-w64 to current git master?

I guess won't be an issue, it's guard by the CI.

PR #6399, opened 2026-06-25
libedit: Update to 20260512 and fix Cygwin build

lygstate:
- Bump snapshot to 20260512_3.1 (pkgrel 2).
- Refactor patches with _patches and apply via prepare() loop.
- Extend cygwin-build.patch with read.c sys/ioctl.h for __CYGWIN__.
- Remove libedit-20191231-3.1.patch; wchar_t display types break Cygwin.


lazka:
> Remove libedit-20191231-3.1.patch; wchar_t display types break Cygwin.

Could you go into detail as to why? It was added in #2241 to fix openssh

lygstate:
The patch changes only `el_display` / `el_vdisplay` in `el.h` from `wint_t **` to `wchar_t **`. That was a minimal workaround for #1697 / #2241: on Cygwin `wchar_t` is 16-bit (Windows ABI) while `wint_t` is 32-bit, and libedit treats the display buffers as `wchar_t *` in several places (e.g. `terminal_overwrite()`, casts in `terminal.c`). Passing a `wint_t[]` where `wchar_t *` is expected makes the prompt look truncated (`sftp>` -> `s     _`).

The problem is that the patch never updated the helpers that allocate/free those buffers. `terminal_alloc_buffer()` and `terminal_free_buffer()` in `terminal.c` still use `wint_t **` / `wint_t ***`, and `refresh.c` also mixes the two types. You noted this at merge time in #2241.

With GCC 15, `-Wincompatible-pointer-types` is a hard error, so the incomplete patch no longer builds. I reproduced this on 20260512 with only `cygwin-build.patch` + `libedit-20191231-3.1.patch`:

```
terminal.c:462:24: error: assignment to 'wchar_t **' from incompatible pointer type 'wint_t **'
terminal.c:481:30: error: passing argument 1 of 'terminal_free_buffer' from incompatible pointer type 'wchar_t ***' (expected 'wint_t ***')
refresh.c:1123:34: error: assignment to 'wint_t *' from incompatible pointer type 'wchar_t *'
```

The same failure happens on the current 20240808 snapshot with GCC 15.3.0, so this is not specific to the version bump; it is the toolchain catching the type mismatch that the one-line `el.h` hack papered over.

Without the patch, 20260512 builds cleanly with `--enable-widec` and `cygwin-build.patch`.

Open question: whether the original openssh/sftp prompt bug still reproduces without the patch on the new snapshot. I have not re-tested sftp interactively yet. If it does, we probably need either a fuller Cygwin patch (also switch `terminal_alloc_buffer` / `terminal_free_buffer` and the `refresh.c` sites to `wchar_t`, as you suggested in #2241) or an upstream fix for the `wint_t`/`wchar_t` mix on 16-bit `wchar_t` platforms.

lazka:
I see, thanks for the details

lygstate:
> I see, thanks for the details

fixed the issue.

PR #6400, opened 2026-06-25
mingw-w64-cross: Update to 31bd54ab7d

lazka:
https://github.com/mingw-w64/mingw-w64/compare/bd663f0e1...31bd54ab7


lygstate:
I think mingw would also suffer from such issue https://github.com/msys2/MINGW-packages

lazka:
https://github.com/msys2/MINGW-packages/pull/30168

PR #6401, opened 2026-06-25
vim: Update to 9.2.0725

lazka:
remove revert since the issue was fixed in
https://github.com/vim/vim/commit/a6b4f97c7f60fe9e80908ceecdeb2f950356f820


dscho:
Thank you! I gave this a quick test in my setup, and 725 works as expected.

PR #6407, opened 2026-06-27
jsoncpp, cppdap: fix cppdap build against jsoncpp

lygstate:
- Increment pkgrel from 3 to 4 in cppdap/PKGBUILD.
- Because of https://github.com/msys2/MSYS2-packages/pull/6391, now cmake is broken as cmake depends on cppdap and cppdap depends on jsoncpp


lazka:
cmake works fine here, what do you mean is broken?

What does the patch fix?

lygstate:
> cmake works fine here, what do you mean is broken?
> 
> What does the patch fix?

Yeap, cmake works, but jsoncpp are not depends on jsoncp, the reason: https://github.com/msys2/MSYS2-packages/actions/runs/28302956827; indeed it's broken before. see the dll-deps.

PR #6410, opened 2026-06-28
sqlite: Update to 3.53.3

lazka:


sqlite/PKGBUILD
lygstate: By the way, icu-devel can be dropped. not depends on it now

lazka: thanks

PR #6454, opened 2026-07-20
trash: New package

lhmouse:
This provides the command `trash` which moves its arguments to the Recycle Bin of the current user.

The idea is from the same command on Ubuntu: https://manpages.ubuntu.com/manpages/bionic/man1/trash.1.html
(It works without a desktop environment too.)

Source code is only one C file: https://github.com/lhmouse/msys2-trash/blob/v1.1/trash.c

```text
Usage: trash [OPTIONS]... [--] PATHS...

Moves PATHS into the Recycle Bin of the current user.

Options:
  -d, --directory       (ignored)
  -f, --force           ignore nonexistent paths
  -h, --help            show help message then exit
  -i, --interactive     prompt before removing every path
  -r, -R, --recursive   (ignored)
  -v, --verbose         print status
  -V, --version         show version information then exit
```

### DIFF
--- a/trash/PKGBUILD
+++ b/trash/PKGBUILD
@@ -1,0 +1,25 @@
+# Maintainer: LIU Hao <lh_mouse@126.com>
+
+_realname=trash
+pkgname=${_realname}
+pkgver=1.1
+pkgrel=1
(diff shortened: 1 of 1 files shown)


lhmouse:
Thanks!

PR #6479, opened 2026-08-02
base: Remove dependency on pacman-contrib

MehdiChinoune:
pacman-contrib is needed for basic installation.
It pulls perl dependency, which could make MinGW-packages use implicitly MSYS perl instead of mingw-w64-perl.

### DIFF
--- a/base/PKGBUILD
+++ b/base/PKGBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Christoph Reiter <reiter.christoph@gmail.com>
 
 pkgname=base
-pkgver=2022.06
+pkgver=2026.08
 pkgrel=1
 pkgdesc='Minimal package set to define a basic MSYS2 installation'
 url='https://www.msys2.org'
@@ -31,7 +31,6 @@ depends=(
   'msys2-runtime'
   'nano'
   'pacman'
-  'pacman-contrib'
   'pacman-mirrors'
   'rebase'
   'sed'


lazka:
hm, updpkgsums and pactree are quite handy though.

I'm worried that this will break things.

lazka:
If we really want to do this, I'd rather make perl optional since it's only used by pacsearch

MehdiChinoune:
> hm, updpkgsums and pactree are quite handy though.
> 
> I'm worried that this will break things.

what things, those tools are only required by someone who is building packages (updpkgsums) or quering (pactree).

lazka:
not sure tbh, maybe some CI scripts?

another thing is `pacdiff` which we mention here https://www.msys2.org/docs/updating/#managing-configuration-file-backups

lazka:
we could do optdepends like Arch: https://gitlab.archlinux.org/archlinux/packaging/packages/pacman-contrib/-/blob/main/PKGBUILD?ref_type=heads#L16-26

MehdiChinoune:
Yes, making perl optional is acceptable.
The goal is making CI jobs not pulling MSYS perl.

MehdiChinoune:
In Arch, pacman-contrib is not a dependency of base.

dragon-archer:
I think we should make `pacman-contrib` optdepends on perl, then move it to `base-devel` instead of `base`? It's unlikely that these tools will be used out of developers (and they can install it explicitly).

lazka:
saves 10mb from the installer, not bad :)

PR #6481, opened 2026-08-02
base-devel: remove flex, bison

lazka:
flex/bison are only required by a small set of tools

this also removes m4 as a transitive dep

there will likely be some fallout from flex/bison during rebuilds

### DIFF
--- a/base-devel/PKGBUILD
+++ b/base-devel/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=base-devel
 pkgver=2026.08
-pkgrel=3
+pkgrel=4
 pkgdesc='Minimal package set for building packages with makepkg'
 url='https://www.msys2.org'
 arch=('any')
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
There are no native flex and bison packages for MinGW-w64, so I think keeping them won't hurt.

lazka:
just to be clear, this is only removing them from the default makepkg install set, so we'd need to add them to makedepends instead.

dragon-archer:
file, gawk, grep, pacman, sed, and tar are alredy deps of base, and I think it's unlikely that we'll remove them from base (they are too commonly used), so perhaps we can remove them from base-devel?

Kreijstal:
lots of packages use this assumption, I guess it is fine but ugh :D

lazka:
I've pushed the depends cleanup and diffstat in master now.

PR #6491, opened 2026-08-04
pacman-contrib: move perl to optdepends

lazka:
It is only used by the pacsearch command, which isn't that useful, and this is the only thing pulling in perl in "base", so move it to optdepends

Arch also has it in optdepends for example.


lazka:
libpsl .pc file is missing some deps

lazka:
See #6479 for context

PR #6494, opened 2026-08-04
autotools: use unversioned real package.

dragon-archer:
After previous PRs, now all packages in MSYS2-packages and MINGW-packges can be built with the latest autotools. This PR aimes to provide unversioned real package of `automake` and `autoconf`.

Previous versions and `autotools-wrapper` are kept in case any package requires a specific version of autotools. Considering Arch always use the latest autoconf and automake, I think old versions of autoconf and automake can be removed here or in the near future.

Note that `${MINGW_PACKAGE_PREFIX}-autotools` also depends on `autoconf-wrapper` and `automake-wrapper`. Not sure we should update that in a transaction, or let `autoconf` and `automake` provides there wrapper (seemes strange, but may let pacman happier? Not sure whether there're any downsides of 2 packages providing each other.)


dragon-archer:
Successfully did the update on my local machine. Seemes like pacman is clever enough for this.

lazka:
How do you know it builds with the latest autoconf/automake? By default it uses the version the files were generated with, not the latest.

And what should we do on the next autoconf version that is incompatible?

dragon-archer:
[Arch](https://archlinux.org/packages/?sort=&q=autoconf&maintainer=&flagged=), [Debian](https://packages.debian.org/search?suite=forky&searchon=names&keywords=autoconf), and [Ubuntu](https://packages.ubuntu.com/search?suite=stonking&searchon=names&keywords=autoconf) all uses the latest autoconf with a few obsolete versions kept (namely 2.13 and 2.69). I think there's nothing special for us to keep so many versions.

Generally, I don't think the next autoconf version will be incompatiable to a large range of packages, and autoconf update very slowly (ususally 2 years or even longer). If some package do requires an older version, i.e. 2.69, then we can simply let it `makedepends` `autoconf2.69` instead of `autoconf`

And, as of automake, Arch, Debian, and Ubuntu all only keep the latest version, so I think it should be safer to remove other versions.

lazka:
Unlike those distros we autoreconf, they usually just call configure as is. Gentoo autoreconfs a lot also, which is why they developed the wrapper.

dragon-archer:
We usually use `autoreconf -fi` (and opionally -v), which is more compatiable then simply `autoreconf` as all files generated from older autoconf will be regenerated.

And, if something do break, we can always specify a specific autoconf version for it.

After all, we cannot keep all autoconf versions for all the time. We don't even keep older GCC (GCC16 update DO break some packages, but it turns out that fixing this type of break is easy).

lazka:
> And, if something do break, we can always specify a specific autoconf version for it.

How would you be able to select an older autoconf/automake version without the wrapper? the calls are hidden in autogen.sh/autoconf scripts and call themselves without any version.

dragon-archer:
> How would you be able to select an older autoconf/automake version without the wrapper? the calls are hidden in autogen.sh/autoconf scripts and call themselves without any version.

You're right. If called directly, the automake version can be fixed togerther with something like:
```bash
AUTOMAKE=automake-1.16 \
ACLOCAL=aclocal-1.16 \
autoreconf-2.72 -fiv
```
But it's difficult to specify it if called from `autogen.sh`, though this isn't a large case (~550 packages depend on `${MINGW_PACKAGE_PREFIX}-autotools`, but only ~50 call `autogen.sh`, and I think most of them can accept a newer autoconf version.)

Then, how about keep `autoconf-wrapper` and `automake-wrapper`, but also add direct `autoconf` and `automake`?

lazka:
> Then, how about keep `autoconf-wrapper` and `automake-wrapper`, but also add direct `autoconf` and `automake`?

They would need to conflict since they provide the same files. That seems to just complicate things imo, two more packages to deal with.

If your goal is to remove old autoconf/automake we could move some autoconf/automake versions in the wrapper from depends to optdepends. So they are not installed by default, but could be used if needed.

lazka:
I'm also fine with removing some old versions altogether.

dragon-archer:
> > Then, how about keep `autoconf-wrapper` and `automake-wrapper`, but also add direct `autoconf` and `automake`?
> 
> They would need to conflict since they provide the same files. That seems to just complicate things imo, two more packages to deal with.
> 
> If your goal is to remove old autoconf/automake we could move some autoconf/automake versions in the wrapper from depends to optdepends. So they are not installed by default, but could be used if needed.

Fine, then how about following (following Gentoo):
1. Removing `automake-1.{12..16}`, and marking `automake-1.{11,17}` as `optdepends`
2. Marking `autoconf-2.{13,69,71,72}` `optdepends`

I'll update this PR as above.

Also, I still suggest depending on `autoconf` and `automake` for `autotools` meta package, instead of explicitly depending on wrappers. Similar for https://github.com/msys2/MINGW-packages/pull/30830

autotools-wrappers/PKGBUILD
```diff
@@ -21,14 +21,12 @@ package_automake-wrapper() {
...
-    'automake1.13'
-    'automake1.14'
-    'automake1.15'
-    'automake1.16'
+    'automake1.18'
+  )
+  optdepends=(
     'automake1.17'
-    'automake1.18')
+    'automake1.11'
```
lazka, line 28: just wondering, what makes 1.11 special?

dragon-archer: Gentoo keeps 1.11 (but marks as outdated), not sure why.

lazka: ok

lazka:
Looks good in general.

I'm a bit worried about removing too much. gcc/binutils still use autoconf 2.69 + automake 1.15 for example

dragon-archer:
AFAIK, automake is more compatiable across versions, and seldomly causes issues. autoconf is relatively less compatiable, especially that some packages (e.g. gcc) requires to use the exact autoconf version. Therefore I kept all autoconf versions, and only removed old automake.

PR #6496, opened 2026-08-05
msys2-runtime: restore zipman

dragon-archer:
Almost 5 years have passed, and it should be safe to reenable zipman now.

Also did a cleanup to PKGBUILD, and updated `update-patches.sh` as well.

### DIFF
--- a/msys2-runtime/update-patches.sh
+++ b/msys2-runtime/update-patches.sh
@@ -59,11 +59,8 @@ git --git-dir msys2-runtime rev-parse --verify $msys2_branch >msys2-runtime.comm
 git add $patches msys2-runtime.commit ||
 die "Could not stage new patch set"
 
-in_sources="$(echo "$patches" | sed "{s/^/        /;:1;N;s/\\n/\\\\n        /;b1}")"
-in_prepare="$(echo "$patches" | tr '\n' '\\' | sed -e 's/\\$//' -e 's/\\/ &&&n  /g')"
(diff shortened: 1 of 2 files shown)


dragon-archer:
The `file-diff` in CI of `msys2-runtime-devel` is broken. Possibly because `msys2-runtime` is a core package?

lazka:
> The `file-diff` in CI of `msys2-runtime-devel` is broken. Possibly because `msys2-runtime` is a core package?

yeah, since our CI script uses bash we can't replace the runtime while it is running. Ideally we would move to native Python at some point there.

msys2-runtime/update-patches.sh
```diff
@@ -59,11 +59,8 @@ git --git-dir msys2-runtime rev-parse --verify $msys2_branch >msys2-runtime.comm
 git add $patches msys2-runtime.commit ||
 die "Could not stage new patch set"
 
-in_sources="$(echo "$patches" | sed "{s/^/        /;:1;N;s/\\n/\\\\n        /;b1}")"
```
lazka, line 62: what's all this about?

lygstate: I also want do this, so that _patches can only appeared a single time.

dragon-archer: Previously patches are repeated both in `sources` and in `apply_patch_with_msg`. I've moved it to an array `_patches`, to remove the duplicate. The changes to update-patches.sh is just a corresponding change.

suchat12052528-ctrl (approved):
X Ai

PR #6537, opened 2026-08-25
rust: update to 1.98.0

ognevny:


ognevny:
1.100.0 will contain my patch which force-disables compress-debuginfo for cygwin

PR #6550, opened 2026-08-28
libevent: Update to 2.1.13

lazka:
* libevent-2.1.12-openssl-compat.patch: fixed upstream
* re key, see https://github.com/libevent/libevent/issues/1894


dragon-archer:
What's the difference between #6548? It seemes that the same change is merged twice?

lazka:
Oh, weird, I assumed I had missed opening the pr because I couldn't find it

PR #6570, opened 2026-09-01
just: Update to 1.58.0

Omikorin:
Excluded some files because they break the workflow:

```text
# bsdtar -xf "just-1.58.0.tar.gz"
just-1.58.0/README.??.md: Can't create 'just-1.58.0/README.??.md': No such file or directory
just-1.58.0/www/man/en: Can't create 'just-1.58.0/www/man/en': No such file or directory
just-1.58.0/www/man/zh: Can't create 'just-1.58.0/www/man/zh': No such file or directory
bsdtar: Error exit delayed from previous errors
```

`target` library is no longer needed in the `Cargo.toml`, so I removed it too:

```text
==>B Starting prepare()...B
patching file src/lib.rs
error: package ID specification `target` did not match any packages

help: a package with a similar name exists: `tabled`
==> ERROR:B A failure occurred in prepare().B 
```


just/PKGBUILD
```diff
@@ -18,25 +18,17 @@ msys2_references=(
...
-        "https://patch-diff.githubusercontent.com/raw/casey/target/pull/20.patch")
+source=("${pkgname}-${pkgver}.tar.gz::https://github.com/casey/${pkgname}/archive/${pkgver}.tar.gz")
 noextract=("${pkgname}-${pkgver}.tar.gz")
-sha256sums=('2976e02f2dffd1ddc9cba57ef2fe75e8f4b97fde1657ee6fd145ab01efd789a7'
-            '1e8f05f774b2db35bdad5a8237a90be1102669f8ea013fea9777b366d34ab145'
-            'c2098edaf413bf7890ba169c9fc97f38539562247ff603fa6e8a7040fbba1a91')
+sha256sums=('c8a36e6e9397f2fdfcb0cc246fcdb790b52a784f3c8cabc0d8baeb031852a148')
 
 prepare() {
-  bsdtar -xf "${srcdir}/${pkgname}-${pkgver}.tar.gz" || true
```
ognevny, line 30: this should still work because of `|| true`. files you excluded here won't be extracted, but the rest will

Omikorin: Okay, I reverted the change. It didn't work at first on a devcontainer but I resolved it, thanks

PR #6571, opened 2026-09-01
ncdu: Update to fork (1.23.0)

Omikorin:
Hi, because of unfortunate event described [here](https://vndb.org/t24787) the work of Yorhel is continued under [this fork](https://github.com/BratishkaErik/ncdu).

There wasn't any big change for the original C version, but it should be better to update the sources to active fork (at least active for Zig rewrite and maintained for C).
It uses GitHub Attestations (that uses sigstore) instead of PGP, so it is impossible to add a new key.

### DIFF
--- a/ncdu/PKGBUILD
+++ b/ncdu/PKGBUILD
@@ -14,9 +14,8 @@ msys2_references=(
 license=('spdx:MIT')
 depends=('ncurses')
 makedepends=('ncurses-devel' 'autotools' 'gcc')
-source=("${pkgname}-${pkgver}.tar.gz::https://github.com/BratishkaErik/${pkgname}/archive/v${pkgver}.tar.gz")
-sha256sums=('086063ac09b293d2510f08af0e9e4eb52d2856f77295989e2fe4e18135c5cb18')
-validpgpkeys=('74460D32B80810EBA9AFA2E962394C698C2739FA') # Yoran Heling <info@yorhel.nl>
+source=("https://github.com/BratishkaErik/${pkgname}/releases/download/v${pkgver}/${pkgname}-${pkgver}-src.tar.gz")
+sha256sums=('2eccb4ce6c4e08d9539155789ff7341fdbde67bd746e9f523d5d3095cd872804')
 
 build() {
   cd "${srcdir}/${pkgname}-${pkgver}"
--- a/ncdu/PKGBUILD
+++ b/ncdu/PKGBUILD
@@ -3,29 +3,39 @@
 pkgname=ncdu
 pkgver=1.23.0
 pkgrel=1
-pkgdesc="Disk usage analyzer with an ncurses interface."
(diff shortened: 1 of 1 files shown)


fallenmi (changes requested):
The fork publishes an attested source artifact, but this PKGBUILD bypasses it. `archive/v${pkgver}.tar.gz` hashes to `086063ac09b293d2510f08af0e9e4eb52d2856f77295989e2fe4e18135c5cb18`, and GitHub has no attestation for that digest. The release asset `ncdu-${pkgver}-src.tar.gz` hashes to `2eccb4ce6c4e08d9539155789ff7341fdbde67bd746e9f523d5d3095cd872804` and has attestations tied to `refs/tags/v1.23.0` and `.github/workflows/c-release.yml`.

Because this change removes the PGP-verified source specifically because upstream moved to Sigstore, please point `source` at the attested release asset, update the checksum, and remove the now-unused `validpgpkeys`. That aligns the pinned bytes with upstream's published provenance. It does not make `makepkg` verify the attestation automatically—that remains out of band until the packaging workflow adds an explicit verification step—but the currently selected archive has no attestation to verify.

Reviewed with OpenAI Codex and Anthropic Claude assistance.

Omikorin:
Okay, fixed.

I understood the [guidelines](https://www.msys2.org/dev/update-package/) as to add a new key without removing the old one.

fallenmi (approved):
Thanks — confirmed that the PKGBUILD now uses the attested `ncdu-1.23.0-src.tar.gz` release asset with its matching SHA-256, and the unused `validpgpkeys` entry is gone. Both checks pass on this exact head.

Reviewed with OpenAI Codex and Anthropic Claude assistance.

lazka:
hm, tricky, have any other distros switched?

Currently there are no real code changes in 1.23.0, so no rush I guess.

Omikorin:
> hm, tricky, have any other distros switched?
> 
> Currently there are no real code changes in 1.23.0, so no rush I guess.

I did not find any, although repology is down, so it's kinda difficult.

We can wait with this. The fork continues Yorhel's Zig reboot with some updates to the latest Zig. I found this fork by searching and I think it is the most active one.
I don't know if Zig makes problems, because each version has breaking changes right now.

The biggest problem is when and whether the source will go offline. I've read something on _vndb_ (Yorhel's another project) that his friends will help with paying for servers.

lazka:
> I did not find any, although repology is down, so it's kinda difficult.

yeah, https://github.com/repology/repology-rs/issues/560

Omikorin:
Anitya shows this fork
https://release-monitoring.org/project/ncdu

iedame:
Hello! Nixpkgs actually switched to the BratishkaErik fork back in July: https://github.com/NixOS/nixpkgs/commit/489562d3b1e895ac40b02ae8766093095b27641d

PR #6574, opened 2026-09-02
whois: Update to 5.6.6

Omikorin:


whois/PKGBUILD
```diff
@@ -1,19 +1,25 @@
...
+pkgver=5.6.6
 pkgrel=1
 pkgdesc="The whois client by Marco d'Itri"
 arch=('i686' 'x86_64')
 url="https://www.linux.it/~md/software/"
 msys2_repository_url='https://github.com/rfc1036/whois'
+msys2_references=(
+  'anitya: 5128'
+  'archlinux: whois'
+  'cpe: cpe:/a:whois:whois'
```
lazka: what's the source of this CPE?

Omikorin: I got it from here https://release-monitoring.org/project/5128/

lazka: I see thanks. That looks like a wrong entry there as it doesn't give anything on the NIST side. We currently decide if a package has potential security reports based on the existence of that field so it would be best to remove it.

Omikorin: I thought the same, but I failed to find other packages on that NIST site, so I just went with using anitya. I'll update it shortly.

lazka: oh, it could be I've also trusted anitya in the past :/

lazka: I should look into validating CPEs against the grype DB to see if there are any CVEs registered for them.

lazka: tracked here now: https://github.com/msys2/msys2-web/issues/80

PR #6577, opened 2026-09-02
lua: update to 5.5.1

Omikorin:


lazka:
needs rebuilds: tio, vim

Omikorin:
I've tested rebuilding locally. Do I need to commit something in this case?

mmuetzel:
> I've tested rebuilding locally. Do I need to commit something in this case?

Yes, please bump the `pkgrel` of these package in their `PKGBUILD`. That indicates that they need to be rebuilt.

Omikorin:
Ah, thank you very much

PR #6579, opened 2026-09-03
uv: update to 0.12.9

Omikorin:
Funnily enough there exists CPE for uv, so it's not a false alarm :D

I've added `*.crate` binary blobs to gitignore.

### DIFF
--- a/uv/PKGBUILD
+++ b/uv/PKGBUILD
@@ -3,12 +3,20 @@
 _realname=uv
 pkgbase=${_realname}
 pkgname=("python-${_realname}-build")
-pkgver=0.12.5
+pkgver=0.12.9
 pkgrel=1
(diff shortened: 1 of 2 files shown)


uv/PKGBUILD
```diff
@@ -19,25 +27,25 @@ makedepends=('rust'
              'git')
 options=('!strip')
 source=("git+${url}.git#tag=${pkgver}"
-        "cap-primitives-4.0.2.crate::https://static.crates.io/crates/cap-primitives/4.0.2/download"
+        "cap-primitives-4.0.3.crate::https://static.crates.io/crates/cap-primitives/4.0.3/download"
```
ognevny:
```suggestion
        "cap-primitives-4.0.3.tar.gz::https://static.crates.io/crates/cap-primitives/4.0.3/download"
```

and no .gitignore needed

PR #6590, opened 2026-09-03
dtc: update to 1.8.1

Omikorin:
dtc moved from make to meson


Omikorin:
Interesting, CI uses git versions of packages instead of those available via pacman.

I have SWIG 4.4.0 installed locally, the latest available version, but CI used 4.5.0, so there is some mismatch.

What do you think about this behavior @lazka? Is it expected?

lazka:
CI uses staging, see the snippet at the bottom of https://packages.msys2.org/queue

Omikorin:
Ah, I've literally read that before and forgot about it :D

PR #6596, opened 2026-09-04
namcap: update to 3.6.0

Omikorin:


lazka:
I considered dropping this again at some point, since we never used it much... and various things are Linux specific which I don't think is worth porting.

Omikorin:
Yeah, it is not a bad idea. It could be dropped later.
MSYS is too different for this tool, so it would be better to have a specially crafted solution that understands both MSYS and MINGW stuff.

PR #6601, opened 2026-09-04
[remove package] mksh

Omikorin:
I believe it is an obscure and rarely has any users. Seems more like i686 package than any usable shell today.

The prompt displays some gibberish like it is not working correctly. I did not make much research on this error.


lazka:
as the only person filing an issue about mksh @agkozak any thoughts on this?

PR #6602, opened 2026-09-04
pax: update to 20240817

Omikorin:


striezel:
Please don't remove the `-N` / `--forward` option from patch commands as in commit bb87e130ac4f41d8f46aa65e0b7aef3c7b9b912d. Otherwise the build may reverse-apply the patch once it has been merged into upstream. Also see the documentation for patch:
```
       -N  or  --forward
          When a patch does not apply, patch usually checks if the patch looks
          like  it  has  been  applied  already by trying to reverse-apply the
          first hunk.  The --forward option prevents that.  See also -R.
[...]
       -R  or  --reverse
          Assume  that  this  patch  was  created  with  the old and new files
          swapped.  (Yes, I'm afraid that does happen occasionally, human  na‐
          ture being what it is.)  patch attempts to swap each hunk around be‐
          fore  applying  it.  Rejects come out in the swapped format.  The -R
          option does not work with ed diff scripts because there is too  lit‐
          tle information to reconstruct the reverse operation.

          If  the  first hunk of a patch fails, patch reverses the hunk to see
          if it can be applied that way.  If it can, you are asked if you want
          to have the -R option set.  If it can't, the patch continues  to  be
          applied normally.  (Note: this method cannot detect a reversed patch
          if  it  is a normal diff and if the first command is an append (i.e.
          it should have been a delete) since appends always succeed,  due  to
          the  fact  that  a  null  context  matches  anywhere.  Luckily, most
          patches add or change lines rather than delete  them,  so  most  re‐
          versed normal diffs begin with a delete, which fails, triggering the
          heuristic.)
```

Omikorin:
Ah, okay. I didn't quite understood the premise because I tested `makepkg` both with `--cleanbuild` and without. I guess it made a difference for me sometimes. Thanks for explaining. I will check my unmerged PRs then.

As I sweeped through packages, it seems that the `patch` usage doesn't follow any recognizable pattern.

Do you think `patch` invocations should always use `-N` option during preparing / building steps or is it something more like "depends"?

striezel:
> Do you think `patch` invocations should always use `-N` option during preparing / building steps or is it something more like "depends"?

In my opinion, `patch` invocations should always use `-N`, because that way the build will fail with a clean error instead of trying to reverse-apply a patch when the patch has been applied upstream and the MSYS2 contributor / maintainer forgets to remove that patch.

PR #6603, opened 2026-09-04
fish: move msys2.fish to conf.d

felipdsa21:
Allows users to easily disable `msys2.fish` by creating an override in `~/.config/fish/conf.d`.


lazka:
hm, why would you want to do that? And why can't you just comment the sourcing in fish/config.fish instead?

The potential fallout seems not worth it to me.

PR #6604, opened 2026-09-05
patch: update to 2.8

Omikorin:
I went down a rabbit hole...
Unfortunately, there are some internal headers that made this adventure a lot worse, but I managed to do it. At least, `patch` is not broken.

It is viable to patch whole `@STDDEF_NOT_IDEMPOTENT@` stuff in makefiles instead of just one broken include, but I did not find any issues with that.


patch/PKGBUILD
```diff
@@ -1,61 +1,89 @@
...
+  'python'
+)
 optdepends=('ed: for patch -e functionality')
-source=(https://ftp.gnu.org/gnu/${pkgname}/${pkgname}-${pkgver}.tar.xz{,.sig}
-        msys2-patch-2.7.1.patch
-        msys2-patch-manifest.patch)
-sha256sums=('ac610bda97abe0d9f6b7c963255a11dcb196c25e337c61f94e4778d632f1d8fd'
+# tag 2.8 is not signed, so .git?signed cannot be used here and it's better not to skip checksum
+source=(
+  "git+https://git.savannah.gnu.org/git/${pkgname}.git#tag=v${pkgver}"
```
lazka, line 28: Do we really need to use git now? (their git is offline now btw, not sure if temporary..)

patch/msys2-patch-2.7.1.patch
```diff
@@ -1,72 +1,31 @@
-diff -Naur patch-2.7.1/src/inp.c patch-2.7.1-p/src/inp.c
---- patch-2.7.1/src/inp.c	2012-09-19 05:07:31 +0400
-+++ patch-2.7.1-p/src/inp.c	2013-02-24 13:50:56 +0400
-@@ -231,7 +231,11 @@
-     {
-       if (S_ISREG (instat.st_mode))
-         {
-+#ifdef __CYGWIN__
-+	  int ifd = safe_open (filename, O_RDONLY | (binary_transput ?: O_TEXT), 0);
```
lazka, line 9: Do you know what this all did?

patch/PKGBUILD
Omikorin: I think I had some problems with previous method. I'll try again.
Yeah, their git is 95% offline for me for quite a while.

patch/msys2-patch-2.7.1.patch
Omikorin: IIRC this was partially fixed in the upstream. Otherwise patch didn't apply correctly.
I think the code was supposed to force binary mode but there is no reasoning in the PKGBUILD, so I'm not sure about the story.

patch/PKGBUILD
Omikorin: I probably used git as I was unable to fix gnulib. I didn't know about `gl_cv_clean_version_stddef=yes` then.
I changed source to tarball again.

patch/msys2-patch-2.7.1.patch
lazka: ok, I've dropped both patches, I don't think they are needed anymore, see https://github.com/msys2/MSYS2-packages/pull/6604/commits/6bba381d12452d7fbbf5e01d538dcf01a8b04289

striezel:
I'm not sure that dropping those patches was the right thing to do. Currently, the pull request for openjph (https://github.com/msys2/MINGW-packages/pull/31730) fails in one of its rebuilds with patch problems (https://github.com/msys2/MINGW-packages/actions/runs/35341135910/job/105587062249#step:9:6380):
```
  ==> Extracting sources...
    -> Extracting imagecodecs-2026.8.16.tar.gz with bsdtar
  ==> Starting prepare()...
  patching file setup.py
  Hunk #1 FAILED at 194 (different line endings).
  1 out of 1 hunk FAILED -- saving rejects to file setup.py.rej
  ==> ERROR: A failure occurred in prepare().
```
Note that neither the patch file itself nor the source of imagecodecs-2026.8.16.tar.gz has changed. However, that build job used the new patch 2.8 package, because it's now in staging. So I guess it has something to do with the removed patch for text mode handling, but I may be wrong. Can someone check this?

lazka:
will do, thanks

lazka:
same without my last commit, I've reverted for now and will look into it later.

lazka:
One more test, 2.7.6 without the patches also works fine, so it might be an upstream difference (or some configure differences)

lazka:
> One more test, 2.7.6 without the patches also works fine, so it might be an upstream difference (or some configure differences)

oh, no, I mixed things up, it is the patches.

lazka:
Second try in #6684

Our "patch" has a MSYS2 specific quirk, it will normalize all target files, so patching a CRLF file with a LF patch works, but the target gets converted to LF then. Upstream patch only supports it the other way around, CRLF patch, LF target -> patched LF target.

PR #6609, opened 2026-09-05
python-pygments: update to 2.21.0

Omikorin:


striezel:
FYI: There is usually no need to explicitly `cd` into `${srcdir}` at the start of the `build()` or `package()` functions. The functions start out in `${srcdir}` anyway.

Omikorin:
Thanks, I'll check my other PRs. I've been misled by some other packages from here / Arch.

striezel:
Package recipes here are not necessarily coherent, because they have been created over the course of several years and the authors used what they perceived as common practice at that time. And yes, sometimes they copy from older stuff, so things got mixed up between older and newer ways to do things. Not ideal, but as long as it builds and the packaged stuff works, there's no real pressure to harmonize the PKGBUILD files.

Omikorin:
Sure, I took time to do it well during updating. It would be cumbersome to even try doing that as a sole task.

PR #6611, opened 2026-09-05
python-charset-normalizer: update to 3.5.1

Omikorin:
Removed not needed mingw logic with a marginal optimization as a bonus

### DIFF
--- a/python-charset-normalizer/PKGBUILD
+++ b/python-charset-normalizer/PKGBUILD
@@ -1,7 +1,7 @@
 # Contributor: Dirk Stolle
 
-_realname=charset-normalizer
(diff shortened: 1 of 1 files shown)


striezel:
Please don't "remove useless mingw logic" as in commit 0b41fae72733541feb231740622dae56179258fa. That logic is there to perform an out-of-source build, leaving the original extracted source intact and building in a separate directory. It's good practice to do so and it should not be changed, unless the packaged software does not support out-of-source builds.

Omikorin:
Understood, thank you. I thought this is some stuff copied from the MINGW packages. I will check my unmerged PRs then.

PR #6613, opened 2026-09-05
python-hatchling: update to 1.32.0

Omikorin:
Blocked by #6612

### DIFF
--- a/python-hatchling/PKGBUILD
+++ b/python-hatchling/PKGBUILD
@@ -1,12 +1,12 @@
 # Maintainer: Christoph Reiter <reiter.christoph@gmail.com>
 
-_pyname=hatchling
-pkgname=python-hatchling
+_realname='hatchling'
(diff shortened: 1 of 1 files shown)


striezel:
I'm not quite sure about the `python-editables`dependency. It doesn't seem to be a required dependency, because it builds with and without it. Furthermore, the [mingw-w64-python-hatchling package](https://packages.msys2.org/base/mingw-w64-python-hatchling) does not require it. So it probably should be moved to `optdepends`, with a short explanation about which cases can benefit from `python-editables`.

Anyways, I'll leave that for later and merge the update as it is now.

Omikorin:
The `python-editables` dependency is highlighted mainly because I sorted deps alphabetically. No clue why MSYS has it but MINGW not. Arch seems to use `editables` too.

I'll research if it matters for this package later.

PR #6615, opened 2026-09-05
python-setuptools-scm: update to 10.2.3

Omikorin:
Blocked by #6614

### DIFF
--- a/python-setuptools-scm/PKGBUILD
+++ b/python-setuptools-scm/PKGBUILD
@@ -1,19 +1,22 @@
 # Maintainer: Alexey Pavlov <alexpux@gmail.com>
 
-_pyname=setuptools_scm
-_realname=setuptools-scm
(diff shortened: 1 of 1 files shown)


striezel:
I've reverted back to version 10.2.3, because 10.3.2 (as well as 10.3.1 and 10.3.0) have been yanked from PyPI: https://pypi.org/project/setuptools-scm/10.3.2/
Currently, the recommended version to use is 10.2.3:

> This release has been [yanked](https://pypi.org/help/#yanked) by its maintainers, and will be ignored by installers, except when explicitly specified.
> Consider using [release 10.2.3](https://pypi.org/project/setuptools-scm/10.2.3/) instead.
> Reason given by maintainers: pretend version supresses file finders

Omikorin:
Oh, MINGW package has been merged already. Fortunately, 1.3.3 is on its way

PR #6618, opened 2026-09-06
windows-default-manifest: sync from MINGW-packages

dragon-archer:


windows-default-manifest/PKGBUILD
```diff
@@ -1,33 +1,21 @@
...
-_GIT_TAG="release-${pkgver//\./_}"
-source=("git://sourceware.org/git/cygwin-apps/${pkgname}.git#tag=${_GIT_TAG}")
-sha256sums=('SKIP')
-
-prepare() {
-  cd "${pkgname}"
-
-  autoreconf -fiv
-}
+license=('custom:Public Domain')
```
lazka, line 9: I missed this in the mingw one. Can we make a different license? "Public Domain" is not open source in some countries. The source is "FSFAP" for example.

dragon-archer: I don't have much idea of the license. I previously chose "Public Domain" because I thought the real RC content is actually "placed in public domain". But considering that we're now generating it from your Python script, I think it's up to you to choose the license.

PR #6619, opened 2026-09-06
[new package] perl-params-someutil 1.09

Omikorin:
This package is needed for the latest `perl-DataOptList` and `perl-Sub-Exporter`. This is a hard fork of `perl-Params-Util` with some long awaited (but not published) fixes.

After update, the original `perl-Params-Util` package is no longer used by any other package.

Some differences:
I've used `any` architecture because I saw that in `LWP-MediaTypes` and many other Perl packages. I guess all of them can be moved to `any` but I could be wrong. Any opinion?

This PKGBUILD is closer to the MINGW (and Arch) one.

I wasn't sure whether I should use the out of source build method (`build-${MSYSTEM}`) @striezel explained to me.
Mostly because the original package wasn't built like that. Although MINGW does that.

Anyway, I followed "best practise" so it should be okay.

### DIFF
--- a/perl-Params-SomeUtil/001-xs-version.patch
+++ b/perl-Params-SomeUtil/001-xs-version.patch
@@ -1,0 +1,11 @@
+--- Makefile.PL.orig	2026-09-07 00:05:39.655735100 +0000
++++ Makefile.PL	2026-09-07 00:32:20.504667900 +0000
+@@ -14,7 +14,7 @@
+ unless ( defined $make_xs ) {
+ 	$make_xs = can_xs();
+ }
+-if ( $^O eq 'cygwin' and $make_xs == 1 and not /^-xs/ ) {
++if ( 0 ) {
+ 	# Cygwin goes bonkers breaking `` if using Params::SomeUtil XS version
+ 	# for no apparent reason.
+ 	$make_xs = 0;
--- a/perl-Params-SomeUtil/PKGBUILD
+++ b/perl-Params-SomeUtil/PKGBUILD
@@ -14,8 +14,18 @@ msys2_references=(
 license=('spdx:Artistic-1.0-Perl OR GPL-1.0-or-later')
 groups=('perl-modules')
 depends=('perl')
-source=("https://cpan.metacpan.org/authors/id/P/PL/PLICEASE/${_realname}-${pkgver}.tar.gz")
-sha256sums=('dd4b2286137560fc905c5e2db0bcae19fef20614ac4238dbcb2a4e78f82eb560')
+source=(
+  "https://cpan.metacpan.org/authors/id/P/PL/PLICEASE/${_realname}-${pkgver}.tar.gz"
+  '001-xs-version.patch'
+)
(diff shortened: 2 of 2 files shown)


striezel:
The old `perl-Params-Util` package also builds a DLL (`/usr/lib/perl5/vendor_perl/auto/Params/Util/Util.dll`) but the `perl-Params-SomeUtil` package does not. Any idea why that is? Is that intended by the fork?

For what it's worth, the corresponding Arch package [perl-params-someutil](https://archlinux.org/packages/extra/x86_64/perl-params-someutil/)  also contains a shared object (`/usr/lib/perl5/5.42/vendor_perl/auto/Params/SomeUtil/SomeUtil.so`), i. e. the Linux equivalent of a Windows DLL, so it probably should be here, too.

Omikorin:
I found the culprit:

https://github.com/uperl/Params-SomeUtil/blob/main/Makefile.PL#L17

I have no idea if it applies to msys the same as to cygwin.

Omikorin:
I changed `Makefile.PL` locally to this:

```diff
foreach ( @ARGV ) {
	/^-pm/ and $make_xs = 0;
	/^-xs/ and $make_xs = 1;
}
+print "after foreach make_xs: $make_xs\n";
unless ( defined $make_xs ) {
	$make_xs = can_xs();
+	print "after unless make_xs: $make_xs\n";
}
if ( $^O eq 'cygwin' and $make_xs == 1 and not /^-xs/ ) {
	# Cygwin goes bonkers breaking `` if using Params::SomeUtil XS version
	# for no apparent reason.
	$make_xs = 0;
+	print "inside if make_xs: $make_xs\n";
}
+print "outside if make_xs: $make_xs\n";
```

Logs without `-xs` provided:

```text
# PERL_MM_USE_DEFAULT=1 PERL_AUTOINSTALL=--skipdeps perl Makefile.PL INSTALLDIRS=vendor
after foreach make_xs:
after unless make_xs: 1
inside if make_xs: 0
outside if make_xs: 0
```

Logs with `-xs` provided:

```text
# PERL_MM_USE_DEFAULT=1 PERL_AUTOINSTALL=--skipdeps perl Makefile.PL -xs INSTALLDIRS=vendor
after foreach make_xs: 1
inside if make_xs: 0
outside if make_xs: 0
```

It is the Cygwin check that resets the DLL building indeed. I will make a simple patch to disable the `if` block.

Omikorin:
I patched it. I am not sure whether this DLL file is of any use, but it's better to copy the original package behavior.

striezel:
Are you sure the patch is working as intended? According to the file diff there still is no DLL file. :thinking:

By the way, version 1.11 has been released (https://metacpan.org/dist/Params-SomeUtil/changes). It only seems to update documentation, so I don't expect any behaviour changes.

Omikorin:
Hmm, it's interesting because it generated the DLL file for me on UCRT64... will need to take a look again.

I will wait with this until lowercased Perl PR is finished.

PR #6625, opened 2026-09-07
libpcap: update to 1.10.7

Omikorin:
I changed variable used in `source` from `pkgname` to `pkgbase` because `pkgname` is an array that just accidentally resolves into `libpcap` (the first item) but it is still not explicit.


striezel (approved):
Looks good to me.

PR #6629, opened 2026-09-07
jq: remove package

oech3:
As said as https://github.com/msys2/MSYS2-packages/issues/2570 , `jq` does not appear in bootstrapping process of any MinGW packages.


lazka:
Since we are closer to cygwin and maintenance of cygwin packages is much simpler (in many cases no patches required), I don't mind having more cygwin packages just because we can.

my2c

PR #6632, opened 2026-09-07
perl-mime-tools: update to 5.518

Omikorin:
Usual cleanup and update


striezel (approved):
Looks OK.

However, I'm not quite sure about the repository URL. As far as I am aware, this is the first repository for MSYS2 packages which is password-protected. I'll leave it in there, because it's technically valid and the "access denied" page tells you how to get access.

Omikorin:
Yeah, it's weird.

This repo was available earlier at https://github.com/Perl-Toolchain-Gang/MIME-tools
but there is nothing now.

PR #6633, opened 2026-09-07
re2c: update to 4.6

Omikorin:
Usual cleanup and update

### DIFF
--- a/re2c/PKGBUILD
+++ b/re2c/PKGBUILD
@@ -1,37 +1,54 @@
 # Maintainer: J. Peter Mugaas <jpmugaas@suddenlink.net>
 # stolen from Archlinux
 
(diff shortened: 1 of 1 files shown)


striezel:
Not a big issue, but adding the `local _cmake_options` stuff for just two options seems like overkill.
There's no hard guideline on that, but as a rule of thumb I'd say if the complete `cmake` invocation with all its options fits on a 80 character line, then it's not really worth the effort to introduce an extra local variable for the different options. Maybe that can save you a bit of time on future cleanups.

Omikorin:
Yeah, looking back at it this case is too simple for this.
You can call it fuzzing through ideas to get it right finally. Going case by case, I learnt it needs more iterations because of many different build systems.

PR #6648, opened 2026-09-10
Lowercase Perl packages names

### LINKED ISSUE #6623: MSYS Perl package names inconsistent with MINGW and Arch
Hello,

I think it would be a good idea to maintain Perl packages' names consistent across the MSYS and MINGW repositories.
We could use the same `_perlname` pattern that MINGW and Arch implemented.

Of course, I am in favor of lowercase names.

Is it feasible, given that these names already exist?

I wrote a oneliner to check the current status of all directories with uppercase names:

```bash
find . -maxdepth 1 -mindepth 1 -type d -name '*[[:upper:]]*' -print | tee /dev/tty | wc -l
```

There are _130_ of them, and these are just Perl-related packages.

<details>
<summary>Results</summary>

```txt
./perl-Algorithm-Diff
./perl-Alien-Build
./perl-Alien-Libxml2
./perl-Archive-Zip
./perl-Authen-SASL
./perl-Benchmark-Timer
./perl-Capture-Tiny
./perl-Carp-Clan
./perl-Class-Inspector
./perl-Class-Method-Modifiers
./perl-Clone
./perl-Compress-Bzip2
./perl-Convert-BinHex
./perl-Data-Munge
./perl-Data-OptList
./perl-Date-Calc
./perl-DBI
./perl-Devel-GlobalDestruction
./perl-Digest-HMAC
./perl-Encode-compat
./perl-Encode-Locale
./perl-Error
./perl-Exporter-Lite
./perl-Exporter-Tiny
./perl-ExtUtils-Config
./perl-ExtUtils-Depends
./perl-ExtUtils-Helpers
./perl-ExtUtils-InstallPaths
./perl-ExtUtils-MakeMaker
./perl-ExtUtils-PkgConfig
./perl-FFI-CheckLib
./perl-File-chdir
./perl-File-Copy-Recursive
./perl-File-Listing
./perl-File-Next
./perl-File-ShareDir
./perl-File-ShareDir-Install
./perl-File-Which
./perl-Font-TTF
./perl-Getopt-ArgvFile
./perl-Getopt-Tabular
./perl-HTML-Parser
./perl-HTML-Tagset
./perl-HTTP-Cookies
./perl-HTTP-Daemon
./perl-HTTP-Date
./perl-HTTP-Message
./perl-HTTP-Negotiate
./perl-Import-Into
./perl-Importer
./perl-IO-HTML
./perl-IO-Socket-INET6
./perl-IO-Socket-SSL
./perl-IO-String
./perl-IO-Stringy
./perl-IPC-Run3
./perl-JSON
./perl-List-MoreUtils
./perl-List-MoreUtils-XS
./perl-Locale-Gettext
./perl-LWP-MediaTypes
./perl-LWP-Protocol-https
./perl-MailTools
./perl-Math-Int64
./perl-MIME-Charset
./perl-MIME-tools
./perl-Module-Build
./perl-Module-Build-Tiny
./perl-Module-Pluggable
./perl-Module-Runtime
./perl-Moo
./perl-Mozilla-CA
./perl-Net-DNS
./perl-Net-HTTP
./perl-Net-IP
./perl-Net-SMTP-SSL
./perl-Net-SSLeay
./perl-PadWalker
./perl-Parallel-ForkManager
./perl-Params-SomeUtil
./perl-Params-Util
./perl-Path-Class
./perl-Path-Tiny
./perl-PerlIO-gzip
./perl-Pod-Parser
./perl-Probe-Perl
./perl-Regexp-Common
./perl-Return-MultiLevel
./perl-Role-Tiny
./perl-Scope-Guard
./perl-Socket6
./perl-Sort-Versions
./perl-Spiffy
./perl-Sub-Exporter
./perl-Sub-Exporter-Progressive
./perl-Sub-Info
./perl-Sub-Install
./perl-Sub-Quote
./perl-TAP-Harness-Archive
./perl-Term-Table
./perl-TermReadKey
./perl-Test-Base
./perl-Test-Deep
./perl-Test-Exit
./perl-Test-Fatal
./perl-Test-Needs
./perl-Test-NoWarnings
./perl-Test-Pod
./perl-Test-Requires
./perl-Test-RequiresInternet
./perl-Test-Script
./perl-Test-Warnings
./perl-Test-YAML
./perl-Test2-Suite
./perl-Text-CharWidth
./perl-Text-Diff
./perl-Text-WrapI18N
./perl-TimeDate
./perl-Try-Tiny
./perl-Unicode-LineBreak
./perl-URI
./perl-WWW-RobotRules
./perl-XML-LibXML
./perl-XML-NamespaceSupport
./perl-XML-Parser
./perl-XML-SAX
./perl-XML-SAX-Base
./perl-XML-Simple
./perl-YAML-Syck
./perl-YAML-Tiny
```

</details>

Edit: As I researched it, it is kinda possible, just a hassle, but could be split into many PRs.

Omikorin:
Resolves #6623

I've taken steps described below:

1. Added `_perlname` usage everywhere. There are some always-lowercased exceptions but consistency is more important when thinking about possible future scripting. I found even a reversed "standarization" that went from lowercase to uppercase :D
2. Fixed some leftovers that automated script didn't cover like really rare and old `_perlmod`
3. Renamed all deps inside `depends`-like arrays for Perl packages
4. Made Perl `provides` lowercase and updated the `patchprov` script. I didn't actually ran the script for clarity because it would update some dependencies and change their order.
5. Added `conflicts, provides, replaces`

It would be better to merge this PR before #6620, #6621, and #6619, so rebasing this monstrosity won't be so harsh.

There won't by any additional cleanup in this PR.


perl/PKGBUILD
```diff
@@ -209,106 +209,106 @@ package_perl() {
...
-          'perl-Tie-RefHash=1.41'
-          'perl-Time-HiRes=1.9778'
-          'perl-Time-Local=1.35'
-          'perl-Time-Piece=1.36'
-          'perl-Unicode-Collate=1.31'
-          'perl-Unicode-Normalize=1.32'
-          'perl-Win32=0.59_01'
-          'perl-Win32API-File=0.1203_01'
-          'perl-XSLoader=0.32'
+provides=('perl-archive-tar=3.04'
```
lazka, line 212: That's in theory a breaking change, it would need to provide both variants for some time.

lazka:
@dscho just to be sure, do these many renames make problems for GFW automations?

dscho:
> @dscho just to be sure, do these many renames make problems for GFW automations?

Thanks for the ping! I think we should be fine, there are a couple of places I'll need to adapt, but nothing causing major heartaches.

perl/PKGBUILD
lazka: Though I'm not sure if there are any users of this we care about, so maybe it's not an issue.

Omikorin: I can add them back, good point.

Omikorin:
Hm, it broke, but seems like a CI thing

dscho:
> Hm, it broke, but seems like a CI thing

I don't quite think so, the real culprit is [this](https://github.com/msys2/MSYS2-packages/actions/runs/35000446757/job/104487239655?pr=6648#step:9:11236):

> ```
> [install] perl-term-table
>   loading packages...
>   resolving dependencies...
>   looking for conflicting packages...
>   warning: removing 'perl-term-table-0.028-3' from target list because it conflicts with 'perl-5.42.3-2'
> ```

So either this Perl module must be made compatible with the current `perl` package, or it needs to be excluded from this here PR, I'd think.

Omikorin:
Okay, I managed to find the culprit.

Perl has the following `provides` list:

```bash
provides=(
          # ...
          'perl-Term-Table=0.024'
          'perl-term-table=0.024'
)

but the `Term-Table` module starts with this:

```bash
_perlname='Term-Table'
_realname="${_perlname,,}"
pkgname="perl-${_realname}"
pkgver=0.028
pkgrel=3
```

I'm not sure why Perl even provides a different version than the package with the same name.
I shouldn't downgrade already existing package, I think.

@dscho can you explain this bit?

Edit: It probably broke because of the new, explicit `provides` in `Term-Table`:

```bash
provides=("perl-Term-Table=${pkgver}")
conflicts=('perl-Term-Table')
replaces=('perl-Term-Table')
```

dscho:
@Omikorin since `perl` says that it already provides this module, I guess that it truly is an incompatibility with the current Perl? I only see two ways out: either delete that package in this PR, or make it require an older Perl version.

lazka:
From what I understand both packages should be able to coexist, the perl provided one and the cpan one.

lazka:
I can't think of a way to fix this though, maybe we should just drop all those provides from perl, to avoid conflicts with the real packages. Or just revert 765bf60c20dd66ce3966e for now.

One could remove the conflicts from Term-Table, but that would break non-system-upgrades. Not supported in theory, but still.

dscho:
Wasn't there a way to allow certain files in packages to be marked as "user may modify"? I think it was [`backup`](https://wiki.archlinux.org/title/PKGBUILD#backup). Maybe those are allowed to live in multiple packages simultaneously?

Otherwise the best course of action might be to exclude the `perl-term-table` files from `perl` e.g. by deleting them in the `package` function of the `PKGBUILD` file (and of course dropping it from the `provides` array).

lazka:
perl internals depend on term-table, so I don't think that's possible.

The issue here is that a package wants to conflict on another package, but not on the virtual package of the same name provided by another package. Not something pacman supports. From what I understand.

Omikorin:
Let me try something. It will need to be fixed for some packages later.

Perl includes dozens of dual-life modules that ship with the core distribution but are also maintained independently on CPAN.
Those packages can be used with a newer version if user wishes so, but if not, then he has the built-in packages available. CPAN modules simply overshadows Perl built-in modules.

From that reasoning, a CPAN module should not have `conflicts` nor `replaces` in PKGBUILD. If I'm right, then keeping Perl's both module versions and removing those options from CPAN modules should make them not conflict anymore.

The disadvantage of that solution is if some external package depends on a newer CPAN module then it will revert to using Perl's built-in module.

We can fix that for MINGW packages quickly on the staging level.

Omikorin:
Hmm, there is no CI workflow running

lazka:
If there is no replaces then upgrades will fail because pacman could have the old naming installed and tries to install the new named package which has file conflicts.

Omikorin:
Maybe this https://github.com/msys2/MSYS2-packages/pull/6648/commits/d6026f8c15caed95e1928f2c8f7881ee08b3819e will change the situation. Otherwise, we need to revert the commit for double-named built-ins.

lazka:
(this might work for now, but once the builtin module gets updated on the next perl minor update, things will conflict again)

Omikorin:
To be honest, if you run `patchprov` right now, there will be updates like you said.

Can we do it temporarily and freeze Perl with its `provides` for some time until we remove the uppercased version?

lazka:
Not quite sure I understand, but I'd be fine with just reverting 765bf60c20dd66ce39

Omikorin:
Okay, nevermind. If reverting works then let's do it.

Omikorin:
Rebased on `master` branch

Omikorin:
Is there still anything blocking or to be updated @lazka? Maybe on the GFW side?

I don't want to hurry, I just want a huge PR's code not to rot if possible.

striezel:
Side note:
Checksum for `perl-alien-build` was wrong and had to be fixed. Done in <https://github.com/msys2/MSYS2-packages/commit/21aab2a10a23d5eb4bd484708f3a51d7012d768b>. 

But there's a bigger problem:
This pull request seems to break the source builds in CI, for example here: https://github.com/msys2/msys2-autobuild/actions/runs/35998175981/job/107628035882#step:10:114
```
  :: Starting full system upgrade...
  resolving dependencies...
  :: installing perl (5.42.3-2) breaks dependency 'perl-IO-Socket-IP' required by perl-HTTP-Daemon
```

Maybe those `provides`/`conflicts` changes are the reason?

piotrvv:
The perl-ack module is missing
ack/PKGBUILD directory

Omikorin:
> The perl-ack module is missing ack/PKGBUILD directory

It's fine, the `depends` list in non-`perl-*` packages will be updated later.

This PR is supposed to lowercase package names and then we will slowly update compatible leftovers.

> Maybe those `provides`/`conflicts` changes are the reason?

No idea. 
Perl was supposed to provide lowercase names. But why is `perl-http-daemon` still uppercased in the logs?

piotrvv:
by the perl-ExtUtils-MakeMaker build

Omikorin:
Is it because existing packages (like `perl-HTTP-Daemon`) still have `depends=('perl-IO-Socket-IP')` in their metadata?

I think Perl really needs those `provides` for the old package names.

piotrvv:
Switch the Perl-built packages back to the old default ones.
Or install a new one before upgrading Perl.

lazka:
I have a suspicion that autobuild should add a fake depend on git there, as it does for base-devel, so the updates for transitive deps of git get pulled int. I'll try a few things.

piotrvv:
Po4a, tig, and git packages to be rebuilt after everything is done.

lazka:
> edit: now it seems to build one package per job and fail during downgrade, good enough for now hopefully. The real fix would be the make the src build install makedepends and uninstall them again I think

real fix deployed now

PR #6654, opened 2026-09-12
[remove package] colormake

Omikorin:
I don't think this simple shell wrapper is of any use today. It's been 12 years since the original repo has been created.
Nor it is on the level needed for a package.


lazka:
yeah, agreed

PR #6661, opened 2026-09-14
aarch64-pc-cygwin: cross toolchain, and cross-building bash and deps

aswin-mcw:
## What this is

Cross toolchain for building MSYS2 on ARM64 Windows, targeting
aarch64-pc-cygwin. We've carried it out of tree for a while and would like to
get it upstream rather than keep rebasing a fork.

It produces a working cross compiler and msys-2.0.dll, and adds the bits the
core userland recipes need to be built with it.

## Build order

Order matters, each one needs the previous:

  mingw-w64-cross-mingwarm64-zlib
  cross-msysarm64-w32api-headers            Win32 API headers
  cross-msysarm64-w32api-runtime            Win32 API import libs
  cross-msysarm64-runtime-devel             newlib + cygwin headers
  cross-msysarm64-binutils                  as, ld, dlltool
  cross-msysarm64-windows-default-manifest  the real default-manifest.o
  cross-msysarm64-gcc-stage1                bootstrap compiler, no libc
  msys2-runtime-aarch64                     newlib
  cross-msysarm64-gcc                       full compiler
  msys2-runtime-aarch64-stage-2             winsup, msys-2.0.dll

## Why gcc is built twice

There's a cycle. binutils and gcc can't be built without libc headers in the
sysroot, and the runtime that owns those headers can't be built without a
compiler.

cross-msysarm64-runtime-devel breaks it by staging the headers on their own,
straight out of the runtime source tree. Then stage 1 builds newlib with a
bootstrap compiler that has no libc, and stage 2 is the real compiler built
against that newlib, which is what winsup gets linked with.

stage1 declares provides=(cross-msysarm64-gcc) so the chain can use it until
stage 2 replaces it.

## Sources

Everything builds from official upstream pinned by commit. Our changes ride on
top as a patch series in the package dir, not a fork branch, so the source is
reproducible and you can see the delta without cloning anything:

  binutils-gdb    1469a04f3 + 13 patches
  gcc             001f9af67 + 29 patches
  msys2-runtime   da84778cb + 109 patches
  mingw-w64       5751d69c8 + 9 patches

They're git format-patch output, so each keeps its original author, date and
message. Filenames carry the subject, which keeps the series greppable.

## Userland recipes

bash, coreutils, gettext, gmp, libiconv, ncurses and readline get an arm
guarded on CARCH == aarch64 || MSYS2_ARM64_CROSS. The native path is
untouched.

Most of what's in those arms can't be inferred from the native build:

- configure probes that AC_TRY_RUN can't answer, because the target doesn't
  run on the build host. bash_cv_dev_stdin, bash_cv_wexitstatus_offset and
  friends are supplied directly.

- generator programs have to be built with the build compiler rather than the
  cross one. ncurses is the awkward one here, it doesn't just compile, it
  generates source with helper programs and runs tic/infocmp during install.

- ncurses also needs --with-strip-program, because `install -s` execs whatever
  program called `strip` is on PATH, and the host one has no pe-aarch64
  support so it rejects a perfectly good ARM64 PE.

- and -D_GNU_SOURCE, because configure's glibc check tests __NEWLIB__, the
  cross sysroot ships a stub _newlib_version.h that never defines it, so the
  probe fails and the widec build loses the wcwidth() prototype.

## CI

.ci/makepkg-aarch64-cross.conf is the cross makepkg profile, CHOST/CC/AR and
the flags every cross package inherits.

.github/lint-arm64-pkgnames.sh rejects a cross package whose pkgname isn't
cross-msysarm64-*, or that installs outside /usr/aarch64-pc-cygwin. We added
it after getting this wrong once: such a package makes pacman overwrite the
host toolchain, including msys-2.0.dll, and you can't repair that from the
shell that's doing the overwriting.

## Testing

Built end to end on a fresh MSYS2 install with nothing cross-related
pre-installed, from a clean clone of this branch:

- all 10 packages build, in the order above
- the runtime scores 264 pass / 4 fail on the winsup testsuite. The four are
  cancel11, link04, mmaptest02 and systemcall, same as our existing builds
- the 7 userland recipes build and stage, 14 packages
- bash's own testsuite runs on the result: 84 groups, 16 failures

## Requirements

One host package beyond base-devel: cocom. It's declared in makedepends, so
makepkg -s will pull it in.

## Splitting

This is a big diff. Happy to split the toolchain packages and the userland
recipe changes into separate PRs if that's easier to review, or to land them
in whatever order suits you.


mmuetzel:
Apparently, there is an issue with the `libiconv-devel` package for x86_64 with this change:
```
:: File /var/cache/pacman/pkg/iconv-1.19-1-x86_64.pkg.tar.zst is corrupted (invalid or corrupted package (checksum)).
error: failed to commit transaction (invalid or corrupted package (checksum))
Do you want to delete it? [Y/n] 
Errors occurred, no packages were upgraded.
Process completed with exit code 1.
```

Maybe, I'm missing the intention of this PR. But shouldn't the x86_64 packages be unaffected?

Edit: I'm very much excited to see that a GCC toolchain for Windows on ARM target is approaching!

msys2-runtime-aarch64-stage-2/PKGBUILD
```diff
@@ -0,0 +1,987 @@
...
+            'SKIP')
+
+_target=aarch64-pc-cygwin
+_build=x86_64-pc-msys
+
+# Derive the gcc version dir from the compiler: it changes between gcc
+# releases (15.0.1 -> 17) and a stale hardcoded path silently breaks the
+# -nostdinc include set and the crt*.o copies.
+GCC_BASE="$(dirname "$(aarch64-pc-cygwin-gcc -print-libgcc-file-name)")"
+prepare() {
```
mmuetzel: This currently produces errors in the CI run like the following:
https://github.com/msys2/MSYS2-packages/actions/runs/34977164646/job/104407716939?pr=6661#step:9:16
```
msys2-runtime-aarch64-stage-2/PKGBUILD: line 258: aarch64-pc-cygwin-gcc: command not found
msys2-runtime-aarch64-stage-2/PKGBUILD: line 258: aarch64-pc-cygwin-gcc: command not found
msys2-runtime-aarch64/PKGBUILD: line 263: aarch64-pc-cygwin-gcc: command not found
msys2-runtime-aarch64/PKGBUILD: line 263: aarch64-pc-cygwin-gcc: command not found
```

Maybe, move that inside the `prepare()` function.
```suggestion
prepare() {
  # Derive the gcc version dir from the compiler: it changes between gcc
  # releases (15.0.1 -> 17) and a stale hardcoded path silently breaks the
  # -nostdinc include set and the crt*.o copies.
  GCC_BASE="$(dirname "$(aarch64-pc-cygwin-gcc -print-libgcc-file-name)")"
```

Afaics, `GCC_BASE` is set again inside the `build()` function anyway.

mmuetzel:
Having a stage-2 cross-compiler is a huge step.
And reading that it is apparently possible to build a `bash` with it that runs natively on Windows on ARM is fantastic. 🎉

I don't know how to bootstrap a new environment from here though. (Would that be `MSYSARM64`? Or a different name?)
Would the next logical step be to use that stage-2 cross-compiler to build a Canadian-cross compiler that runs natively on Windows on ARM?

Is it even ready for that? Last I heard from it, it wasn't completely implemented yet (e.g., no SEH). Is it in a shape now where bootstrapping a new environment would make sense?

msys2-runtime-aarch64-stage-2/woarm64-0081-ci-arm64-wire-the-toolchain-pipeline-to-the-ext.patch
mmuetzel, line 1: Is this patch (and follow ups) needed for MSYS2?

mmuetzel:
Building `cross-msysarm64-windows-default-manifest` failed with the following error:
https://github.com/msys2/MSYS2-packages/actions/runs/34987055221/job/104442307226?pr=6661#step:9:64362
```
  aarch64-pc-cygwin-windres --preprocessor=gcc --preprocessor-arg=-E --preprocessor-arg=-xc --preprocessor-arg=-DRC_INVOKED default-manifest.rc -o default-manifest.o
  /bin/sh: line 1: gcc: command not found
  aarch64-pc-cygwin-windres: preprocessing failed.
```

Maybe, the cross-`windres` needs to call the prefixed `gcc`?

Edit: Never mind, I see you are already working on it: https://github.com/Multicorewareinc/MSYS2-packages/commit/3ec68fcb03dc60d456896899ff59bd9b5af3b4f5

aswin-mcw:
> Having a stage-2 cross-compiler is a huge step. And reading that it is apparently possible to build a `bash` with it that runs natively on Windows on ARM is fantastic. 🎉
> 
> I don't know how to bootstrap a new environment from here though. (Would that be `MSYSARM64`? Or a different name?) Would the next logical step be to use that stage-2 cross-compiler to build a Canadian-cross compiler that runs natively on Windows on ARM?
> 
> Is it even ready for that? Last I heard from it, it wasn't completely implemented yet (e.g., no SEH). Is it in a shape now where bootstrapping a new environment would make sense?

Thanks!

On SEH — that's moved on since you last looked. It's implemented now. The series carries seven patches covering exception and signal handling on AArch64, including Cygwin: Add ARM64 support for SEH handler symbols, Add arm64 support for exceptions.cc, Cygwin: SEH: Fix crash and handle second unwind pass, and Cygwin: Fix SEH and signal handling on AArch64.

Several related pieces are already merged into Cygwin upstream: AArch64 autoload (13 Jun), ssp (14 Jun), dumper (16 Jul), exceptions: Fix AArch64 non-incyg signal handling (3 Aug), and alternate signal stack + ucontext (6 Sep). Three patches remain in review on cygwin-patches — the ws2_32 chained-init fix, the gendef setjmp fix, and cpuid stubs — none of which blocks building a userland

msys2-runtime-aarch64-stage-2/PKGBUILD
aswin-mcw: Done, thanks — moved into prepare() in both recipes; note build() only re-derived it in stage-2, so I added the same guarded derivation to msys2-runtime-aarch64 as well.

aswin-mcw:
> Apparently, there is an issue with the `libiconv-devel` package for x86_64 with this change:
> 
> ```
> :: File /var/cache/pacman/pkg/iconv-1.19-1-x86_64.pkg.tar.zst is corrupted (invalid or corrupted package (checksum)).
> error: failed to commit transaction (invalid or corrupted package (checksum))
> Do you want to delete it? [Y/n] 
> Errors occurred, no packages were upgraded.
> Process completed with exit code 1.
> ```
> 
> Maybe, I'm missing the intention of this PR. But shouldn't the x86_64 packages be unaffected?
> 
> Edit: I'm very much excited to see that a GCC toolchain for Windows on ARM target is approaching!

You're right — the checksum error was a version collision (CI rebuilt iconv-1.19-1 at a name the repo already publishes), fixed with pkgrel bumps.

Your instinct caught more than that though: x86_64 genuinely was being changed — swapped source mirrors, a dropped validpgpkeys, and bash rewriting environ/optind in the native binary. All reverted or arch-guarded.

.ci/fix-msys-libtool.sh
mmuetzel, line 1: I'm probably just missing something obvious. But why is this fixup needed only for the `aarch64` target?
All of what it is changing seems to be related to `cygwin` vs. `msys`. Is that right?
Why isn't that also an issue for the `x86_64` target?

Is it possible that this fix-up was only needed before transitioning to sharing the target triplet with Cygwin?
Could it be dropped now that that transition is done?

aswin-mcw: You're right on both counts — it was only needed while the target was aarch64-pc-msys; now that every cross arm configures with --host=aarch64-pc-cygwin, $host_os is cygwin and libtool's own cygwin* cases match, so the fixup is dead code.

Dropped in 4deda10b along with the private copy gettext carried, verified by rebuilding libiconv and gmp with it removed — both still report libtool supports shared libraries... yes and install their DLLs into usr/bin.

jon-turney:
Re: cross-msysarm64-gcc/99995-msys-aarch64-pseudo-reloc.patch

I think the solution I prefer here is to just remove the code which generates a reference to _pei386_runtime_relocator  from ld, since this function is always being called by the runtime anyhow (and the exported version is just a stub which does nothing, only there to keep ld happy...) (That stub can then be dropped on aarch64. Unfortunately we'll still need to keep it around on x86_64 for backwards compatibility)

See the discussion at https://sourceware.org/pipermail/binutils/2026-July/149962.html

PR #6679, opened 2026-09-17
Update grep, at long last

dscho:
Since GNU grep dropped handling CRLF in the way that MSYS2's own tooling requires, MSYS2 was blocked from upgrading the `grep` package. It is stuck at v3.0.

Let's just bite the bitter apple and carry patches to reinstate (and verify) the previous CRLF handling, with `-U` as an "escape hatch" for those who require official GNU grep behavior (which treats CR as a regular, non-newline character that can be matched via regular expresions).

With this, we can finally upgrade to a current GNU grep version at long last.


Omikorin:
For the problems with gnulib's stddef you should be able to use `gl_cv_clean_version_stddef=yes`.

Issue #6466, opened 2026-07-27
[patch] Broken CRLF handling


QBos07:
### Description / Steps to reproduce the issue

Repro:
```bash
echo 1 > a
echo 2 > b
unix2dos a
unix2dos b
diff a b > d
patch a -o c -i d
```
using a freshly build patch from upstream git works but not msys packaged version

### Expected behavior

exit code 0. the new file c should b identical to b

### Actual behavior

exit code 1. output:
```
Hunk #1 FAILED at 1 (different line endings).
1 out of 1 hunk FAILED -- saving rejects to file c.rej
```

### Verification

- [x] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MINGW64_NT-10.0-26200

QBos07:
patch has a test for this. It can be run using `make check`

Issue #4519, opened 2024-04-08
`svn diff` creates a .patch file with CRLF EOL


ognevny:
### Description / Steps to reproduce the issue

1. clone any subversion repository (`svn checkout ...`)
2. make some changes
3. run `svn diff > test.patch` to create the .patch file
4. try to apply it with `patch`

### Expected behavior

patch applied with no errors

### Actual behavior

`Hunk #1 FAILED at 50 (different line endings).`
this is usually because almost all repositories uses LF ending for files.
is there a way to change this behavior?

### Verification

- [X] I have verified that my MSYS2 is up-to-date before submitting the report (see https://www.msys2.org/docs/updating/)

### Windows Version

MSYS_NT-10.0-22631

### Are you willing to submit a PR?

no

lazka:
Same issue as https://github.com/msys2/MSYS2-packages/issues/6466 and also came up in https://github.com/msys2/MSYS2-packages/pull/6684

ognevny:
not relevant for me anymore, but IIUC it will be fixed, so it's good 😅

lazka:
good you're not needing svn anymore ( I hope) :)

Since #6466 is easier to reproduce than finding an svn repo with CLRF in them, I'm going to close this as a duplicate.

PR #6684, opened 2026-09-18
patch: Update to 2.8 (v2)

lazka:
Re-apply the previous change, but restore the hunks that are still needed:

* Open targets by default in text mode, so CRLF is converted to LF, and
  LF patches can be applied to CRLF targets
* Allow binary patch via tty again

The other hunks should not be needed anymore (assuming things are mounted
as "binary" in cygwin)

One more test fails due to O_TEXT


lazka:
Open questions / tasks:

* Is ac_cv_func_setmode_dos (and the tty patch) needed?
* Is there a better fix possible where we support mixing CRLF + LF while keeping line endings of the target the same?
* Add a regression test to our global test suite (I'm working on that)

Omikorin:
I wonder if `crlf-handling` test is failing because of that.

lazka:
> I wonder if `crlf-handling` test is failing because of that.

yeah, the O_TEXT auto converting to LF breaks it.

> Is ac_cv_func_setmode_dos (and the tty patch) needed?

It is, it makes "--binary" work.

> Is there a better fix possible where we support mixing CRLF + LF while keeping line endings of the target the same?

Here is a vibe-coded patch for that (not usable as is I'd guess): https://gist.github.com/lazka/7d1347fdeb6ebf677e1b9958c9c5f9dc trying to preserve the line endings of the target or falling back of the patch line endings, globally, not on a per-line basis though.

This would make the test pass while preserving the MSYS2 quirk, not sure if we want to maintain that.

lazka:
The better fix would also fix https://github.com/msys2/MSYS2-packages/issues/6466 and #4519

lazka:
Follow-up issue, so we don't forget: #6705

PR #6694, opened 2026-09-20
znc: drop i686

Omikorin:
The IRC bouncer is not an essential package, it isn't even a build tool, so let's drop the 32-bit version.


lazka:
No one builds i686 anymore (??), so not sure why we need to change this.

Omikorin:
I've just checked the repo. I see it now, it is really disabled.

PR #6699, opened 2026-09-20
gtk-doc: update to 1.36.1

Omikorin:
Is `yelp-tools` package necessary? MINGW doesn't use it and even set `-Dyelp_manual=false` option.


lazka:
Not strictly I guess, but does it hurt?

Omikorin:
Just a question that surfaced up during comparison of both versions.

